PR #109, opened 2015-11-15
update scopes to have callable object for upgrading to rails4

digitalcake:
specs passing locally


sethherr:
This looks great. Thank you!

PR #131, opened 2016-03-20
Developer users

sethherr:
Add ability to superadmin developers to add more developer users.


AWildColin:
Idly glancing through this push request, are you sure you want

`          banned: true `

in https://github.com/bikeindex/bike_index/pull/131/commits/031412cda40020fedf4dec374035b97fa2a7cbcb?

sethherr:
@AWildColin thanks for looking through! It's fine though, since that's a spec and all we care about in that test is that the user is updated to `developer` true.

PR #136, opened 2016-03-25
Soulheart replace soulmate

sethherr:
Finally


sethherr:
- Stop loading everything for search into the dom on all the pages. Use async!
- Remove legacy search
- Remove multiple different autocomplete javascript plugins, switch exclusively to selectize
- Remove Soulmate, exclusively switch to Soulheart

PR #144, opened 2016-04-06
Sign up revised styling

sethherr:
*Need mobile styles to understand the page layout*


sethherr:
Opening a separate PR, because starting over is easier

PR #160, opened 2016-04-18
Add revised components fields

sethherr:
Added revised component fields


sethherr:
Not perfect - created issues #162 and #161 - but good enough to move forward.

PR #196, opened 2016-05-03
revised organizations new

timlkelly:


app/helpers/application_helper.rb
```diff
@@ -23,11 +23,13 @@ def current_page_skeleton
...
     when 'info'
       'content_skeleton' unless %w(terms vendor_terms privacy support_the_index).include?(action_name)
+    when 'welcome' 
+      'content_skeleton' if %w(goodbye).include?(action_name)
+    when 'organizations'
+      'content_skeleton' if %w(new).include?(action_name)
     when 'news' then 'content_skeleton'
     when 'feedbacks' then 'content_skeleton'
     when 'manufacturers' then 'content_skeleton'
-    when 'welcome'
```
sethherr: Why remove this? I think goodbye looks good with the content skeleton.

sethherr:
Don't forget to reference the issue (#174) in a commit message and close it!

app/helpers/application_helper.rb
timlkelly: I was just reorganizing them in the order.  It's all still there, but now the one liners are separate from the multi line items

PR #201, opened 2016-05-04
Revised lightspeed

timlkelly:


sethherr:
overwriting with #202

PR #221, opened 2016-06-28
Upgrade to rails 4

sethherr:
I can't put up with 3 anymore. It's time


sethherr:
Switched to puma too, because unicorn was being a jerk and I'd prefer puma anyway.

sethherr:
Yeah... well, we actually did want acts as paranoid ;)

PR #223, opened 2016-06-29
New bike form (partial registration)

sethherr:


sethherr:
Running into some issues with `BParam` parsing, so merging this in - functionality wise, it's complete - but the styles aren't done and the js isn't added.

PR #225, opened 2016-07-01
Legacy rails version

sethherr:
Opened pull request so that travis runs tests - this is not going to be merged.


sethherr:
Functioning Rails4 deploy! Deleting branch because we're good.

PR #230, opened 2016-07-17
I want to help you translate site in Russian

Chiliec:


sethherr:
Hey @Chiliec - we're changing too many pages right now to be able to support additional locales. We're having trouble keeping up with our own development, and I really don't want to provide incorrect translations that are more confusing that just uses google translate.

We'll definitely get in touch as soon as something changes!

PR #253, opened 2016-09-01
lock_type error

timlkelly:


app/assets/javascripts/views/locks/locks_form.js.coffee
```diff
@@ -1,12 +1,12 @@
 class BikeIndex.Views.LocksForm extends Backbone.View
```
sethherr: Can you checkout the original version of this file? None of these changes need to happen.

PR #260, opened 2016-09-07
Fix Bike Manufacturer name issue

sethherr:
- Remove bike.manufacturer_name because it overlaps.
- Dry shit up, only use the attribute that is actually set on bike instead.


sethherr:
For some reason, Travis is building it twice and fails one of the builds, so it behaves as if it failed the specs.

PR #271, opened 2016-09-20
Add/create api v3

sethherr:
Just inherit all things from v2 for now.

Need this in place so it's possible to extend and create new functionality in #256


sethherr:
Helped by http://code.dblock.org/2013/07/19/evolving-apis-using-grape-api-versioning.html

PR #288, opened 2016-09-29
Attempt to resolve issue with spec failure

sethherr:
Something to do with stolen params `date_stolen_input` in embeded entries.

Failure does not happen locally.


sethherr:
Unclear what's causing the error, but it didn't happen when travis was re-run.

PR #289, opened 2016-09-29
Finish up Bike Organizations

sethherr:
- [x] Switch to bike_organizations for association
- [x] Add methods for creating and destroying `bike_organizations`
- [x] Add interactions for `bike_organizations` in bike edit
- [ ] Deprecate creation attributes on bike


sethherr:
Since this will add the ability to add organizations to bikes, we'll have to reference that fact somewhere. Here are the two potential titles I've come up with -

title-comparison

1. Groups + Ownership + Delete
2. Groups, Ownership + Delete

I also tried out "Groups, Owner + Delete" but I didn't like the look of it.

@moleon in particular, looking for your input

moleon:
Groups and delete are such different things, I would avoid putting all
three of them together.

Groups + Ownership
Delete this Bike


Design/Puppetry Portfolio <http://mikeo.net>
Artistic Director: Rough House <http://roughhousetheater.com>

sethherr:
Here's the updated version (with delete separated out) and the adding organization functionality:

screen shot 2016-09-29 at 7 29 11 pm

PR #301, opened 2016-10-21
add cruiser handlebar type

realchrisolin:
Went to add a bike that has cruiser bars, which doesn't exist in the handlebar type dropdown. Pretty sure this commit will fix that moving forward.

### DIFF
--- a/db/seeds/seed_bike_associations.rb
+++ b/db/seeds/seed_bike_associations.rb
@@ -53,6 +53,7 @@ end
 # Seed the handlebar types
 handlebar_types = [
   { name: 'Flat or riser', slug: 'flat' },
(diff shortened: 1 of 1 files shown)


sethherr:
Cruiser handlebars actually fit into one of our previous categories (specifically "rear facing").

We used to have a help section that described [the different handlebar types](https://github.com/bikeindex/bike_index/blob/4a62bf98a1384787383e966bb22a5412498ba821/app/views/bikes/form_help/_handlebar_tip.html.haml), and it would be great to put it back in, similar to how we have modals show up for help in some other places.

Ideally it would also use the bar sprites image that we have to add more clarity:

Bar sprites

PR #302, opened 2016-10-21
update automailer subject to something more accurate

realchrisolin:
Registered a new bike, received an email with the subject "Confirm your Bike Index registration", but since I've already registered other bikes, the email was more of a welcome email and I didn't need to confirm anything. This commit should update the subject to something more relevant for future emails sent to existing users. Feel free to rework the wording if necessary.


sethherr:
Finished Fixing issue in #304.

PR #309, opened 2016-12-22
change "the Bike Index" to "Bike Index" for consistency and grammar

realchrisolin:
Consensus is that it is a proper noun and doesn't need an definite article before it. Case in point, one of the modified files in https://github.com/bikeindex/bike_index/commit/e9e85a9dab1e6f621978d19f9c0ba10772adc22d#diff-db6a2d21e117f910bcd60fb53822a062.


sethherr:
This looks great! Thanks @realchrisolin

PR #310, opened 2016-12-23
Enhancement/continue oauth after accepting terms

ilalkovski:
Continue the oAuth 2.0 doorkeeper flow when the user is signing up with Facebook or Strava for the first time. It will help the mobile/other clients to receive the access token immediately.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -775,7 +775,7 @@ DEPENDENCIES
   wkhtmltopdf-binary
 
 RUBY VERSION
-   ruby 2.2.5p319
+   ruby 2.2.2p95
 
 BUNDLED WITH
-   1.13.5
+   1.12.5
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -775,7 +775,7 @@ DEPENDENCIES
   wkhtmltopdf-binary
 
 RUBY VERSION
-   ruby 2.2.2p95
+   ruby 2.2.5p319
 
 BUNDLED WITH
-   1.12.5
+   1.13.5
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -20,13 +20,16 @@ class WelcomeController < ApplicationController
   end
 
   def user_home
-    bikes = current_user.bikes
-    page = params[:page] || 1
-    @locks_active_tab = params[:active_tab] == 'locks'
-    @per_page = params[:per_page] || 20
-    paginated_bikes = Kaminari.paginate_array(bikes).page(page).per(@per_page)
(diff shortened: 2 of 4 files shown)


Gemfile.lock
```diff
@@ -775,7 +775,7 @@ DEPENDENCIES
   wkhtmltopdf-binary
 
 RUBY VERSION
-   ruby 2.2.2p95
+   ruby 2.2.5p319
```
sethherr: Could you remove the `gemfile.lock` from the PR? This isn't a necessary change.

app/controllers/welcome_controller.rb
```diff
@@ -20,13 +20,16 @@ def goodbye
...
   def user_home
-    bikes = current_user.bikes
-    page = params[:page] || 1
-    @locks_active_tab = params[:active_tab] == 'locks'
-    @per_page = params[:per_page] || 20
-    paginated_bikes = Kaminari.paginate_array(bikes).page(page).per(@per_page)
-    @bikes = BikeDecorator.decorate_collection(paginated_bikes)
-    @locks = LockDecorator.decorate_collection(current_user.locks)
+
+    unless return_to_if_present
```
sethherr: This shouldn't be necessary, we're already [authenticating the user](https://github.com/bikeindex/bike_index/blob/master/app/controllers/welcome_controller.rb#L3) (which [redirects them to accept the terms](https://github.com/bikeindex/bike_index/blob/master/app/controllers/concerns/authentication_helper.rb#L6) if they haven't).

config/initializers/doorkeeper.rb
```diff
@@ -10,7 +10,7 @@
   # This block will be called to check whether the resource owner is authenticated or not.
   resource_owner_authenticator do
     @user ||= User.from_auth(cookies.signed[:auth])
-    unless @user
+    if @user.blank? || !@user.terms_of_service
```
sethherr: Could you add a test for this?

Crucially, what should happen if the user hasn't accepted the terms of service? There should be some sort of error that is included in the API response, right?

sethherr:
I fixed the issue with the broken spec, if you merge in master it should pass.

app/controllers/welcome_controller.rb
ilalkovski: I've added this line so that the client is redirected to the "return_to" (doorkeeper authorization URL) after accepting the terms of service. Without it, the user will be redirected to user home  instead of continuing the oAuth 2.0 authorisation flow. That's why we are storing the "return_to" even when the client has not accept the terms of service. Let me know if it makes sense.

sethherr:
Closed because this is stale.

PR #332, opened 2017-06-29
updated copy for_schools.haml

runt2pb:
i think i did this right


app/views/landing_pages/for_schools.haml
```diff
@@ -9,29 +9,30 @@
           = render "/landing_pages/icons/header_#{landing_page_target}_svg"
         .banner-text
           %h1 Bike Index for <br> Universities
-          %h3 With bicycle theft a growing problem across campuses everywhere, students’ bikes can be difficult to protect without help.
-    .ribbon
```
sethherr: This line `.ribbon` shouldn't have been removed. Otherwise, yes! It's great!

PR #340, opened 2017-07-04
Update for_law_enforcement.haml

runt2pb:


app/views/landing_pages/for_law_enforcement.haml
```diff
@@ -14,21 +14,21 @@
...
-
-        = render partial: '/landing_pages/tweet', locals: { tweet_id: 875747265167491072 }
+          Bike Index users provide serial numbers, proof of ownership and identifying information about their bikes. When a bike is reported lost or stolen, our partners can search Bike Index to identify, locate, and track it from any computer or smartphone. Our extensive archive has helped officers recover bikes involved in criminal situations, from simple theft to chop shops, drug houses - even fencing rings and online black markets.
 
         %p
-          Read about our over #{ number_with_delimiter recovered_bike_count.round(-2) } recoveries on our <a href="/news">blog</a>. For a complete list of Bike Index’s partners, visit <a href="/where">our partners page</a>, and for information on partnering with Bike Index, please contact us.
+          Read about our over #{ number_with_delimiter recovered_bike_count.round(-2) } recoveries on our <a href="/news">blog</a>. For a complete list of Bike Index’s partners, visit <a href="/where">our partners page</a>, and for information on partnering with Bike Index, please <a hreft="mailto:lily@bikeindex.org">contact us</a>.
```
sethherr: I don't think we should include a mailto link here, because there is already a big "contact us" box that it would be better for people to use.

PR #341, opened 2017-07-05
use ISO 8601 date format across site

### LINKED ISSUE #326: Feature request: use the ISO 8601 date format (YYYY-MM-DD)
Hello bike indexers, 

I'd like to suggest a feature that i think will improve the usability of the project: use the [ISO 8601 date format](https://www.cl.cam.ac.uk/~mgk25/iso-time.html) (YYYY-MM-DD) throughout the site, documentation, etc.
It's an international standard and clearer than the website's current USA-centric formatting of MM.DD.YYYY and MM/DD/YYYY.

realchrisolin:
work in progress, having a difficult time figuring out where the ```@stolen_record.date_stolen``` method is defined. resolves #326.

### DIFF
--- a/app/models/stolen_record.rb
+++ b/app/models/stolen_record.rb
@@ -51,6 +51,10 @@ class StolenRecord < ActiveRecord::Base
     '%a %b %d %Y'
   end
 
+  def self.iso8601_date_format
+    '%Y-%m-%d'
+  end
+
   def self.revised_date_format_hour
     "#{revised_date_format} %H"
   end
--- a/app/views/admin/bikes/edit.html.haml
+++ b/app/views/admin/bikes/edit.html.haml
@@ -20,7 +20,7 @@
         - if @bike.current_stolen_record.present?
           = @bike.dl_list_item(@bike.current_stolen_record.phone, 'Phone')
       %dl.dl-horizontal
-        = @bike.dl_list_item(@bike.created_at.strftime('%m-%d-%Y'), 'Created')
+        = @bike.dl_list_item(@bike.created_at.strftime('%Y-%m-%d'), 'Created')
         - if @bike.creation_organization.present?
           = @bike.dl_list_item(@bike.creation_organization.name, 'Creation org')
           = @bike.dl_from_attribute('cached_data')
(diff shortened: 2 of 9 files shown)


app/views/admin/bikes/edit.html.haml
```diff
@@ -20,7 +20,7 @@
         - if @bike.current_stolen_record.present?
           = @bike.dl_list_item(@bike.current_stolen_record.phone, 'Phone')
       %dl.dl-horizontal
-        = @bike.dl_list_item(@bike.created_at.strftime('%m-%d-%Y'), 'Created')
+        = @bike.dl_list_item(@bike.created_at.strftime('%Y-%m-%d'), 'Created')
```
sethherr: We don't need to change the admin sections. While it might be useful, there is some js ordering stuff that uses the date - the dates are in the hidden spans - and changing that would have negative impacts.

Plus, it's the admin section, it isn't user facing.

app/views/admin/bikes/missing_manufacturer.html.haml
```diff
@@ -51,7 +51,7 @@
                 - elsif Date.yesterday().beginning_of_day() <= bike.created_at
                   Yesterday at #{bike.created_at.strftime("%-I %p")}
                 - else
-                  = bike.created_at.strftime("%-m.%-d.%Y")
+                  = bike.created_at.strftime("%Y-%m-%d")
```
sethherr: ditto - it's the admin section, it isn't user facing.

app/views/admin/dashboard/index.html.haml
```diff
@@ -34,16 +34,16 @@
           %td.medium-screens
             - if user.created_at.today?
               %span.hidden
-                = user.created_at.strftime("%-m.%-d.%Y %-H.%M")
+                = user.created_at.strftime("%Y-%m-%d %-H.%M")
```
sethherr: And here as well, we don't need to change admin stuff

app/models/stolen_record.rb
```diff
@@ -51,6 +51,10 @@ def self.revised_date_format
     '%a %b %d %Y'
   end
 
+  def self.iso8601_date_format
```
sethherr: Is this method used anywhere? If not, there is no reason to add it.

app/views/admin/bikes/edit.html.haml
realchrisolin: I was going to ask you about this when I noticed the file path and realized what those pages are for (d'oh!). I had already made some commits by then, but now I know. I'll back them out before the final version

app/models/stolen_record.rb
realchrisolin: Not yet. I saw the revised_date_format method and thought it was worth adding. I'll nuke it if I don't end up using it.

realchrisolin:
Haven't forgot about this, but I've been really busy between work and Bike Index/local cycling advocacy work. Hoping to get to this two Sundays from now.

PR #349, opened 2017-07-28
Fix vagrant environment

realchrisolin:
Time and updates have broken the auto provisioning logic using Chef. The commits in this PR fixes it.


sethherr:
Thanks!

Issue #353, opened 2017-08-19
Fix vagrant dev environment autoprovisioning in Windows


realchrisolin:
```
(work) [pty1] ~/bike_index (branch:master) » vagrant up
Bringing machine 'default' up with 'virtualbox' provider...
==> default: Importing base box 'v0rtex/xenial64'...
==> default: Matching MAC address for NAT networking...
==> default: Checking if box 'v0rtex/xenial64' is up to date...
==> default: Setting the name of the VM: bike_index_default_1503112204589_40863
==> default: Installing Chef cookbooks with Librarian-Chef...
==> default: Auto-generating node name for Chef...
==> default: Destroying VM and associated drives...
C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:309:in `unpack_package!': The package archive has too many children! (RuntimeError)
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:224:in `cache_version_uri_unpacked!'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:93:in `block in version_uri_manifest'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:382:in `memo'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:92:in `version_uri_manifest'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:88:in `version_manifest'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:54:in `version_dependencies'
        from C:/Users/Chris Olin/.vagrant.d/gems/2.3.4/gems/librarian-chef-nochef-0.2.0/lib/librarian/chef/source/site.rb:463:in `fetch_dependencies'
[...]

PR #366, opened 2018-01-22
Delete user if email isn't confirmable

### LINKED ISSUE #130: Handle incorrect emails
SparkPost sends us errors when it can't send because of invalid emails.

`Net::SMTPSyntaxError: 501 5.5.2 RCPT TO syntax error`

We should probably delete the user account if this happens. We definitely shouldn't retry the `email_confirmation_worker` 18 times, which is what we do right now :joy_cat:

mftaff:
fixes #130 

Note: I am unable to test this solution because of the inability to setup locally using vagrant (see #353). However, it is a starting point, and I am happy to make any changes required :-)

### DIFF
--- a/app/workers/email_confirmation_worker.rb
+++ b/app/workers/email_confirmation_worker.rb
@@ -6,5 +6,7 @@ class EmailConfirmationWorker
   def perform(user_id)
     user = User.find(user_id)
     CustomerMailer.confirmation_email(user).deliver_now
+  rescue Net::SMTPSyntaxError
+    user.destroy
   end
 end
--- a/spec/workers/email_confirmation_worker_spec.rb
+++ b/spec/workers/email_confirmation_worker_spec.rb
@@ -8,4 +8,11 @@ describe EmailConfirmationWorker do
     EmailConfirmationWorker.new.perform(user.id)
     expect(ActionMailer::Base.deliveries.empty?).to be_falsey
   end
+
+  it 'deletes user if email is invalid' do
+    user = FactoryGirl.create(:user)
+    user.update(email: 'notaemail@fakeonotreal.blorgh')
+    EmailConfirmationWorker.new.perform(user.id)
+    expect(User.count).to be_zero
+  end
 end


sethherr:
Hey @mftaff - bummer the vagrant setup isn't working. We definitely need to have tests for this though. Once you push up a change to this pull request, it will run on Travis - so even if you can't run the tests locally, you can add a new test to the spec file and it will be tested on our CI - _warning: this isn't the fastest way to do development, it's much better to get things running locally_

The place you need to add a spec is in [spec/workers/email_confirmation_worker_spec.rb](https://github.com/bikeindex/bike_index/blob/master/spec/workers/email_confirmation_worker_spec.rb).

The spec can stub raising the exception, it should test that when the exception is raised it deletes the user.

mftaff:
👍

mftaff:
I am having a bit of trouble with creating the test. It seems that when in test mode, sidekiq isn't actually sending emails: `[rspec-sidekiq] WARNING! Sidekiq will *NOT* process jobs in this environment. See https://github.com/philostler/rspec-sidekiq/wiki/FAQ-&-Troubleshooting`

I may have misunderstood what that link said, but it seems that when testing, the mailers are just being created, but not actually sent out.

Do you know how to get around this? Some way to actually try and deliver (or test the emails validity) the email and this way we should be able to catch any SMTPSyntaxErrors. 

Here is what I have so far:
```
describe EmailConfirmationWorker do
  ...
  it 'deletes user if email is invalid' do
    user = FactoryGirl.create(:user)
    user.update(email: 'notaemail@fakeonotreal.blorgh')
    EmailConfirmationWorker.new.perform(user.id)
    # assuming the EmailConfirmationWorker raised an SMTPSyntaxError the user should be deleted:
    expect(User.count).to be_zero 
  end
end
```

mftaff:
As an aside I just came up with `expect { user.reload }.to raise_error ActiveRecord::RecordNotFound` as the way to check if the user was deleted, as apposed to checking `User.count` which is less specific.

sethherr:
Sorry for the slow response @mftaff ! I think checking that the error is raised is good. I think you could actually stub the delivery of the message and the raising of the error, since you're correct, the mailers don't actually deliver in testing.

mftaff:
No worries!

I have been trying to do exactly that - however, so far I haven't been successful. I haven't written tests for mailers before, so this is a bit new for me. Do you have an idea what this test needs to look like?

PR #423, opened 2018-08-13
Add partial registration list view

sethherr:


sethherr:
Stupid external request failures because travis is busted af

PR #424, opened 2018-08-14
Updated nav scroll in the API docs page

tarnelope:
Clicking the links in the API docs nav wasn't scrolling to the corresponding page position. I updated `scrollToRef` to scroll using `window.scrollTo`. 
Additionally, I removed `$('#header').headroom()` to keep the API token header fixed and keep the offset consistent as sometimes it would appear with scrolling and other times it wouldn't.

Happy to make additional changes!

Note: Scroll animation works fine in Chrome and Mozilla but jumps to the position in Safari. 
bike_index_api_scroll_to


sethherr:
Whooo! Great, thanks.

And thanks for including the gif too 🙂

PR #441, opened 2018-09-15
Widen user/organization column to fill parent div

stevepiercy:
The user/organization show page squishes up its name. This PR fixes the issue. See screenshots of before and after.

## before
before

## after
after

This also respects the empty column to the right, which I assume is for future advertising space. By the way, if you start to display advertising, I would turn off my ad blocker if you follow the [Ethical Ad](https://docs.readthedocs.io/en/latest/advertising/ethical-advertising.html) model established by ReadTheDocs.

### DIFF
--- a/app/views/users/show.html.haml
+++ b/app/views/users/show.html.haml
@@ -7,7 +7,7 @@
             - if @user.avatar
               - unless @user.avatar.url(:medium) == 'https://files.bikeindex.org/blank.png'
                 = image_tag @user.avatar.url(:medium), class: 'users-show-avatar'
-          .col-xs-3
+          .col-xs-9
             %header.with-subtitle
             %h2
               - if @user.title.present?


sethherr:
Thanks so much @stevepiercy - that was a good catch!

sethherr:
With regard to advertising - we do display advertising on some pages, and we do follow the Ethical Ad model with our advertisers. However, we also have google ads (adsense) on a few pages, because finding and getting advertisers is work and we haven't had the time to find new advertisers. If you're volunteering...

stevepiercy:
If you provide me some selling points or sell sheet, I'll do what I can. It would greatly help if you promote that you don't use tracking ads and strictly follow the Ethical Ad model throughout the site. You can't both promote that and use Google AdSense, though.

I will always block tracking ads. It's now a default trend in some web browsers, too, so perhaps the writing is on the wall for unethical ad services. Or at least we can only hope.

PR #443, opened 2018-09-17
Indicate that a hidden bike can be shown

stevepiercy:
Without this statement, I assumed that the setting would be permanent.


sethherr:
Great! Thanks

PR #449, opened 2018-09-26
Prevent double submission of registrations

stevepiercy:
I'm not familiar with Coffeescript, so please improve as needed. I'm uncertain about how to implement the `function` callable.

### DIFF
--- a/app/assets/javascripts/external_scripts/embed_script.js.coffee
+++ b/app/assets/javascripts/external_scripts/embed_script.js.coffee
@@ -164,12 +164,13 @@ $(document).ready ->
   $(".hiddenFieldTimezone").val(window.timezone)
   
   # prevent double posting
-  $('#new_bike').submit(function()) -> 
+  $('#new_bike').submit -> 
     $this = $(this)
-    if ($this.data().isSubmitted) ->
-        return false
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/external_scripts/embed_script.js.coffee
```diff
@@ -162,5 +162,14 @@ $(document).ready ->
   # Update any hidden fields with current timezone
   window.timezone ||= moment.tz.guess()
   $(".hiddenFieldTimezone").val(window.timezone)
-
-  new window.CheckEmail('#bike_owner_email')
\ No newline at end of file
+  
+  # prevent double posting
+  $('#new_bike').submit(function()) ->
```
sethherr: Close! The direct translation from the js you posted in the issue is this:

```coffeescript
# prevent double posting
$('#new_bike').submit ->
  $this = $(this)
  if $this.data().isSubmitted
    return false
  # mark the form as processed, so we will not process it again
  $this.data().isSubmitted = true
```

The `->` is function in coffeescript. I'd recommend [js2coffee](http://js2.coffee/) to convert js, if you ever need to do something like this again.

stevepiercy:
Updated in 7dbc52b

PR #457, opened 2018-11-07
Convert headers for bulk imports

sethherr:


sethherr:
Using some info from https://dalibornasevic.com/posts/68-processing-large-csv-files-with-ruby

PR #465, opened 2018-11-26
Remove unused updated_styles layout

sethherr:


sethherr:
The original style compilation files (pre removing updated_styles) are visible Style compilation files removal are visible in this commit: https://github.com/bikeindex/bike_index/commit/1a8ebdfe98eadd9fa5872cf5b98bd80d2900fb74 - PR #224

PR #466, opened 2018-11-27
Organizations interface improvements, prep export update

sethherr:


sethherr:
Reverting xlsx update because this looks like a little harder of a process. Going to try to figure out another way - focusing on other priorities first. Much of the stuff in this PR is useful though, so I'm keeping it.

PR #470, opened 2018-12-01
Zoom in on messages view to select messages

sethherr:


sethherr:
It's doing things!

screen shot 2018-11-30 at 8 07 08 pm

PR #471, opened 2018-12-03
Add caching for active_model_serializers

sethherr:
`active_model_serializers` after version `0.8` don't actually support caching (yeah, wtf ¯\\\_(ツ)\_/¯).

So switch to that version and cache us some serialializers.


sethherr:
Tried merging in master but it got messy, so skipping

sethherr:
Replaced by #2123

PR #488, opened 2019-01-09
Avery export skip blank labels

sethherr:
Depends on stuff in #487


sethherr:
closed by 6e31dbe in #487

PR #491, opened 2019-01-09
Add codeclimate test reporting

sethherr:
So far I've used these references:

- https://docs.codeclimate.com/docs/circle-ci-test-coverage-example#section-circleci-20-parallel-builds
- https://docs.codeclimate.com/docs/configuring-test-coverage#section-locations-of-pre-built-binaries
- https://github.com/davehenton/identity-pki/blob/master/.circleci/config.yml
- https://github.com/paritytech/clachair/pull/18/files

This is a continuation of #490 - which migrated us to circleci. Instead of holding up the migration on getting codeclimate test coverage working, I open a new branch and let the migration roll


sethherr:
So... this isn't correctly reporting the parallelized results. But I figure some reporting is better than no reporting!

Going to work on this again at some point.

PR #495, opened 2019-01-10
After create user jobs - merge phone & address

sethherr:


sethherr:
Still to do - make this actually do what it's suppose to do, which is merge in phone & address from bikes to users

PR #503, opened 2019-01-18
associate stickers on avery label creation

sethherr:


sethherr:
Actually resolved by #528

PR #514, opened 2019-01-28
add 'payments' to variable graph endpoint

Robertgervais:


spec/controllers/admin/graphs_controller_spec.rb
```diff
@@ -45,19 +53,32 @@
...
-          get :variable, kind: "users", start_at: start_at,
-                         end_at: end_at, timezone: "America/Los_Angeles"
-          expect(response.status).to eq(200)
-          json_result = JSON.parse(response.body)
-          expect(json_result.keys.count).to be > 0
-          Time.zone = TimeParser.parse_timezone("America/Los_Angeles")
-          expect(assigns(:start_at).strftime("%Y-%m-%dT%H:%M")).to eq start_at
-          expect(assigns(:end_at).strftime("%Y-%m-%dT%H:%M")).to eq end_at
-          expect(assigns(:group_period)).to eq "day"
+    context "payments" do
```
sethherr: Indent!

spec/controllers/admin/graphs_controller_spec.rb
```diff
@@ -27,6 +32,9 @@
     end
   end
 
+  describe "variable" do
```
sethherr: What’s this? Probably remove...

spec/controllers/admin/graphs_controller_spec.rb
```diff
@@ -45,19 +53,32 @@
         expect(assigns(:end_at)).to be_within(1.minute).of Time.now
         expect(assigns(:group_period)).to eq "month"
       end
-      context "passed date and time" do
```
sethherr: Looks like this update removed the existing user test. 

The user test should still exist!

Also, it would be great if you tested something more useful than just `keys.count > 0` - I know I didn’t initially, but 🤗

sethherr: was this resolved?

Robertgervais: Yeah, I still have the user and payment test on the branch I just pushed up.

PR #529, opened 2019-02-02
Custom date input ranges

Robertgervais:


app/javascript/source/javascript/initializer.js
```diff
@@ -101,6 +101,7 @@ $(document).ready(function() {
       if ($("#admin-content").length > 0) {
         window.binxAdmin = new BinxAdmin();
         binxAdmin.init();
+        binxAdmin.initGraphs();
```
sethherr: you shouldn't add this here, the `init()` method on `binxAdmin` should determine whether something should be called or not (i.e. you should put the `if (window.location.href.match('\\admin/graphs')){` in the init method, and then call `initGraphs()` if it should be.

app/javascript/source/javascript/binx_admin.js
```diff
@@ -13,4 +15,17 @@ window.BinxAdmin = class BinxAdmin {
         .css("max-width", "100%");
     }
   }
+  initGraphs() {
+    if (window.location.href.match('\\admin/graphs')){
+      window.onload = function(e) {
```
sethherr: You don't need this `window.onload` function - `initGraphs` is only called when page has finished initializing.

app/javascript/source/javascript/binx_admin.js
```diff
@@ -12,5 +14,29 @@ window.BinxAdmin = class BinxAdmin {
         .parents(".receptacle")
         .css("max-width", "100%");
     }
+    if (window.location.href.match('\\admin/graphs')){
+      this.initGraph()
+    };
+  }
+
+  initGraph() {
+    $("select#graph_date_option_choice").on("change", e => {
```
sethherr: All the things in this function other than `preventDefault` should be separated out to a separate function (which you can call with `window.binxAdmin.otherFunctionName()`) both because it will make this cleaner, but also because you should call it when the page loads (so if you're on a custom page it shows the right things)

app/javascript/source/javascript/binx_admin.js
```diff
@@ -12,5 +14,29 @@ window.BinxAdmin = class BinxAdmin {
...
+      e.preventDefault()
+      if ($("select#graph_date_option_choice")[0].value === "custom") {
+        $("#calendar-box").slideDown()
+      }
+      else {
+        $("#calendar-box").slideUp()
+        let $this = $('#start_at')
```
sethherr: No reason to assign `$this` here, since you only use it once. You should call `.val` directly on `$("#calendar-box")`

app/javascript/source/javascript/binx_admin.js
Robertgervais: Okay, I think I've fixed this. I'm wondering about the .on("change") though. Part of me feels like this should go, but also I'm not sure where I'm getting the e in preventDefault() without it. You said that prevenDefault() shouldn't be pulled out of this function, so do I leave it where it is?

Issue #516, opened 2019-01-28
API endpoint for creating organizations on Bike Index


sethherr:
This will require some sort of authentication to prevent spam

PR #541, opened 2019-02-10
Adds organizations api

gadogado:
This resolves https://github.com/bikeindex/bike_index/issues/516

I made a few assumptions on the best implementation and would appreciate any feedback - Here's some of the changes:

- Adds a new `write_organizations` scope which atm is only utilized by the respective v3 endpoint to create an organization.
- Adds ability to create an `organization` with a *single* location.  This can be altered if you'd rather have the ability to provision multiple locations ( or none at all ) but I figured this implementation is more aligned with the current workflow that's on the site.
- Moves most of the state and country seed data into an isolated module which is now also used by the grape param validators.  With this in place you avoid making a DB call with every grape param validation, i.e., `values: Country.pluck(:name)`
- Added some notes and links for the new endpoint: [localhost api docs](http://localhost:3001/documentation/api_v3#!/organizations/POST_version_organizations_format_post_0)
- Fixes organizations <-> locations has_many/belongs_to to include an `inverse_of`.  This will allow you to be able to build nested locations on a new organization without throwing a validation error when the foreign_key/organization_id is missing
- Adds `airborne` which is nice for api focused specs


lib/states_and_countries.rb
```diff
@@ -0,0 +1,308 @@
+module StatesAndCountries
```
sethherr: Why make this a module and not a class?

app/serializers/organization_serializer.rb
```diff
@@ -1,5 +1,4 @@
 class OrganizationSerializer < ActiveModel::Serializer
-  attributes :name, :slug, :url
+  attributes :name, :website, :kind
```
sethherr: Looking through the project, it appears that there is no other place that `OrganizationSerializer` is used. Which means that this change is totally great and makes perfect sense! Is that what you found too?

app/models/organization.rb
```diff
@@ -47,7 +47,6 @@ class Organization < ActiveRecord::Base
   scope :shown_on_map, -> { where(show_on_map: true, approved: true) }
   scope :paid, -> { where(is_paid: true) }
   scope :valid, -> { where(is_suspended: false) }
-  scope :valid, -> { where(is_suspended: false) }
```
sethherr: Good catch!

app/serializers/organization_serializer.rb
gadogado: Ya I don't think it was being used 👍

spec/requests/api/v3/organizations_request_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+        post url, org_json, json_headers
+        expect_status 400
+        expect_json(error: "website is invalid")
+      end
+
+      it "creates a new organization with one location" do 
+        post url, organization_json, json_headers
+        expect(response).to be_successful
+        expect_status 201
+        expect_json(organization: {
```
sethherr: Minor nit: I like using the existing data rather than writing it again. It can result in awkward code (like having to separate out the location attrs into a separate let block above). What I'm thinking is more of - it's nice not only for dryness but because it highlights the differences between request and response. So this would be something like `expect_json(organization: organization_json.merge(locations: [location_attrs.merge(address: "1111 SE Belmont Street, Portland, OR, 97215, United States")))`

This isn't something you need to change, just something for future reference.

lib/states_and_countries.rb
gadogado: You don't really need state or initialization with this so probably makes more sense to keep it lightweight and composable.

spec/requests/api/v3/organizations_request_spec.rb
gadogado: Sure ya that sounds good thanks 👍

sethherr:
Good work, thanks @gadogado !

Some comments about things

- I didn't know about `inverse_of`, that's great!
- I didn't explain the requirements clearly enough with regard to preventing organization creation spam - I want to restrict creating organizations to just applications that have asked us for access to the feature. Since anyone can create OAuth applications on Bike Index, adding a scope won't prevent the thing I'm worried about (people creating nonsense/spammy organizations). The solution I have 
- I think I would prefer passing the location parameters as an array. While it feels mostly unnecessary from the way Bike Index is being used, it makes the response reflect the passed data, which feels like a good practice. Good thought on reducing complexity though!

sethherr:
w/r/t `StatesAndCountries` - I have the same feeling about this as I have about `Color` - that it shouldn't be database backed.

I think that moving the states and countries into a module is a definite step in the right direction, but I'd like to move them entirely out of the database in the future (I just added an issue for it, #545). I'm not sure if we should make this intermediate step, rather than just fully transitioning to non-database storage.

This is up to you @gadogado

gadogado:
@sethherr Thanks for the feedback

- I went ahead and changed the shape of the endpoint to now accept an array of locations;  I'm assuming you'd want some sort of limit on the optional number of locations that can be created per call so I set up a constant to enforce a limit of `five`.  This can be adjusted to whatever works best
- I added an env to capture the client ids that should have access to the create organizations endpoint
- Let's keep the states and countries implementation for now and adjust things when we restructure Color, etc in #545 

For some reason the build is failing for a completely unrelated spec but everything passes locally - wondering if you've encountered this before.

```
# local
Finished in 8 minutes 54 seconds (files took 9.14 seconds to load)
1992 examples, 0 failures, 14 pending

#circleci
Failures:

  1) MergeAdditionalEmailWorker confirmed existing user account merges bikes and memberships and deletes user
     Failure/Error: expect(old_user.ownerships.first).to eq old_user_ownership
 ...
```

sethherr:
I don't think we need to worry about restricting the number of locations for shops. If we've green-lit a client to create organizations, I think we're going to trust them (or remove their access).

With regard to the failing spec - I've had some problems with that. My solution is to try to figure out what is making the spec flaky and then fix the issue

gadogado:
@sethherr ok sounds great - this PR should be good to go now.

PR #542, opened 2019-02-11
[WIP] Update bike through create endpoint if bike already exists

stevenbuccini:
WIP

### DIFF
--- a/app/controllers/api/v3/bikes.rb
+++ b/app/controllers/api/v3/bikes.rb
@@ -171,7 +171,7 @@ module API
               error!("Unable to update bike: #{e}", 401)
             end
             # Manually set the return code in a feeble attempt to be more ~~semantic~~
-            status 200
+            status :ok
             return existing_bike.reload
           end
 
--- a/app/controllers/api/v3/bikes.rb
+++ b/app/controllers/api/v3/bikes.rb
@@ -84,9 +84,21 @@ module API
           error!("You do not own that #{@bike.type}#{addendum}", 403)
         end
 
-        def email_attached_to_bike()
-          return true
-          error!("You do not own that #{@bike.type}", 403)
+        def email_is_attached_to_bike_owner(email)
+          normalized_email = EmailNormalizer.normalize(email)
+          
+          # check primary email first in case we can short-circuit
+          if @bike.owner_email == normalized_email
+            return true
+          end
+
+          secondary_emails = @bike.owner.secondary_emails
+
+          if secondary_emails.include?(normalized_email)
+            return true
+          end
+
+          error!("#{email} does not own that #{@bike.type}", 403)
         end
 
         def ensure_required_stolen_attrs(hash)
@@ -164,7 +176,7 @@ module API
             authorize_bike_for_user
             
             # Additionally, check to ensure the email is attached to the bike (either as primary or secondary)
-            # email_attached_to_bike
+            email_is_attached_to_bike_owner(b_param['params']['bike']['owner_email'])
 
             begin
               # We only want to update the Bike Attrs, which is a subset of the creation parameters
@@ -173,8 +185,9 @@ module API
               error!("Unable to update bike: #{e}", 401)
             end
             # Manually set the return code in a feeble attempt to be more ~~semantic~~
+            existing_bike.reload
             status :ok
-            return existing_bike.reload
+            return existing_bike
           end
 
           bike = BikeCreator.new(b_param).create_bike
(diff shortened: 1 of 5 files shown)


app/controllers/api/v3/bikes.rb
```diff
@@ -0,0 +1,176 @@
+module API
+  module V3
+    class Bikes < API::Base
```
stevenbuccini: I can't figure out the proper way to inherit from the previous API version. Something like `class Bikes < API::V2::Bikes` doesn't pull in any of the methods or helpers, so I replicated them all here so I wouldn't get blocked.

app/controllers/api/v3/bikes.rb
```diff
@@ -0,0 +1,176 @@
...
+          optional :organization_slug, type: String, desc: "Organization bike should be created by. **Only works** if user is a member of the organization"
+          optional :cycle_type_name, type: String, values: CYCLE_TYPE_NAMES, default: 'bike', desc: "Type of cycle (case sensitive match)"
+          use :bike_attrs
+          optional :components, type: Array do
+            use :components_attrs
+          end
+        end
+        post '/', serializer: BikeV2ShowSerializer, root: 'bike' do
+          declared_p = { "declared_params" => declared(params, include_missing: false).merge(creation_state_params) }
+          b_param = BParam.create(creator_id: creation_user_id, params: {'bike': declared_p['declared_params']}, origin: 'api_v3')
```
stevenbuccini: I'm running into issues here -- a lot of this code is replicated from V2, but for some reason a lot of the code expects `declared_params` to be under the `bike` key, but for some reason neither requests fired through the Swagger documentation nor Rspec coerced it properly. I added this key manually to resolve some errors and begin testing logic in the happy case.

app/controllers/api/v3/bikes.rb
```diff
@@ -0,0 +1,176 @@
...
+          ensure_required_stolen_attrs(b_param.params)
+          # Prevents unnecessary DB hits
+          if b_param.errors.present? or b_param.bike_errors.present?
+            e = b_param.errors
+            return error!(e.full_messages.to_sentence, 401)
+          end
+
+          existing_bike = Bike.find_by(serial_normalized: SerialNormalizer.new({serial: b_param[:serial]}).normalized)
+
+          # Assume only one bike for now, but may need to check len of existing bike later depending on what Seth says
```
stevenbuccini: Curious to hear your thoughts on if you think it's likely that there will ever be two normalized serials under the same owner, and if so, how you want to handle that case.

app/controllers/api/v3/bikes.rb
```diff
@@ -0,0 +1,176 @@
...
+          if existing_bike.present?
+            # Does this check secondary emails?
+            authorize_bike_for_user
```
stevenbuccini: Does this function take care of looking at all secondary emails of your owners? I'm also concerned about the secondary DB access here -- I was trying to figure out a way to combine it all into one on line 150, but I couldn't figure out a good way to do so since it looks like you would have to join from Bikes through Ownerships to Users.

stevenbuccini:
Got this feature working in the happy case. There are a few major issues  I still need to work through:
- [ ] Some parameters that are required by the `create` endpoint aren't used in updating and break existing codepaths -- need a better way of identifying these parameters and filtering them from the output
- [ ] Currently authorization is done through the API user, when it really needs to be keyed off the user's email in the update case.
- [ ] We set the `origin` field when creating a `BParams` object to `api_v2`, this is no longer correct but is not just a simple log-only field -- we will need to write code to make this implementation work.
- [ ] Need to come up with a sane response for when we find multiple bikes with a normalized serial number (don't think this case will be very likely, so it's the lowest priority right now).

There are some smaller, code cleanup questions that still apply. Mainly, is it possible to only override one method in the v3 implementation and then inherit the rest of the methods from the v2 module?

app/controllers/api/v3/bikes.rb
```diff
@@ -0,0 +1,300 @@
...
+            @bike = existing_bike
+            # Check organization ownership
+            authorize_bike_for_user
+            
+            # Additionally, check to ensure the email is attached to the bike (either as primary or secondary)
+            email_is_attached_to_bike_owner(b_param['params']['bike']['owner_email'])
+
+            begin
+              # We only want to update the Bike Attrs, which is a subset of the creation parameters
+              BikeUpdator.new(user: current_user, bike: existing_bike, b_params: BParam.v2_params(declared_p['declared_params'].slice(:bike_attrs).as_json)).update_available_attributes
```
stevenbuccini: One interesting thing is that things like color and manufacturer get stripped here, but they're still marked as required by the endpoint. Is that okay?

PR #546, opened 2019-02-12
More variable graphs

Robertgervais:


spec/controllers/admin/graphs_controller_spec.rb
```diff
@@ -73,10 +73,18 @@
           get :variable, kind: "payments", timezone: "America/Los_Angeles"
           expect(response.status).to eq(200)
           json_result = JSON.parse(response.body)
-          expect(json_result["error"]).to_not be_present
-          expect(json_result.keys.count).to be > 0
-          expect(json_result.keys.first.class).to eq(String)
-          expect(json_result.values.first.class).to eq(Integer)
+          expect(json_result[0]["error"]).to_not be_present
```
sethherr: rather than doing repeating all these specs for `json_result[0], json_result[1], json_result[2]` switch to 

```ruby
json_result.each do |data_group|
  expect(data_group.keys.count).to be > 0
  expect(data_group.keys.first.class).to eq(String)
  expect(data_group.values.first.class).to eq(String)
  expect(data_group["error"]).to_not be_present
end
```

Issue #545, opened 2019-02-11
Switch static models to be enums that don't touch the database


sethherr:
There are a number of models that aren't going to change (or will change very, very infrequently and slowly), and have a low number of instances:

- [ ] Color (largest, most difficult one)
- [x] CycleType #591 
- [x] FrameMaterial #552
- [x] HandlebarType #586 
- [x] PropulsionType #594
- [ ] Country (maybe)
- [ ] State (maybe)

Rather than using the database for these models, I think it would be better to move them entirely into the rails model file. We could maintain the same id's, so that the existing associations would be unchanged, and would basically make these enum fields.

PR #552, opened 2019-02-15
Refactor frame_materials

gadogado:
This addresses part of https://github.com/bikeindex/bike_index/issues/545.  The scope of the refactor would likely result in a large and long running branch/PR so if possible I'd like to break it up into smaller pieces respective to what's bulleted.

This PR focuses on restructuring `frame_materials` as an enum on `bike` instead of a separate relation.  Here's some highlights:

* Added a migration that batches an update_all on `bike`s that have an existing `frame_material` relation.  This updates the new enum column with whatever the existing relation was.
* Adds a decorator that normalizes the enum name  ( `"wood"` vs `"Wood or organic material"` ) 
* Add's an optional `frame_material` param to the bikes api endpoint

There were a lot of changes involved in this restructure and I had to make a few assumptions on the best approach so I'd appreciate any feedback or suggestions!  This should be reviewed locally if possible to catch any unintended side-effects from the substantial changes.


gadogado:
@sethherr thanks for spending time with this.  I just pushed up a fix for updating

spec/controllers/api/v1/bikes_controller_spec.rb
```diff
@@ -69,7 +69,7 @@
             rear_wheel_size: nil,
             rear_gear_type_slug: nil,
             handlebar_type_slug: nil,
-            frame_material_slug: nil,
```
sethherr: @gadogado this really, **really** shouldn't have happened. 

I'm sorry that I missed it

PR #553, opened 2019-02-18
Demonstrate potential bug in bike update endpoint

stevenbuccini:
Came across this case when working on another PR and wanted to throw it up to see if I'm missing something obvious or not using a `selection` correctly.

Sample request in documentation page running locally:
image

Response from documentation page running locally:
image

Also added a test to replicate the issue I faced.

### DIFF
--- a/spec/requests/api/v3/bikes_request_spec.rb
+++ b/spec/requests/api/v3/bikes_request_spec.rb
@@ -442,6 +442,19 @@ describe 'Bikes API V3' do
       expect(bike.owner).to eq(user)
       expect(bike.year).to eq(params[:year])
     end
+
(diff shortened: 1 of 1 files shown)


sethherr:
That is indeed a bug! Good catch.

stevenbuccini:
Good to have confirmation of this. This is blocking my implementation of #542. I propose fixing this bug in that PR by changing `BikeUpdator` so it takes in a `BParam` class rather than hash, as the object has coercion methods to fix bugs like this. However, it would be a pretty hefty change so I wanted to seek additional input.

sethherr:
Fixed by #715

PR #558, opened 2019-02-26
More variable graphs revisited

Robertgervais:


app/views/landing_pages/ambassadors_current.haml
```diff
@@ -127,4 +127,4 @@
         %h4
           What's your story?
         %p
-          Become a Bike Index Ambassador and help stolen bikes find their way home
\ No newline at end of file
+          Become a Bike Index Ambassador and help stolen bikes find their way home
```
Robertgervais: I never even opened this file! I don't know what's up with this.

app/controllers/admin/graphs_controller.rb
```diff
@@ -35,7 +35,10 @@ def variable
                         .count}
                     ]
       elsif @kind == "recoveries"
-        chart_data = StolenRecord.where(date_recovered: @start_at..@end_at)
+        recoveries = StolenRecord.unscoped
+        chart_data = recoveries.where(date_recovered: @start_at..@end_at)
+                         .group_by_period(@group_period, :date_recovered, time_zone: @timezone)
+                         .count
```
Robertgervais: I checked all the other graphs, they seem to working well enough, this is the only one I was getting an error on because... I did a really bad job. Seems to be working now.

PR #569, opened 2019-03-05
Revise feedbacks page

Robertgervais:


app/views/admin/feedbacks/index.html.haml
```diff
@@ -53,16 +51,20 @@
...
   = paginate @feedbacks
 
-= javascript_include_tag "//www.google.com/jsapi", "chartkick" 
+%h5.mt-5.mb-0.ml-5
+  Feedback and messages by date
+
+= javascript_include_tag "//www.google.com/jsapi", "chartkick"
 %pre
-  = params.to_yaml
+  -# Unsure of what this is. It renders as gobbledygook, and when I remove it nothing bad seems to happen?!?
```
sethherr: You can remove it. It was to help diagnose a problem and was never removed.

PR #576, opened 2019-03-06
Removes frame_material_id

gadogado:
### Overview
* Removes deprecated `frame_material_id` from `bike`


db/migrate/20190214192448_adds_frame_material_to_bikes.rb
```diff
@@ -4,7 +4,7 @@ def up
     add_column :bikes, :frame_material, :integer
 
     # bulk update enum before type cast - one statement per frame material
-    Deprecated::FrameMaterial.find_each do |fm|
+    defined?(Deprecated::FrameMaterial) && Deprecated::FrameMaterial.find_each do |fm|
```
sethherr: Good catch. I always forget to do stuff like this

PR #586, opened 2019-03-11
Handlebar type refactor

gadogado:


app/models/concerns/enumable.rb
```diff
@@ -0,0 +1,31 @@
+module Enumable
```
sethherr: Oooh. Slick, good plan

db/seeds/seed_bike_associations.rb
```diff
@@ -50,19 +50,6 @@
   color.save
 end
 
-# Seed the handlebar types
-handlebar_types = [
```
sethherr: Good catch.

spec/controllers/api/v1/bikes_controller_spec.rb
```diff
@@ -151,7 +150,7 @@
           frame_size_unit: nil,
           rear_gear_type_slug: rear_gear_type.slug,
           front_gear_type_slug: front_gear_type.slug,
-          handlebar_type_slug: handlebar_type.slug,
+          handlebar_type_slug: "other",
```
sethherr: Could you make this test something other than `other`? Or the other one, but that does seem to be testing that it fuzzy finds.

Not really that big of a deal, just seems like an easy way to add some extra coverage.

spec/controllers/bikes_controller_spec.rb
```diff
@@ -911,7 +911,7 @@
 
         it 'updates the bike and components' do
           component1 = FactoryBot.create(:component, bike: bike)
-          handlebar_type_id = FactoryBot.create(:handlebar_type).id
+          another_handlebar_type = "bmx"
```
sethherr: would you switch this to be something other than "bmx", so that `another_handlebar_type` is distinct?

spec/controllers/api/v1/bikes_controller_spec.rb
gadogado: Sure thing sounds good 👍

PR #591, opened 2019-03-14
Cycle type refactor

gadogado:
### Overview

* Reworks `CycleType` as an enum instead of a separate relation or string column

### DIFF
--- a/app/models/cycle_type.rb
+++ b/app/models/cycle_type.rb
@@ -1,9 +1,48 @@
-class CycleType < ActiveRecord::Base # Defines things like unicycles and recumbent
-  include FriendlySlugFindable
+class CycleType
+  include Enumable
 
-  has_many :bikes
+  SLUGS = %i[
+    bike
+    tandem
+    unicycle
+    tricycle
+    stroller
+    recumbent
+    trailer
+    wheelchair
+    cargo
+    tall-bike
+    penny-farthing
+    cargo-rear
+    cargo-trike
+    cargo-trike-rear
+    trail-behind
+    pedi-cab
+  ].freeze
 
-  def self.bike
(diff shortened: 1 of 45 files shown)


app/models/cycle_type.rb
```diff
@@ -1,9 +1,48 @@
-class CycleType < ActiveRecord::Base # Defines things like unicycles and recumbent
-  include FriendlySlugFindable
+class CycleType
+  include Enumable
 
-  has_many :bikes
+  SLUGS = {
```
sethherr: Is there a reasonable way you can think of combining this `SLUGS` hash with the `NAMES` hash?

gadogado: @sethherr tbh I'm not sure it would be worth the added complexity. It would resemble something like:

```

# CycleType:

  MAPPING = [
    { slug: :cargo, name: "Cargo Bike (front storage)", enum: 0 },
    ...

# Enumable:

  class_methods do
    %w( names slugs ).each do |kind|
      define_method(kind.to_sym) do
        key = kind.singularize.to_sym;
        self::MAPPING.map{ |m| m.fetch(key).to_s }
      end
    end

    def enums
      Hash[self::MAPPING.map { |m| [m[:slug], m[:enum]] }]
    end

    def select_options
      self::MAPPING.map { |m| [m[:name], m[:slug]] }
    end

    def legacy_selections
      self::MAPPING.map{ |m| m.slice(:name, :slug).stringify_keys }
    end

# Bike:

enum cycle_type: CycleType.enums
```

Down for whatever though!

sethherr: yeah. I can't think of anything super smooth either. Let's keep it the way it is 🙂

PR #595, opened 2019-03-14
Revise users index

Robertgervais:


sethherr:
This can't be merged until the dashboard index view - the other place that uses the admin users table partial - is updated to the new styles

Robertgervais:
Why is code climate so angry at me today. I can't even see any details with this one.

PR #605, opened 2019-03-16
add additional_registration to get bike api

laraferroni:
Just an add of additional_registration to the serializer


laraferroni:
I can add one, but I can't currently run the tests because I can't get the bundler working properly for some reason. I'm getting some weird dependency errors that I can't seem to resolve.

laraferroni:
got bundler going finally, so tests now pass

PR #609, opened 2019-03-19
Revise ctype views

Robertgervais:


app/views/admin/failed_bikes/show.html.haml
```diff
@@ -1,17 +1,17 @@
 %p
```
sethherr: Going forward, if you see things like this, put them into a `table-list`

PR #611, opened 2019-03-19
Revised failed bikes views

Robertgervais:


Robertgervais:
logoshot

app/views/admin/failed_bikes/show.html.haml
```diff
@@ -1,31 +1,31 @@
 %p
-  ID: 
+  ID:
```
sethherr: This looks like a `table-list` to me!

Robertgervais:
woah i don't even know how that got in here. Yeah, sorry! I revised this way before I started doing table-lists anywhere else. Will go back and revise!

PR #615, opened 2019-03-21
revise mail snippets views

Robertgervais:


Robertgervais:
mailsnippsedit
mailsnippsindex

PR #617, opened 2019-03-21
Revise memberships views

Robertgervais:


app/views/admin/memberships/show.html.haml
```diff
@@ -1,23 +1,21 @@
 %h1
   Showing membership
 
-.row
-  .span6.well.padded
-    %p
-      = link_to 'Edit', edit_admin_membership_url(@membership), class: 'button-blue pull-right'
```
sethherr: What about the edit button here?

Robertgervais: Replaced it!

PR #624, opened 2019-03-22
Fix kaminari

Robertgervais:


Robertgervais:
kaminari

app/views/admin/feedbacks/index.html.haml
```diff
@@ -11,7 +11,7 @@
     #{params[:type]}
 
 
-  = paginate @feedbacks, views_prefix: 'admin'
+  = paginate @feedbacks, views_prefix: 'admin', window: 3
```
sethherr: Why are you doing this here?

Robertgervais:
paginate

Robertgervais:
kaminari

Robertgervais:
kaminaribig
kaminarismall

app/views/admin/kaminari/_gap.html.haml
```diff
@@ -1,2 +1,2 @@
-%li.disabled.inline-item.align-bottom
-  = content_tag :a, raw(t 'views.pagination.truncate')
+-# %li.disabled.list-inline-item.align-bottom.pt-2
```
sethherr: I don't think you should comment these things out. I think you should remove them and remove the places that render them.

app/views/admin/kaminari/_next_page.html.haml
```diff
@@ -1,2 +1,6 @@
-%li.list-inline-item
+%li.list-inline-item.p-2.d-none.d-md-inline
```
sethherr: Rather than managing this via two different list items that are shown or hidden, you can do something like this:

```haml
= link_to_unless current_page.last?, url, :rel => 'next', :remote => remote, class: "btn btn-secondary", role: "button" do
  = "&rsaquo;"
  %span.d-none.d-md-inline
    = raw(t 'views.pagination.next')
```

Robertgervais:
kaminarifinal

PR #629, opened 2019-03-26
Revise Stolen Bikes views

Robertgervais:
stolenbikesindex
stolenbikesedit1
stolenbikesedit2


app/views/admin/stolen_bikes/edit.html.haml
```diff
@@ -1,110 +1,135 @@
...
-        %dl.dl-horizontal
-          = @bike.dl_list_item(@bike.created_at.strftime("%m-%d-%Y"), "Created")
-          = @bike.dl_from_attribute("owner_email")
-          = @bike.dl_list_item(@bike.owner && @bike.owner.email, "Current Owner")
-          - if @bike.creation_organization_id.present?
-            = @bike.dl_list_item(@bike.creation_organization.name, "Creation Organization")
-          %dt
+    %h4
+      Send bike owner an email
+    .form-group
```
sethherr: I think you should add some rows and columns here, at least on large screens.

app/views/admin/stolen_bikes/edit.html.haml
```diff
@@ -1,110 +1,135 @@
...
+      = f.text_area :body, placeholder: "You can write HTML in here. There will be a link to the #{@bike.type} included automatically.", rows: 7, class: "form-control w-25"
+    .row
+      .col-sm
+      .col-sm
+      .col-sm
+        = f.submit 'Send it!', class: 'btn btn-success'
+
+  %ul.bike-boxes
+    -# Should I create a new partial here?
+    = render partial: "/bikes/bike", locals: {bike: @bike, target: "show" }
```
sethherr: I believe the only place that is using the partial `/bikes/bike` is admin. Look around and make sure?

If that is the case, you should totally create a new partial - put it in `admin/bikes/bike` bonus points if you delete `/bikes/bike` and replace all the instances in admin with the new admin partial (even if the other places in admin don't look perfect because they aren't using the revised view)

app/views/admin/stolen_bikes/edit.html.haml
```diff
@@ -1,110 +1,135 @@
...
             %input{ disabled: true, type: 'text', value: edit_bike_recovery_url(bike_id: @bike.id, token: @stolen_record.find_or_create_recovery_link_token).to_s }
-          - if @stolen_record.approved
-            %h2
-              Stolen bike approved
-          - else
-            = link_to 'Approve listing', approve_admin_stolen_bike_url(@bike), method: :post, confirm: "Approving will:\n\n- Send out a tweet\n- The #{@bike.type} will be posted to any pertinent blogs.\n- It can NOT be effectively reversed\n", class: 'button-blue padded'
-        %dl.dl-horizontal
-        
-          %dt
+        %tr
```
sethherr: On large screens, I think this table should break into two tables, half screen width each

Robertgervais:
editmed
editsmall1
editsmall2
editlarge

app/views/admin/stolen_bikes/edit.html.haml
```diff
@@ -1,110 +1,143 @@
...
-        Stolen record not created yet.
-      %p
-        Wait a minute or two and it will be here!
+    = f.submit 'Send it!', class: 'btn btn-success mb-4'
 
-    %hr.padded
+  .row
+    .col-xs-12.col-lg-6
+      %ul.list-unstyled
+        -# Should I create a new partial here?
```
sethherr: Remove comment 😉

PR #635, opened 2019-03-28
Multiple serial search with React.  Upgrades webpack, webpacker, and adds Jest

gadogado:
## Overview
* Rewrite multiple serial search with latest React 
* Upgrades `webpack` to 4.x
* Upgrades `webpacker` to 4.x
* Adds Jest and respective specs for `multi_serial_search`
* Adds two new `circleci` steps to run `jest` and `eslint`


.circleci/config.yml
```diff
@@ -81,6 +81,17 @@ jobs:
                               --out test-results/rspec/rspec.xml \
                               --format progress \
                               -- ${TESTFILES}
+      - run:
+          name: "eslint"
+          command: "yarn lint"
+      - run:
+          name: "Jest: Install junit coverage"
```
sethherr: Questions about this and CI: 

- Does this integrate with Code Climate test coverage?
- Does a failure in `Jest: Tests` show up/alert CircleCI?

.eslintignore
```diff
@@ -0,0 +1 @@
+app/javascript/packs/jest
```
sethherr: Why not ignore all the packs? and node_modules?

.eslintrc
```diff
@@ -0,0 +1,26 @@
+
```
sethherr: Is there a way to use this config for prettier? Or switch to using prettier instead of eslint?

Why do we need eslint if we have prettier configured?

app/javascript/packs/multi_serial_search/api.js
```diff
@@ -0,0 +1,40 @@
+/*
+  Private
+*/
+
+const url = urn => (process.env.RAILS_ENV === 'development'
+  ? `http://localhost:3001/${urn}`
+  : `https://bikeindex.org/${urn}`);
+
+const serialSearchUrl = serial => url(
+  `api/v1/bikes?multi_serial_search=true&serial=${serial}`,
```
sethherr: Can you switch to v2 or v3 here?

app/javascript/packs/multi_serial_search/api.js
```diff
@@ -0,0 +1,40 @@
+/*
+  Private
+*/
+
+const url = urn => (process.env.RAILS_ENV === 'development'
```
sethherr: Should use `ENV["BASE_URL"]`

app/javascript/packs/utils/honeybadger.js
```diff
@@ -0,0 +1,16 @@
+import Honeybadger from 'honeybadger-js';
+import log from './log';
```
sethherr: Oh snap. This is a big improvement - setting the things we want in `utils/log.js` 💯

.circleci/config.yml
gadogado: Ya I followed the suggestions here to get the junit formatter setup for circle: https://circleci.com/docs/2.0/collect-test-data/#jest

Ya any Jest spec failure will prevent the build from passing

.eslintignore
gadogado: Right now the npm script ( package.json ) just runs eslint against `app/javascript/pack` but we want to ignore the jest config.

sethherr: 👍

app/javascript/packs/multi_serial_search/api.js
sethherr: Sorry for not making it clear that I wanted this!

.eslintrc
gadogado: Usually its good to have both imo.  Prettier is pretty great at guarding against style issues but doesn't really do much for the other things that are necessary: https://prettier.io/docs/en/comparison.html

app/javascript/packs/multi_serial_search/api.js
gadogado: No problem!  I'll swap it in

.eslintrc
sethherr: Makes sense! Thanks for the explanation

sethherr:
A few more things:

- [ ] I shouldn't see `Uncaught TypeError: BikeIndex.Init is not a constructor` in console
- [ ] I shouldn't see `[Honeybadger] Unable to send error report: no API key has been configured.` in console - it probably shouldn't be initialized in development?
- [ ] The `scroll-to-ref` behavior isn't around anymore - where when you click on one of the reference serials, it scrolls to the reference rather than jumping

gadogado:
@sethherr thanks for the feedback!  I'll get some changes in shortly

gadogado:
@sethherr re: `scroll-to-ref` - Is there some jquery scroll animation or something that maybe I missing from the original implementation? For whatever reason when I test with the **same query and apis** on local vs prod the behavior seems to be the same.  I can attach a gif if that helps?

Other than this I think everything should be good to go.  Getting there!

app/views/stolen/index.html.haml
```diff
@@ -83,14 +62,21 @@
               .contact-actions.control-group
                 = f.email_field :email, placeholder: "Your email address", class: 'form-control'
                 = f.submit 'Send alert!', class: 'btn pull-right'
+                
+        -# Stolen Bike Widget
+        %script{type: "text/javascript", src: "http://widget.bikeindex.org/stolen-widget.js"}
+        :javascript
```
sethherr: Nice solution!

PR #648, opened 2019-04-02
Revise failed bikes views

Robertgervais:


Robertgervais:
failedindex
failedshow

app/views/admin/failed_bikes/index.html.haml
```diff
@@ -28,10 +27,10 @@
             - if b_param.created_at.today?
```
sethherr: Should be replaced with updated date functionality

app/views/admin/failed_bikes/show.html.haml
```diff
@@ -1,30 +1,40 @@
...
+        Creator
+      %td
+        -if @b_param.creator.present?
+          = @b_param.creator.email
+        -else
+          No Creator
+    %tr
+      %td
+        created_at
+      %td.convertTime
```
sethherr: I think these should be `preciseTime`. 

In general, on show pages, things should show precise time because we don't need to conserve space.

Robertgervais: Got it. I'll go over this one and paints and see where this should be added.

sethherr: Great!

PR #650, opened 2019-04-02
Concerns::BikeSearchable: Refactored the searching funtionlity by introducing bike_searchable_service and also improved readability

usmanasif:
- Created separate BikeSearchableService to handle Search logic.
- Optimized code quality to avoid unnecessary memory operation and database calls.
- Improved code readability.
- Everything is refactored such that it wont break the functionality.

The feature can further be refactored after knowing the complete flow and requirements.

P.S There are some invalid test cases that are not mapped correctly with the desired flow. Some `query_params` are set for each test case but they often are used by the `expect` block in `rspec`.
The test cases also need to be refactored, updated and optimized. For now I just skipped a few test cases else working perfectly fine with the revamped optimized and readable codebase.


sethherr:
This has useful parts! When we approach this again we'll utilize things from this solution.

I've added an issue to track this, #673

PR #662, opened 2019-04-04
Revise orgs form, new, and edit views

Robertgervais:


Robertgervais:
orgs_edit
orgs_edit1
new_org

Robertgervais:
Screenshot from 2019-04-04 21-13-17

Whoops, forgot to add the prettier picture of the _location_fields. must have forgotten to copy and paste it over when I split up the PR's

Robertgervais:
All right, I'm thinking this is done. The button works now/renders an additional form. One question, the `%p` tag that renders the locations element, should that be dynamic at all? As is it is always showing 203, I guess? Since it's keyed into Country.united_states.id, just wondering!

app/views/admin/organizations/_form.html.haml
```diff
@@ -1,118 +1,88 @@
...
+          %label
+            Kind
+          %ul.list-unstyled
+            - Organization.kinds.each do |kind|
+              %li
+                = f.radio_button :kind, kind
+                = kind.humanize
+      .col-md-6
+        .form-group
+          = f.label :show_on_map
```
sethherr: I think, in this configuration, I'd prefer the checkboxes to be left aligned.

app/views/admin/organizations/_location_fields.html.haml
```diff
@@ -1,29 +1,24 @@
...
+        = f.text_field :street, placeholder: "Street", class: "form-control mt-1"
+      .form-group
+        = f.text_field :city, placeholder: "City", class: "form-control mt-1"
+        = f.collection_select(:state_id, State.all, :id, :name, {prompt: "State", include_blank: true}, class: "form-control mt-1" )
+      .form-group
+        = f.text_field :zipcode, placeholder: "Zipcode", class: "form-control mt-1"
+        = f.collection_select(:country_id, Country.all, :id, :name, {prompt: "Choose country"}, class: "form-control mt-1" )
+      .form-group
+        = f.check_box :_destroy
+        = f.label :_destroy, "Remove", class: "remove_fields"
```
sethherr: The checkbox for this used to be hidden, and this was styled. One of your other PRs removed the classes that set styles for it. You should add some styles back!

app/controllers/admin/organizations_controller.rb
```diff
@@ -36,10 +36,12 @@ def recover
 
   def new
     @organization = Organization.new
+    render layout: "new_admin"
   end
 
   def edit
     @embedable_email = @organization.auto_user.email if @organization.auto_user
+    render layout: "new_admin"
```
sethherr: The layout also needs to be added to the `create` and `update` actions, so if there is an error it still renders nicely

app/views/admin/organizations/_form.html.haml
```diff
@@ -1,118 +1,88 @@
...
         = f.label :available_invitation_count, "Available invitations"
-      .controls
-        = f.number_field :available_invitation_count
-      .controls.extra-field
+        = f.number_field :available_invitation_count, class: "form-control"
         %p
           = "Invitations sent: #{@organization.sent_invitation_count}"
+    .row
+      .col-md-6
+        .form-group.mt-4
```
sethherr: This feels like too much space to me

app/views/admin/organizations/_form.html.haml
```diff
@@ -1,118 +1,88 @@
...
-  #alert-block
-    .alert.alert-error
-      %h4
-        Please fix the following
-        = pluralize(@organization.errors.count, "error")
-      %ul
-        - @organization.errors.full_messages.each do |msg|
-          %li
-            = msg
+.card.bg-light
```
sethherr: Why did you decide to put this in a card?

app/views/admin/organizations/edit.html.haml
```diff
@@ -1,45 +1,42 @@
...
+      #new_bike_notification.collapse
+        .form-group
+        = f.label :new_bike_notification
+        = f.text_area :new_bike_notification, class: "form-control"
+    .col-md-2
       %p
         %strong
           Locations
-      %p#us-country-code.other-value
+      %p#us-country-code
```
sethherr: This was hidden on the old view, probably should be hidden again.

app/views/admin/organizations/_location_fields.html.haml
```diff
@@ -1,29 +1,24 @@
...
-    .form-group
-      = f.text_field :zipcode, placeholder: "Zipcode", class: "form-control mt-1"
-      = f.collection_select(:country_id, Country.all, :id, :name, {prompt: "Choose country"}, class: "form-control mt-1" )
-    .form-group
-      = f.check_box :_destroy
-      = f.label :_destroy, "Remove", class: "remove_fields"
-
-%hr
+      .form-group
+        = f.label :country_id, "Address"
```
sethherr: Address fields should render side by side, this location box takes up too much space

app/views/admin/organizations/edit.html.haml
```diff
@@ -1,45 +1,42 @@
...
-          = f.text_area :new_bike_notification
-
-      - if @organization.new_bike_notification.present?
-        .new-notification-container
-          .header-content
-            %img{src: "https://files.bikeindex.org/bikeindex-white-tilt.png"}
 
-          .new-notification
-            = @organization.new_bike_notification.html_safe
+  - if @organization.new_bike_notification.present?
```
sethherr: If the organization has a new bike notification (and a location), it is entirely separated from the form where they enter the new bike notification. They should be together.

app/javascript/packs/pages/binx_admin.js
```diff
@@ -65,4 +68,29 @@ window.BinxAdmin = class BinxAdmin {
...
+        const l = $(location);
+        if (!(l.val().length > 0)) { l.val(us_val); }
+      }
+      const names = $(this).closest('fieldset').find('.location-name-field input');
+      for (let name of Array.from(names)) {
+        const n = $(name);
+        if (!(n.val().length > 0)) { n.val($('#organization_name').val()); }
+      }
+
+      return $('.chosen-select select').select2();
```
sethherr: This throws an error because select2 isn't defined.

I'm opening another PR to add fancy select classes, but this isn't finished because this doesn't work

sethherr: Fancy selects were (just) added in #690 - I have already merged it in to this branch.

You don't need to have this return anything. All you need to do is replace this line with `LoadFancySelects()`

You'll also need to add the classes `.unfancy.fancy-select` to the form group (see [an example here](https://github.com/bikeindex/bike_index/pull/690/files#diff-acfdc28655d470fe581af313b3141c77R9)) for each of the selects that should be fancy (hint - any ones that used to have the class `.chosen-select`) - this goes for all future selects as well.

app/views/admin/organizations/_form.html.haml
```diff
@@ -1,118 +1,88 @@
...
+              = @organization.slug
+    .row
+      .col-md-6.pl-4
+        .form-group
+          %label
+            Kind
+          %ul.list-unstyled
+            - Organization.kinds.each do |kind|
+              %li
+                = f.radio_button :kind, kind
```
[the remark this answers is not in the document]
Robertgervais: UNADJUSTEDNONRAW_mini_b

So, this took me a lot longer than it should have because I'm a ding dong. Couldn't get those labels clickable, but I did it, and... without the list it looks like this. Do you want me to keep it like this or use a list? Or is there some other radio functionality that makes it a list without doing using li's?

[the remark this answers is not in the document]
Robertgervais: Image 4-18-19 at 2 42 PM

app/views/admin/organizations/_form.html.haml
Robertgervais: I've just been putting a lot of the forms in cards, but I guess in the particular case there are other fields that render. I can remove it if it makes you feel better. I just got into a habit of doing it.

Robertgervais:
This is why I'm supposed to make smaller pull requests.

app/views/admin/organizations/_form.html.haml
sethherr: It wasn't bad before, it just didn't seem like it was helping the flow of the page.

I think I like it more like this, let me know if you disagree 🙂

Robertgervais: No, I agree with you. I think on the pages where the form is this only thing that renders it makes sense to do it in a card, because it looks nice, but when this form renders there's possible a lot of other stuff rendering with it and it does get a little chunky.

app/views/admin/organizations/edit.html.haml
sethherr: It doesn't need to hide and appear - it actually never needs to be seen, it only needs to be accessed by the js. It should, in all correctly be made a `data-` property on one of the fields rather than text - but regardless - it definitely doesn't need to be seen by the user.

sethherr: What the attribute should actually be doing is making it so that new address blocks start, by default, with US as the country. Which isn't happening right now - but should

app/views/admin/organizations/edit.html.haml
sethherr: Actually, as discussed in https://github.com/bikeindex/bike_index/issues/705 - this should be removed from this form (and in a separate PR, removed from Bike Index)

PR #665, opened 2019-04-04
Minor changes to new_admin navbar

Robertgervais:


Robertgervais:
navsmall

app/javascript/packs/application_css.scss
```diff
@@ -10,13 +10,25 @@ footer.primary-footer {
...
   }
+  // I added this because several elements actually would show up above the navbar
+  z-index: 999;
 }
 
 .full-screen-table-wrapper {
   overflow-x: scroll;
 }
+
+@media screen and (min-width: 600px) {
```
sethherr: Rather than manually writing @media queries, I'd recommend using bootstraps [media query mixins](https://getbootstrap.com/docs/4.3/layout/overview/#responsive-breakpoints)

app/javascript/packs/application_css.scss
```diff
@@ -10,13 +10,25 @@ footer.primary-footer {
 .admin-navbar {
   .navbar-brand img {
     height: 35px;
-    width: 35px;
   }
   .dropdown-toggle {
     cursor: pointer;
   }
+  // I added this because several elements actually would show up above the navbar
+  z-index: 999;
```
sethherr: Firstly, if you are applying a zindex you use one of [Bootstraps variables](https://getbootstrap.com/docs/4.3/layout/overview/#z-index) for this value?

Secondly, looking at chrome's inspector, it seems like this value isn't the value that is applied, are you sure this is having an effect?

app/views/layouts/new_admin.html.haml
```diff
@@ -7,70 +7,63 @@
     = javascript_pack_tag "admin_application"
     = stylesheet_pack_tag "application_css"
 
+  -# To check for active page for dropdown menus
+  - paid_links = %w[payments paid_features invoices]
```
sethherr: I don't think you need these anymore

app/javascript/packs/application_css.scss
```diff
@@ -10,13 +10,25 @@ footer.primary-footer {
...
 }
 
 .full-screen-table-wrapper {
   overflow-x: scroll;
 }
+
+@include media-breakpoint-up(md) {
+  .sticky-admin-nav-md-screens {
```
sethherr: Why didn't you just apply this to `.admin-navbar` without adding an extra class?

Also, looking at the currently applied css from `.sticky-top` in chrome inspector, it looks like the styles are:

```
.sticky-top {
    position: -webkit-sticky;
    position: sticky;
    top: 0;
    z-index: 1020;
}
```

Is this what you see? If so, you should copy bootstrap 😉 

And now I understand why you were added zindex above! You should instead copy what bootstrap is doing (and use the appropriate variable instead of the number for zindex)

PR #667, opened 2019-04-05
Revise orgs invoice index

Robertgervais:


Robertgervais:
Image 4-15-19 at 4 45 PM

PR #668, opened 2019-04-05
Add recovered bikes page

### LINKED ISSUE #524: Add a recovered stolen bikes list page
Add a page that lists the recoveries, rather than just displaying them on the homepage. Paginate it.

Update the homepage to only display 5 and then have a link to the list page

jmromer:
- Updates the home page to display 5 recovery displays
- Adds a link to `/recovery_stories`
- Displays 5 displays at a time, paginated, at `/recovery_stories`

Resolves #524 

### Demo

demo

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -43,6 +43,7 @@ Bikeindex::Application.routes.draw do
   get 'choose_registration', to: 'welcome#choose_registration'
   get 'goodbye', to: 'welcome#goodbye'
   get 'bike_creation_graph', to: 'welcome#bike_creation_graph'
+  get "recovery_stories", to: "welcome#recovery_stories", as: :recovery_stories
 
   resource :session, only: [:new, :create, :destroy]
   get 'logout', to: 'sessions#destroy'
--- a/app/assets/javascripts/init.coffee
+++ b/app/assets/javascripts/init.coffee
@@ -26,6 +26,7 @@ class BikeIndex.Init extends BikeIndex
     # All the rest per-page javascripts
     pageClasses =
       welcome_index: BikeIndex.WelcomeIndex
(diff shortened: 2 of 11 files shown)


config/routes.rb
```diff
@@ -43,6 +43,7 @@
   get 'choose_registration', to: 'welcome#choose_registration'
   get 'goodbye', to: 'welcome#goodbye'
   get 'bike_creation_graph', to: 'welcome#bike_creation_graph'
+  get "recovery_stories", to: "welcome#recovery_stories", as: :recovery_stories
```
jmromer: Just idle curiosity:
Re. hyphens vs. underscores -- turns out that default I'd internalized at some point was [motivated by SEO](https://www.woorank.com/en/blog/underscores-in-urls-why-are-they-not-recommended). (I think I'd prioritize consistency with existing code anyway, though.)

> Google views hyphens in URLs as word separators while an underscore in your url will not be recognized. This means using hyphens makes it much easier for Google to figure out what a page is about. For example, a URL containing the phrase "my_page" would be interpreted as "mypage" instead of "my page".

sethherr: I went looking after our conversation and found that too!

I think this is a place where consistency trumps that, since the controller action (and the named route) uses underscores.

I'm open to figuring out how to put dashes in all the routes somehow, if it was convenient, since the google reference that site links to does say:

>  We recommend that you use hyphens (-) instead of underscores (_) in your URLs.

... though this seems like a low priority

sethherr:
After looking at [this live](https://bikeindex.org/recovery_stories) - it looks great, it's super exciting to be able to browse through them 🙂- I think it would make sense to have more than 5 per page, and maybe to put multiple per row?

PR #675, opened 2019-04-05
Fix broken seed data

hahmed:
I was trying to get this repo up and running and it looks as though the seed data is broken when I run the command `rake seed_test_users_and_bikes` I was getting the error `NoMethodError: undefined method `find_by_name' for CycleType:Class`.

Looking at the history `CycleType` was previous an AR, once it was changed to a PORO this broke.

This PR also fixes assigning `cycle_type_id` which is now `cycle_type`.


hahmed:
Actually I must have had an older branch tracked and realised it's actually been fixed :=)

sethherr:
Yeah. Just fixed 🙂

PR #678, opened 2019-04-08
Changes to recovery display form

Robertgervais:


Robertgervais:
Screen Shot 2019-04-08 at 6 49 32 PM

Robertgervais:
Okay, I ~think~ this works. In development I'm not getting any photos displayed with this, but also none of the photos in development have photos display. When i check in rails console they appear to have images attached in the same ways I see the ones in production have which is like "Recovery_#" Some things I'm wondering are, am I loading this JS in the right place? And also, should I be loading anything right when the page loads?

app/views/admin/recovery_displays/_form.html.haml
```diff
@@ -3,53 +3,63 @@
...
+            = f.label :date_recovered
+            - f.object.date_recovered ||= Time.now
+            = f.datetime_local_field :date_recovered, max: Time.now + 1.day, step: 60, required: true, class: 'form-control dateInputUpdateZone', "data-initialtime" => l(f.object.date_recovered, format: :convert_time)
+        .col-lg-6
+          - if @bike.present?
+            .form-group
+              %label
+                Bike url
+              %input.less-strong.w-100{ type: 'text', value: bike_url(@recovery_display.bike) }
+      .form-group.mt-4
```
sethherr: I think `form-group` should be nested _inside_ of the column, and the row should be the top level.

app/views/admin/recovery_displays/_form.html.haml
```diff
@@ -3,53 +3,63 @@
...
+        .row
+          .col-sm
+            = f.label :image, "Recovery photo"
+            - if @recovery_display.image?
+              = image_tag(@recovery_display.image_url(:thumb))
+            - else
+              %p
+                No picture!
+          .col-sm
+            .form-group
```
sethherr: There shouldn't be a `form-group` inside of a `form-group`

app/views/admin/recovery_displays/_form.html.haml
```diff
@@ -3,53 +3,63 @@
...
+              %label{:for => "recovery_upload"} Recovery Photo upload
+              %input#recovery_upload.form-control-file{:type => "file"}/
+      .form-group.avatar-upload
```
sethherr: Is this inside one of the columns?

Robertgervais:
Image 4-9-19 at 3 59 PM

PR #684, opened 2019-04-11
Admin Recoveries views fix - again

Robertgervais:


Robertgervais:
So when I "fixed" this earlier I thought that was it, but when I went to publish some recoveries nothing was working. I fixed the file upload button so now it definitely will upload files from your computer, but in development when I use the "use first image of bike" button I get one of our errors - Image trying to download a file which is not served over HTTP, which is a step up from before where I was getting nothing, and no image. I realized the hidden_field that was supposed to be getting set with JS was actually gone (couldn't find it in the inspector, searched the dom and also tried calling on it with the console) after the button was being clicked. I moved the hidden_field farther down the page, and now I'm hoping it's working but I'm unsure. Please and thanks.

app/views/admin/recovery_displays/_form.html.haml
```diff
@@ -59,12 +59,13 @@
                     #{link_to "Bike", bike_url(@bike), target: "_blank"} #{text}
                   %a.btn.btn-success#use_image_for_display{ href: '#', data: {url: url}}
                     Use first image
-                    = f.hidden_field :remote_image_url
```
sethherr: I think you need this field.

You'll need javascript to fill in the remote image url there

PR #688, opened 2019-04-15
Fix orgs show, children cell rendering problem

Robertgervais:


Robertgervais:
Image 4-15-19 at 5 53 PM

PR #692, opened 2019-04-17
Improve admin bikes tables

Robertgervais:
…nside partials, checked other views

### DIFF
--- a/app/views/admin/dashboard/maintenance.html.haml
+++ b/app/views/admin/dashboard/maintenance.html.haml
@@ -1,20 +1,20 @@
 %header.with-subtitle
   %h1 Maintaining things
-  
+
 %section.full-screen-table
   %header.with-subtitle
     %h2
       Recent example bikes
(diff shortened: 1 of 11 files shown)


Robertgervais:
I think it's all good to go - Need to make the users table responsive too because it doesn't smoosh well. I stopped to turn the bulk imports show into a table list since I didn't originally and I was like no that is NOT the task but otherwise all of the pages seem fine. Except for duplicates which I don't completely understand but it seems like the table's being modified there (on purpose). I can add screenshots if you'd like, but am wary of... PII?

Robertgervais:
Also - I took out .full-screen-table-wrapper since the idea was to have it be responsive, but does it NEED an outer container? Can add one if so.

app/views/admin/dashboard/maintenance.html.haml
```diff
@@ -79,12 +79,12 @@
             %td
               - if f.created_at.today?
```
sethherr: This looks like a special date displayer that could be replaced

Issue #175, opened 2016-04-27
Revised layout for admin pages


sethherr:
~Lowest priority. Shouldn't be done till absolutely everything else is~

sethherr:
At this point, this is something we want to do. The steps I think should be followed to take care of this:

1. Create a new partial layout with the admin header in it
2. Render that partial below the standard header in the revised layout and make the normal header not fixed if in the admin section (I think both should just stay at the top of the screen in admin)
3. Switch admin controllers gradually over to use the new layout, keeping things looking roughly similar to the way they look in admin right now.

Looking at the uncompiled, unminified scss files for admin might be useful. I figured out the commit where they were removed in #465 -

> The original style compilation files (pre removing updated_styles) are visible Style compilation files removal are visible in this commit: 1a8ebdf - PR #224

Robertgervais:
- [ ] SOMETHING blah blah

Robertgervais:
- [x] Ads
- [x] Ambassador Tasks Assignments
- [x] Ambassador Tasks
- [x] Bike Codes
- [x] Bikes
- [x] Bulk Imports
- [x] Ctypes
- [x] Dashboard
- [x] Exports
- [x] Failed bikes
- [x] Feedbacks
- [x] Graphs
- [] Mail Snippets
- [x] Manufacturers
- [x] Memberships
- [] News
- [x] Organization Invitations
- [] Organizations (Almost, just needs custom layouts done)
- [] Ownerships
- [x] Paid Features
- [x] Paints
- [x] Partial Bikes
- [] Payments
- [x] Recoveries
- [x] Recovery Displays
- [x] Stolen Bikes
- [x] Stolen Notifications
- [x] Tweets
- [x] Users

Robertgervais:
Hooooo yeah, this is DONE. I can either close this, or tomorrow I can go through and reference all the PR's that are involved in this.

PR #696, opened 2019-04-18
Revise orgs invoice edit

Robertgervais:


app/javascript/packs/pages/binx_admin.js
```diff
@@ -109,6 +113,48 @@ function BinxAdmin() {
         }
         return (location.href = `${current_url}${joiner}period=${period}`);
       });
+    },
+
+    initializeInvoiceForm() {
```
sethherr: I think I would like this to be put into a separate js file

app/views/admin/organizations/invoices/edit.html.haml
```diff
@@ -1,53 +1,61 @@
...
-    %ul
-      %li
-        Previous:
-        - if @invoice.previous_invoice.present?
-          #{link_to @invoice.previous_invoice.display_name, edit_admin_organization_invoice_path(organization_id: @invoice.organization.to_param, id: @invoice.previous_invoice.id)}
-      %li
-        Next:
-        - if @invoice.following_invoice.present?
-          #{link_to @invoice.following_invoice.display_name, edit_admin_organization_invoice_path(organization_id: @invoice.organization.to_param, id: @invoice.following_invoice.id)}
+%table.table-list
```
sethherr: Does this look really spread out on a big screen? Would it be better in multiple columns?

Robertgervais:
#175

PR #698, opened 2019-04-19
Revise custom layouts for orgs edit/form

Robertgervais:
…igure out why Sortable isn't being included.


Robertgervais:
Hey, so this is "edited" but I can't access the live version of this because my account isn't developer so I can't look at it. I'm not really sure what the page looks like with the JS, so I'm going to wait until I get that to finish this, unless it doesn't need any JS ported over.

PR #699, opened 2019-04-19
Revise admin dashboard

Robertgervais:


Robertgervais:
The only thing on this page I can see myself wanting to change is how the duplicate bikes are displayed but I don't know how important they are to you. I was thinking about having that section collapse, but I don't know when they get used.

Robertgervais:
Also, sorry, one other thing, was wondering if you wanted me to add something like a new orgs table? I doesn't seem dire, but it could be useful as orgs seem to be fairly important to the site right now.

sethherr:
I agree with your assessment

- "suspicious duped bikes" isn't used very often
- New organizations are important!

I think it would make sense to add a new organizations table, and to remove the duped bicycles - but you need to add duped bikes to the header navigation, since there isn't any other way to access them. 

Also, the duped bicycles display seems to have been broken at some point, I'm not sure when? We can discuss this (and what the duped bicycles page should look like) in a separate PR, where they are added.

Robertgervais:
So, I tried to turn the orgs table into a partial but I forgot about all of the sorting functionality and it seems like more of a headache than it's worth for like, ten orgs. If you want, I can do it, but I'd love to not.

app/views/admin/dashboard/index.html.haml
```diff
@@ -1,107 +1,49 @@
...
+
+%h2.mb-4.mt-4
+  Recent
+  = link_to "Bikes", admin_bikes_url
+
+= render partial: "/admin/bikes/admin_bikes_table", locals: { bikes: @bikes }
+
+
+.row
+  .col-md-7
```
sethherr: Why are these columns here?

Robertgervais:
#175

PR #700, opened 2019-04-19
Rework initialize

gadogado:
### Overview

* Use factory functions instead of class when there's no constructor
* Use a default export for classes and factory fn's
* Use es6 methods instead of underscore/lodash when possible
* Removes `log` when declared but never used
* Rework classes to use `this` for methods instead of the intermediate window object
* Reworks some `BinxOrgMessages` methods to use state implicitly within closure:
  * NOTE: I couldn't directly test some of the functionality here since the google maps key is over quota. Some of these changes are maybe a good first step for when you're ready to refactor but will require more investigation, testing, etc before merging.


app/javascript/packs/pages/binx_admin.js
```diff
@@ -2,98 +2,120 @@ import log from "../utils/log";
...
-      $(".calendar-box").slideDown();
-    } else {
-      $(".calendar-box").slideUp();
-      this.startGraphTimeSet();
-    }
-  }
+    startGraphTimeSet() {
+      let graphSelected = $("select#graph_date_option")[0].value.split(",");
+      let amount = Number(graphSelected[0]);
+      let unit = graphSelected[1];
```
jmromer: Any objection to defaulting to `const`?

sethherr: Is there any difference? Can you mutate `const`?

jmromer: That's the main difference—you can't re-assign `const` (so the binding is immutable, but any non-primitive object bound to the variable can still be mutated). Scoping is the same. `const` provides marginally more assurance about state, but...the emphasis there is on _marginal_, so I don't feel too strongly about it.

sethherr: I’m happy to default to const, seems good.

sethherr: I switched these to `const`, but didn't switch the `for (let name of Array.from(names))` - which might be rewriteable in some other way, but I worry about reassigning the variable? Let me know if you think I should switch those too.

jmromer: > for (let name of Array.from(names))

I wasn't sure about this myself, but it looks like `const` work as expected in for-of / for-in loops but not in imperative for loops:

```js
> arr = [1,2,3,4]

> for (let num in arr) { console.log(num) }
0
1
2
3

> for (const num in arr) { console.log(num) }
0
1
2
3

> for (const i in arr) { console.log(arr[i]) }
1
2
3
4

> for (const i = 0; i < arr.length; i++) { console.log(arr[i]) }
1
TypeError: Assignment to constant variable.
```

sethherr: Yeah. apparently because ++ is reassignment. I'm happy with how it is now, since `const` as an iterator seems weird to me - but I'm happy to swap it out if you think we should use exclusively `const`

PR #703, opened 2019-04-20
Revise admin tsvs

Robertgervais:


Robertgervais:
Image 5-16-19 at 2 16 PM

Robertgervais:
It's a little sparse but it's supposed to FUNCTIONAL not fun. It also has a vertically scrollable table when `@tsvs` are present

Robertgervais:
#175

PR #704, opened 2019-04-22
Update UIs for Recovery Stories

jmromer:
- Moves the "read more recovery stories" link to the end of the slideshow
- Doubles the number of recovery stories displayed
- Displays recovery stories in two columns (past "xl" width breakpoint)
- Adds the recovery stories page to the sitemap

### Demos

Full width

demo

Narrow

demo2

Implements suggestions from https://github.com/bikeindex/bike_index/pull/668

### DIFF
--- a/app/views/welcome/recovery_stories.html.haml
+++ b/app/views/welcome/recovery_stories.html.haml
@@ -5,10 +5,10 @@
 
   #recovery-stories-listing
     .container.clearfix
-      .recovery-stories-container{data: {stories_count: @left.length}}
-        = render @left
-      .recovery-stories-container{data: {stories_count: @right.length}}
-        = render @right
+      .recovery-stories-container{data: {stories_count: @slice1.length}}
+        = render @slice1
+      .recovery-stories-container{data: {stories_count: @slice2.length}}
+        = render @slice2
 
     .paginate-container.paginate-container-bottom
       = paginate @recovery_displays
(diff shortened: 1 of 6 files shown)


config/sitemap.rb
```diff
@@ -53,7 +53,12 @@
   end
 
   group(filename: :contact) do
-    paths = ['/help']
+    paths = ["help"]
+    paths.each { |i| add "/#{i}", priority: 0.8 }
```
sethherr: Good catch!

app/views/welcome/recovery_stories.html.haml
```diff
@@ -4,9 +4,11 @@
       Recovery Stories
 
   #recovery-stories-listing
-    .container
-      #recovery-stories-container{data: {stories_count: @recovery_displays.length}}
-        = render @recovery_displays
+    .container.clearfix
+      .recovery-stories-container{data: {stories_count: @slice1.length}}
```
sethherr: nit: I have a slight preference for making these a quoted key value pair (`"stories-count" => ...`), because the transformation of property names confuses me. Doesn't need to be changed, just a feeling I thought I'd burden you with 🙂

jmromer: no prob thanks -- I may have caught a bug in some more testing just now so I'll take another look at these

PR #706, opened 2019-04-22
Fix bug in Slick initialization

jmromer:
Currently `slick` isn't being initialized on recovery stories containers, so images aren't loading. This patch fixes that by correcting the dom query string to `.recovery-stories-container`. 

NB: There's an issue where the height of a container isn't computed correctly by slick if elements within it are missing images, but all the recovery stories on prod seem to have images, so this wouldn't be an issue in practice.

demo

### DIFF
--- a/app/assets/javascripts/revised/pages/welcome/recovery_stories.coffee
+++ b/app/assets/javascripts/revised/pages/welcome/recovery_stories.coffee
@@ -1,9 +1,10 @@
 class BikeIndex.RecoveryStories extends BikeIndex
   constructor: ->
-    $recovery_stories_containers = $('#recovery-stories-container').find('.recovery-stories-container')
-    for $recovery_stories_container in $recovery_stories_containers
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/revised/pages/welcome/recovery_stories.coffee
```diff
@@ -1,9 +1,10 @@
 class BikeIndex.RecoveryStories extends BikeIndex
   constructor: ->
-    $recovery_stories_containers = $('#recovery-stories-container').find('.recovery-stories-container')
```
jmromer: (This query string is incorrect)

sethherr: Just to make sure - it's now the correct query string, with the class, and this is ready?

jmromer: Yup, thanks—this is the old line. It should be `.recovery-stories-container`.

PR #710, opened 2019-04-23
Fix recovery stories for mobile clients / web crawlers

jmromer:
- Handles queries for an invalid page more gracefully than raising an exception:  https://app.honeybadger.io/projects/35931/faults/48555100
- Fixes mobile view by disabling image lazy-loading on recovery stories (it breaks the layout on the recovery stories page, and we're only loading a maximum of 5 images on the root url)
- Display fewer pagination links to account for smaller screen width at mobile breakpoints

### Before

demo

### After

demo2

Screen Shot 2019-04-23 at 11 45 15 AM


app/controllers/welcome_controller.rb
```diff
@@ -56,6 +56,8 @@ def authenticate_user_for_welcome_controller
   # Split the given array `list` into two halves
   # Return a tuple with each half as an array.
   def list_halves(list)
+    return [[], []] if list.empty?
```
sethherr: Would you add a test for the bug?

app/views/welcome/recovery_stories.html.haml
```diff
@@ -11,4 +11,4 @@
         = render @slice2
 
     .paginate-container.paginate-container-bottom
-      = paginate @recovery_displays
+      = paginate @recovery_displays, window: 0
```
sethherr: Nice! This is an elegant solution, and a good usability fix

app/controllers/welcome_controller.rb
jmromer: Added -- https://github.com/bikeindex/bike_index/pull/710/commits/d04c8bb61d3c269da6796e5b019ea4a0d024b165#diff-50d29b5fa2c03989d9bc2490d16e6b9bR156

PR #711, opened 2019-04-23
Consistify whitespace, code style in Ruby, CoffeeScript, SCSS

### LINKED ISSUE #674: Add Rufo for ruby formatting
I _strongly_ agree with this statement from @patio11 in [What Working At Stripe Has Been Like](https://www.kalzumeus.com/2019/3/18/two-years-at-stripe/)

>  the right way to do all stylistic arguments is to enforce them with code, and make the sole supported interface for stylistic arguments pull requests against the linter/compiler rather than comments on individual engineers’ work

To that end, remove rubocop and replace it with [rufo](https://github.com/ruby-formatter/rufo).

This will involve adding a `.rufo` config file, which should probably be modeled after our rubocop setup.

jmromer:
- Runs rufo on project ruby (NB: the `.rufo` config file is being [deprecated](https://github.com/ruby-formatter/rufo/issues/2))
- Autoformats SCSS with `prettier`
- Kills trailing whitespace in coffeescript with sed
- Adds an .editorconfig to automatically consistify whitespace

@sethherr The diff is mostly quotes and whitespace changes, but if any of those commits seem like overkill or too much to pack into a single PR, I'm happy to remove them

Fixes #674


jmromer:
> * We have a `.eslintrc` and a `.rubocop.yaml` - do these matter anymore?

I think `.eslintrc` is worth keeping for JS files, unless we want to add something like [`prettier-standard`](https://github.com/sheerun/prettier-standard), which is effectively the most opinionated autoformatter for JS I know of.

Rubocop has checks that go beyond rufo's intended purpose (e.g., for Rails suggested conventions), so we may want to keep it around and continue using it as a linter. I can add some config rules to bring it in closer in line with rufo's settings. I think this would be as simple as disabling style cops:

https://github.com/ruby-formatter/rufo/blob/14802ef1881acf6ac116368ed357918cd1014262/.rubocop.yml#L7-L14

jmromer:
> Could you add something to the readme explaining what is up with code formatting, or would you prefer to add that to a separate PR? Two specific topics (though I think more could be covered):

Will do! Forgot it requires some setup, thanks for the head's up

jmromer:
@sethherr Rendered README update: https://github.com/bikeindex/bike_index/blob/e5f910439df169840f806db155ad338444c80ce2/README.markdown

NB: Looking a bit more closely, I see that [`prettier`](https://prettier.io/) has Ruby support, which is interesting--potentially we could just use the one tool for JS, CoffeeScript (maybe ?), Sass, and Ruby. My impression is that it's faster than Rubocop but slower than Rufo. (We probably don't want to go too far down a rabbit hole evaluating formatting tools though.)

sethherr:
The idea of having a single formatter is definitely appealing - particularly  if it means that all the style config would be in one file (which it looks like it would be?)

It also seems like [prettier/plugin-ruby](https://github.com/prettier/plugin-ruby) is under more active development than [rufo](https://github.com/ruby-formatter/rufo) and has more stars.

So... give it a try if you want to?

jmromer:
> So... give it a try if you want to?

No prob! I'll take a look.

sethherr:
Ok. Is this ready to merge @jmromer ?

jmromer:
@sethherr All set! 👍

Issue #548, opened 2019-02-13
Add pos_ascend description for ascend bikes


sethherr:
In `CreationState` there is an attribute `is_pos` (AKA _is registered at point-of-sale_) for bulk imports (of ascend type).

This update should:

- Mark creation states `is_pos` true (during creation) if they come from ascend bulk imports
- Switch the existing `creation_description` "pos" to be "Lightspeed POS" and make ascend creation states return "Ascend POS"

sethherr:
Some context: 

We have another app that integrates with the main bike shop point of sale (POS) service, Lightspeed. For shops that have integrated with us, via that secondary app, we automatically register the bikes they sell as soon as they sell them.

We recently added an [ascend integration](https://bikeindex.org/news/bike-index-partners-with-ascend-retail-to-offer-pos-registration-to-th) - ascend is another point of sale service that bike shops use.

We like to keep track of bikes that were registered at point of sale because:

1. That is considered the best source of authority. If your bike was registered automatically when it was purchased, we can be very sure it was registered correctly
2. We want to make sure the integrations are operating correctly!
3. We want to see how many bikes are registered by integration
4. We want to give users (and bike shops) visibility into which bikes were registered automatically, as opposed to manually.

---

We track whether bikes were registered at Point of Sale via the `is_pos` attribute on `CreationState` 

Right now, Bikes that are registered via our Ascend integration aren't marked as `is_pos` - but they should be!

You can tell bikes were registered via our Ascend integration because they came in through a bulk import for which `ascend?` is truthy

Once we're successfully marking bikes registered via our Ascend integration `is_pos`, we'll need to differentiate between Ascend integration bikes and Lightspeed integration bikes - by changing the `CreationState` description

sethherr:
Closed by #712

PR #712, opened 2019-04-23
Improve Admin Ascend viewing

sethherr:
I know this will cause some merge conflicts with the style update, but  ¯\\_(ツ)_/¯

I needed some extra tools for looking at Ascend imports.

Updated view:

Screen Shot 2019-04-23 at 4 17 03 PM

I also figured we needed a standardized admin subheader, so I've added that.


sethherr:
... and we need to be able to view ascend bikes, and lightspeed bikes. Adding some new stuff to the `admin/bikes` header so that we can sort them (making use of the new `admin-subnav` class :wink:)

Screen Shot 2019-04-23 at 5 11 30 PM

Fixes #548 (when complete)

sethherr:
After shipping, run this in console:

```ruby
CreationState.find_each { |c| c.update_attributes(updated_at: Time.now) }
```

app/models/bike.rb
```diff
@@ -149,6 +150,16 @@ def get_listing_order
     stock_photo_url.present? || public_images.present? ? t : t/100
   end
 
+  def creation_state; creation_states.first end
+
+  def creation_description; creation_state&.creation_description end
+  
+  def bulk_import; creation_state&.bulk_import end
+
+  def pos_kind; creation_state&.pos_kind end
```
jmromer: Suggestion / Question — [delegate](https://apidock.com/rails/Module/delegate) provides a succinct way to define this behavior / Should we prefer the longhand form?

sethherr: Yeah. This changed it from being `delegate` - I couldn't get the join working when it was `has_one`, and I couldn't get delegate working without a relation.

There is probably a way to fix either of those, and I'd be happy to use it! I'd prefer delegate for sure. I just didn't want to mess with it too much.

jmromer: Ah gotcha. Just checking, thanks 👍

PR #715, opened 2019-04-24
Ensure bike creation API endpoint upserts

### LINKED ISSUE #517: Dedupped registration API endpoint
Registration through the API for bikes that may have been registered already - only register if the bike hasn't been registered yet.

Potentially add any information that wasn't filled in for the bike already.

This will be useful for recording the serial number of bikes that are being worked on in shops

jmromer:
When creating a bike, first [search for the record](https://github.com/bikeindex/bike_index/pull/715/files#diff-2a964c740a025559cefad2c1348689ba) being created by serial number and email address (tests [here](https://github.com/bikeindex/bike_index/pull/715/files#diff-9aa7bec756e779b03037858befd4f7ff)):

```rb
# app/controllers/api/v2/bikes.rb L140-143 (d171f69d)

post "/", serializer: BikeV2ShowSerializer, root: "bike" do
  # Search for a bike matching the provided serial number / owner email
  bike_attrs = %w{serial owner_email}.map { |key| [key.to_sym, params[key]] }.to_h
  found_bike = BikeFinder.find_matching(**bike_attrs)
```

<details>

<summary>BikeFinder generated SQL</summary>

```sql
-- BikeFinder.find_matching(serial: "serial-1", owner_email: "user@example.com")
-- Bike Load (12.4ms)

SELECT
  "bikes".* 
FROM
  "bikes" 
  RIGHT JOIN
    ownerships 
    ON bikes.id = ownerships.bike_id 
WHERE
  "bikes"."example" = false
  AND "bikes"."hidden" = false 
  AND "bikes"."serial_normalized" = 'serial-1'
  AND 
  (
    bikes.owner_email = 'user@example.com' 
    OR ownerships.user_id IN 
    (
      SELECT DISTINCT
        "users"."id" 
      FROM
        "users" 
        LEFT JOIN
          user_emails 
          ON user_emails.user_id = users.id 
      WHERE
        (
          users.email = 'user@example.com' 
          OR user_emails.email = 'user@example.com'
        )
    )
    OR ownerships.creator_id IN 
    (
      SELECT DISTINCT
        "users"."id" 
      FROM
        "users" 
        LEFT JOIN
          user_emails 
          ON user_emails.user_id = users.id 
      WHERE
        (
          users.email = 'user@example.com' 
          OR user_emails.email = 'user@example.com'
        )
    )
  )
ORDER BY
  listing_order desc LIMIT 1 

```
</details>

If the bike is not found, proceed as we currently do, creating the bike record. 

If it is found, update instead of creating:

```rb
# app/controllers/api/v2/bikes.rb L164-190 (795dbb2d)

  # bike was found, update instead of creating
  if found_bike.present?
    # prepare params
    declared_p = { "declared_params" => declared(params, include_missing: false) }
    b_param = BParam.new(creator_id: creation_user_id, params: declared_p["declared_params"].as_json, origin: "api_v2")
    b_param.clean_params
    ensure_required_stolen_attrs(b_param.params)

    @bike = found_bike
    authorize_bike_for_user

    if b_param.params.dig("bike", "external_image_urls").present?
      @bike.load_external_images(b_param.params["bike"]["external_image_urls"])
    end

    begin
      BikeUpdator
        .new(user: current_user, bike: @bike, b_params: b_param.params)
        .update_available_attributes
    rescue => e
      error!("Unable to update bike: #{e}", 401)
    end

    status :found
    return @bike.reload
  end
end
```

Tests: [spec/requests/api/v3/bikes_request_spec.rb](https://github.com/bikeindex/bike_index/pull/715/files#diff-bcd0e37d08fb96ca09bc0c02dbc423f2)

Fixes #517 
Related: https://github.com/bikeindex/bike_index/pull/542

### DIFF
--- a/app/services/bike_finder.rb
+++ b/app/services/bike_finder.rb
@@ -11,10 +11,12 @@ module BikeFinder
   #
   # Return a Bike object, or nil
   def self.find_matching(serial:, owner_email:)
+    email = EmailNormalizer.normalize(owner_email)
+
     candidate_user_ids =
       User
         .joins("LEFT JOIN user_emails ON user_emails.user_id = users.id")
-        .where("users.email = ? OR user_emails.email = ?", owner_email, owner_email)
+        .where("users.email = ? OR user_emails.email = ?", email, email)
         .select(:id)
         .uniq
 
--- a/app/services/bike_finder.rb
+++ b/app/services/bike_finder.rb
@@ -1,0 +1,32 @@
+module BikeFinder
+  # Find a bike with the given serial number `serial` associated with email
+  # address `owner_email`.
+  #
+  # Matches based on the normalized serial number, and `owner_email` should be
+  # found via:
+  #
+  # - the bike's owner_email attribute
+  # - the email attribute of any of the bike's owner's (user / creator: via bike.ownerships)
+  # - any email associated with any owner (via user.user_emails)
+  #
+  # Return a Bike object, or nil
+  def self.find_matching(serial:, owner_email:)
+    candidate_user_ids =
+      User
+        .joins("LEFT JOIN user_emails ON user_emails.user_id = users.id")
+        .where("users.email = ? OR user_emails.email = ?", owner_email, owner_email)
+        .select(:id)
+        .uniq
+
+    Bike
+      .joins("LEFT JOIN ownerships ON bikes.id = ownerships.bike_id")
+      .where(serial_normalized: SerialNormalizer.new(serial: serial).normalized)
+      .where(
+        "bikes.owner_email = ? OR ownerships.user_id IN (?) OR ownerships.creator_id IN (?)",
+        owner_email,
+        candidate_user_ids,
+        candidate_user_ids
+      )
+      .first
+  end
+end
--- a/spec/models/b_param_spec.rb
+++ b/spec/models/b_param_spec.rb
@@ -77,10 +77,7 @@ describe BParam do
       b_param.clean_params
       clean_params2 = b_param.params
 
-      clean_params2["bike"]
-        .sort
-        .zip(clean_params1["bike"].sort)
-        .each { |(k1, v1), (_, v2)| expect(v1).to eq(v2), "#{k1} : #{v1} != #{v2}" }
+      expect_hashes_to_match(clean_params2["bike"], clean_params1["bike"])
       expect(clean_params2["bike"].keys).to match_array(clean_params1["bike"].keys)
       expect(clean_params2).to eq(clean_params1)
     end
--- a/spec/models/b_param_spec.rb
+++ b/spec/models/b_param_spec.rb
@@ -49,6 +49,61 @@ describe BParam do
     it "has before_save_callback_method of clean_params" do
       expect(BParam._save_callbacks.select { |cb| cb.kind.eql?(:before) }.map(&:raw_filter).include?(:clean_params)).to eq(true)
     end
+    context "run multiple times" do
+      let(:initial_params) do
+        { "serial" => "69 non-example",
+          "manufacturer" => "Special_name3",
+          "owner_email" => "fun_times@examples.com",
+          "color" => "Special_name4",
+          "cycle_type_name" => "bike",
+          "rear_wheel_bsd" => 559,
+          "rear_tire_narrow" => true,
+          "year" => 1969,
+          "frame_material" => "steel",
+          "is_bulk" => nil,
+          "is_pos" => nil,
+          "is_new" => nil }
+      end
+      let(:target_params) do
+        { "bike" => {
+          "serial_number" => "69 non-example",
+          "manufacturer_id" => manufacturer.id,
+          "owner_email" => "fun_times@examples.com",
+          "primary_frame_color_id" => color.id,
+          "cycle_type" => :bike,
+          "front_wheel_size_id" => nil,
+          "rear_wheel_size_id" => wheel_size.id,
+          "rear_tire_narrow" => true,
+          "handlebar_type" => nil,
+          "year" => 1969,
+          "frame_material" => "steel",
+          "is_bulk" => nil,
+          "is_pos" => nil,
+          "is_new" => nil,
+          "send_email" => true,
+        },
+          "test" => nil,
+          "id" => nil,
+          "components" => nil }
+      end
+      let(:b_param) { BParam.new(params: initial_params, origin: "api_v2") }
+      let!(:wheel_size) { FactoryBot.create(:wheel_size, iso_bsd: 559) }
+      let!(:color) { FactoryBot.create(:color, name: "Special_name4") }
+      let!(:manufacturer) { FactoryBot.create(:manufacturer, name: "Special name3") }
+      it "returns the same thing" do
+        b_param.clean_params
+        expect(b_param.params.keys).to eq target_params.keys
+        expect(b_param.params["bike"].keys).to match_array target_params["bike"].keys
+        expect(b_param.params).to eq target_params
+        b_param.clean_params
+        expect(b_param.params["bike"].keys).to match_array target_params["bike"].keys
+        b_param.params["bike"].keys.each do |k|
+          pp k
+          expect(b_param.params["bike"][k]).to eq target_params["bike"][k]
+        end
+        expect(b_param.params).to eq target_params
+      end
+    end
   end
 
   describe "massage_if_v2" do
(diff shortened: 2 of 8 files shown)


jmromer:
@sethherr Re [this comment](https://github.com/bikeindex/bike_index/issues/517#issuecomment-463732224): 

> If a registration is posted without a serial number, this updated endpoint should only register the bike if there aren't any bikes by the same manufacturer (regardless of what their serial numbers are)

If that's still a requirement, can you elaborate on that? So for any given manufacturer, there should be at most one bike record with no serial number?

app/services/bike_finder.rb
```diff
@@ -0,0 +1,32 @@
...
+  #
+  # Matches based on the normalized serial number, and `owner_email` should be
+  # found via:
+  #
+  # - the bike's owner_email attribute
+  # - the email attribute of any of the bike's owner's (user / creator: via bike.ownerships)
+  # - any email associated with any owner (via user.user_emails)
+  #
+  # Return a Bike object, or nil
+  def self.find_matching(serial:, owner_email:)
```
sethherr: Questions about this class:

- Why not use `User.friendly_id_find`?
- What about bikes with absent serials?

spec/models/b_param_spec.rb
```diff
@@ -46,9 +46,44 @@
...
+      b_param.clean_params
+      clean_params1 = b_param.params
+
+      b_param.clean_params
+      clean_params2 = b_param.params
+
+      clean_params2["bike"]
+        .sort
+        .zip(clean_params1["bike"].sort)
+        .each { |(k1, v1), (_, v2)| expect(v1).to eq(v2), "#{k1} : #{v1} != #{v2}" }
```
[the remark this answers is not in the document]
jmromer: Ah nice! Just curious, have you tried https://github.com/vitalish/rspec-deep-matchers
(Doesn't look maintained though.)

app/services/bike_finder.rb
jmromer: > Why not use User.friendly_id_find?

I was trying to minimize db calls, but `friendly_id_find` will trigger a query and instantiate a model. That said, I didn't realize we had an `EmailNormalizer`, so I'll add that in here — it'll still be a single query, but it'll match more robustly. Good call-out, thanks.

> What about bikes with absent serials?

Discussed offline: I'll make an issue for handling the `absent` serial corner cases 👍

sethherr: Is there a way to move this sequel logic into `User.friendly_id_find`?

jmromer: Overall performance wouldn't seem to be impacted too much:
```
# Bike Load (47.6ms)  SELECT  "bikes".* FROM "bikes" LEFT JOIN ownerships ON bikes.id = ownerships.bike_id WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."serial_normalized" = $3 AND (bikes.owner_email = 'elvis8tcb@yahoo.com' OR ownerships.user_id IN (SELECT DISTINCT "users"."id" FROM "users" LEFT JOIN user_emails ON user_emails.user_id = users.id WHERE (users.email = 'elvis8tcb@yahoo.com' OR user_emails.email = 'elvis8tcb@yahoo.com')) OR ownerships.creator_id IN (SELECT DISTINCT "users"."id" FROM "users" LEFT JOIN user_emails ON user_emails.user_id = users.id WHERE (users.email = 'elvis8tcb@yahoo.com' OR user_emails.email = 'elvis8tcb@yahoo.com')))  ORDER BY listing_order desc LIMIT 1  [["example", "f"], ["hidden", "f"], ["serial_normalized", "F6110677"]]
```
vs. 
```
# UserEmail Load (0.2ms)  SELECT  "user_emails".* FROM "user_emails" WHERE (confirmation_token IS NULL) AND "user_emails"."email" = $1 LIMIT 1  [["email", "elvis8tcb@yahoo.com"]]
# User Load (0.5ms)  SELECT  "users".* FROM "users" WHERE "users"."id" = $1 LIMIT 1  [["id", 145]]
# Bike Load (48.5ms)  SELECT  "bikes".* FROM "bikes" LEFT JOIN ownerships ON bikes.id = ownerships.bike_id WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."serial_normalized" = $3 AND (bikes.owner_email = 'elvis8tcb@yahoo.com' OR ownerships.user_id IN (145) OR ownerships.creator_id IN (145))  ORDER BY listing_order desc LIMIT 1  [["example", "f"], ["hidden", "f"], ["serial_normalized", "F6110677"]]
```

But:
(1) the result set might be different (`friendly_id_find` only queries in `UserEmail`, not `User`, which has its own `email` column, and restricts results to confirmed emails)
(2) `friendly_id_find` and the methods it depends on would have to be refactored to return an `ActiveRecord::Relation` (rather than an Integer or nil)

sethherr: Shouldn’t both methods search the same thing?

jmromer: > Shouldn’t both methods search/return the same thing?

I can't speak to the intent of `User.friendly_id_find`, but I would think that for the purposes of upserting, we would not want to restrict ourselves to confirmed emails, no? 

(i.e., I'm not sure I see why we would only want to upsert for bikes with an associated user that has a confirmed email—and create duplicates for others—given that we want to upsert for bikes with no associated user yet. Let me know if there's a faulty assumption in there though.)

There's also the `users.email` column, which contains emails not found in `user_emails.emails`.

sethherr: I think the `users.emails` column is guaranteed to be covered by user_emails emails. 

But I think your logic is correct - re: unconfirmed emails.

So you can ignore this

jmromer: I thought there might be application logic enforcing that duplication (in general I'm a little wary of having application logic enforce database constraints), but I'm seeing a significant number of emails on `users` that aren't on `user_emails`:

```rb
User.connection.execute("SELECT COUNT(*) FROM users AS u WHERE u.email NOT IN (SELECT ue.email from user_emails AS ue WHERE ue.user_id = u.id)").to_a.first
# (174.3ms)  SELECT COUNT(*) FROM users AS u WHERE u.email NOT IN (SELECT ue.email from user_emails AS ue WHERE ue.user_id = u.id)
# {
#  "count" => "24471"
# }
```

```ruby
Set.new(User.pluck(:email)).difference(UserEmail.pluck(:email).flatten).count
#   (47.7ms)  SELECT "users"."email" FROM "users"
#   (24.8ms)  SELECT "user_emails"."email" FROM "user_emails"
# 24381
```

```ruby
User.find_by(email: "mary8284@gmail.com").user_emails
```

Possibly an artifact of anonymization?

sethherr: Hmmmm. are those unconfirmed users? Looks like there are 150,125 confirmed vs 124,961 unconfirmed

jmromer: Almost all, from the looks of it -- there's only 12 user emails not in users_emails once I limit the query to confirmed users.

spec/requests/api/v3/bikes_request_spec.rb
```diff
@@ -74,6 +74,83 @@
       end
     end
 
+    context "if the bike being created already exists" do
```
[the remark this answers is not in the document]
jmromer: > Bike was created by a user in an organization and has been claimed - creation organization posts a matching bike (what should happen?)

Intuitive default: update the ownership 
(Could this be a situation where a bike owner has sold a bike?)

> User who was never involved with original bike posts a bike that matches a bike a different user owns - including matching that users email (what should happen?)

Intuitive default: reject the request
Alternatively: would it be useful to have some record of that attempt? (in the form of a duplicate bike record, that is)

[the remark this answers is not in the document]
sethherr: First one - I absolutely agree. (This would be the user returning to the shop they bought the bike from to get it serviced and the shop registers it)

Second one - I think either option could result in unintuitiveness/frustration - rejecting the request or creating a duplicate record. I think we need another way to deal with this.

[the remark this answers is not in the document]
jmromer: For the second case, the user story I mainly have in mind is of a malicious actor, but I'm not totally sure it applies or, if it does, that it exhausts the possibilities 🤔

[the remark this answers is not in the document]
sethherr: We can discuss the second case offline. I don't think it's important to deal with right now.

PR #717, opened 2019-04-24
Fixed duplicates index

Robertgervais:


Robertgervais:
Image 4-24-19 at 3 01 PM

PR #724, opened 2019-04-26
Orgs locations revision

Robertgervais:


Robertgervais:
Okay, I had to turn the "Add new location" button into a link because it kept rendering as a huge, weird blob. I tried giving it it's own col but it kept kind of screwing up the layout. Also - I'm not sure if I understood what you wanted me to do with the location, so I'm just having the new layout location forms render with the the US code selected by default. Parent orgs select is fancified, too.

PR #726, opened 2019-04-26
Minor revision to nav bar

Robertgervais:


Robertgervais:
Ah yeah I know I was just thinking that with the controller_name check the active_link helper was redundant but I changed it back.

app/views/layouts/new_admin.html.haml
```diff
@@ -21,13 +21,13 @@
         .collapse.navbar-collapse#collapsible-admin-nav
           %ul.navbar-nav
             %li.nav-item
-              = active_link "Users", admin_users_path, role: "button", class: "nav-link"
+              = active_link "Users", admin_users_path, role: "button", class: controller_name == "users" ? "nav-link active" : "nav-link"
```
sethherr: These should all use the `match_controller` property of `active_link`

```haml
= active_link "Users", admin_users_path, role: "button", class: "nav-link", match_controller: true
```

PR #727, opened 2019-04-27
Revise admin graphs

Robertgervais:


Robertgervais:
Okay - I tried two things. Commenting out `moment.tz.setDefault(window.timezone);` seemed to stop the error I was getting, but then I was like yeah but we probably need that, so then I went into the graphs index and commented out the hidden_field_tag that was handling timezone, and that also seems to fix it. There's a bunch of stuff in the graphs controller useing `@timezone` but they don't seem to mind not having access to it. Anyways, this seems like a fix, but I'm not sure what you want me to do here.

Robertgervais:
Oh, the other thing, both locally AND in production the custom selects are no longer styled like calendars. I was doing some reading about datepicker?!?!? I'm assuming you want me to make these bad boys look like calendars.

Robertgervais:
lol okay one LAST LAST thing, there are a few pages in graphs that I can't find? I tried finding them via url but got some weird looking sidekiq-esque looking pages, like /admin/graphs/bikes or whatever. But yeah, review, stolen_locations, not totally sure where these are.

Robertgervais:
Never mind, this still needs some work - Sorry. I'll request you when I'm ready. I forgot that I had like, a lot of maybes about this one.

Robertgervais:
Okay, so, yeah, this seems to work with or without `time_zone: @timezone` on each chart. The front end is all styled, looks good, js works, etc. but wondering how you'd like me to handle the controller.

Robertgervais:
Image 5-3-19 at 6 56 PM

app/views/admin/graphs/index.html.haml
```diff
@@ -1,70 +1,60 @@
 
 / inline scss is wrong and shouldn't be done - but because we're avoiding adding new things to the existing admin stylesheets, putting it here anyway
 / TODO: put into admin stylesheets when we update admin stylesheets
```
sethherr: I think you can delete these comments, since they've been resolved? There is now scss in the new scss 🙂

Robertgervais:
#175

Issue #636, opened 2019-03-28
remove orphaned partial (when it's actually orphaned)


sethherr:
During admin revision, @Robertgervais figured out that admin was the only place that used the `bikes/bike` partial - I brought up the issue here: https://github.com/bikeindex/bike_index/pull/629#discussion_r269364568

Once admin has finished being revised and all the existing places that references that partial have been updated to reference the admin partial, we need to remove the `bikes/bike` partial (and preferably, move the `bikes/revised_bike` partial to `bikes/bike` and updated that)

Robertgervais:
`bikes/bike` removed in #728

Robertgervais:
closed with #785

PR #728, opened 2019-04-27
Revise admin bikes edit

Robertgervais:
Before Screenshot:

Screen Shot 2019-05-01 at 6 43 59 PM


Robertgervais:
Hey - I requested you on this, but then I realized that I should revise _fast_attr as well, since it gets rendered in edit sometimes too (honestly, I'm not totally sure when)

app/views/bikes/_bike.html.haml
```diff
@@ -1,32 +1,33 @@
 %li{ class: bike.stolen && "blist-stolen" }
   %a.blist-image-holder{href: bike_url(bike) }
```
sethherr: Also, it looks like the only place that is being rendered is fast_attr_update - to reduce confusion, delete this partial.

You can use the existing admin bike show on fast_attr_update, it'll be good enough

sethherr: https://github.com/bikeindex/bike_index/issues/636

Robertgervais:
Image 5-2-19 at 4 34 PM
Image 5-2-19 at 4 38 PM

app/views/bikes/_bike.html.haml
sethherr: ?

Robertgervais:
Image 5-2-19 at 6 16 PM

app/javascript/packs/application_css.scss
```diff
@@ -12,3 +12,39 @@
 footer.primary-footer {
   display: none;
 }
+
+#admin-bike-box img {
```
sethherr: I added I added an admin stylesheet in https://github.com/bikeindex/bike_index/pull/736 - merge conflict, plus the admin-bike-box should go there

app/views/admin/bikes/_bike.html.haml
```diff
@@ -1,32 +1,40 @@
...
-            %span.convertTime.preciseTime
-              = l bike.current_stolen_record.date_stolen, format: :convert_time
-        %tr
-          %td
-            From
-          %td
-            = bike.short_address(bike.current_stolen_record)
+  .row.align-items-end
+    .col-md-2
+      #admin-bike-box
```
sethherr: There should only be one ID per page. Also, in general, you shouldn't apply styles to IDs unless you really need to

app/views/admin/bikes/_bike.html.haml
```diff
@@ -1,32 +1,40 @@
...
+        - if bike.thumb_path.present?
+          %img{:alt => bike.title, src: bike.thumb_path}
+    .col-md-10
```
sethherr: I think you can fit this all on one line on mobile!

app/javascript/packs/pages/binx_admin.js
```diff
@@ -1,6 +1,9 @@
 import log from "../utils/log";
 import moment from "moment-timezone";
 import LoadFancySelects from "../utils/LoadFancySelects";
+import BinxAdminBikesEdit from "./binx_admin_bikes_edit.js"
```
sethherr: You're still including everything in this file from this

app/javascript/packs/pages/binx_admin.js
```diff
@@ -24,11 +25,30 @@ function BinxAdmin() {
       if ($("#admin-locations-fields").length > 0) {
         this.adminLocations();
       }
+      if ($("#bike_stolen").length > 0) {
```
sethherr: Good catch!

app/views/admin/bikes/edit.html.haml
```diff
@@ -1,226 +1,232 @@
...
+    /   = f.label :paint_id, "Paint description", class: "control-label"
+    /   .controls
+    /     = collection_select( :bike, :paint_id, Paint.all, :id, :name, required: true, prompt: "Choose paint" )
+    .row
+      .col-lg-6
+        .form-group.fancy-select.unfancy
+          = f.label :primary_frame_color_id
+          = collection_select(:bike, :primary_frame_color_id, Color.all, :id, :name, required: true, prompt: "Choose color" )
+      .col-lg-6
+        -# Do you want this here or in the controller?
```
sethherr: I think putting this into the controller is a great idea!

Robertgervais:
ahh it was overflowing because my styles got deleted somewhere and the image wasn't getting sized down. Doesn't overflow anymore/The columns aren't stretched anymore.

app/views/admin/bikes/edit.html.haml
```diff
@@ -1,251 +1,256 @@
...
         %th
           Claimed?
-        %th
+        %th.d-none.d-md-table-cell
           User hidden
       %tbody
         - @bike.ownerships.each do |ownership|
           %tr
-            %td
+            %td.d-none.d-lg-table-cell
```
sethherr: While I agree that the ID doesn't need to be visible on small screens - this makes it so that the link to edit the ownership isn't visible on small screens 🙃.

I also think the "created at" column is critical functionality to view.

I think you could make this be like the bikes table, where the link is in the created at and then use the less-strong right to put the ID.

Robertgervais:
Made the created_at's link to page, stopped hiding creator, hiding did we help can we share on small screens.

Robertgervais:
#175

PR #729, opened 2019-04-27
Revise missing manufacturer view

Robertgervais:


Robertgervais:
So I "revised" this page, but I'm not a hundred percent sure i know what's going on on this page. Whenever I try to update the page, locally or in production, I get an error (handled by us, not a bad one)

Robertgervais:
Oooooh, I actually understand what this page is doing now.

app/views/admin/bikes/missing_manufacturer.html.haml
```diff
@@ -1,25 +1,21 @@
...
-%header.with-subtitle
-  %h1
-    Multi manufacturer update
-  .sharing-links
-    %label.checkbox.disabled
-      = check_box_tag :time_ordered, 'time_ordered', session[:missing_manufacturer_time_order]
-      time ordered
+%h1
+  Multi manufacturer update
+%label.checkbox.disabled
```
sethherr: This isn't ideal right now, you have to pass a parameter to get it to do time ordering (which you can only find by looking at the code).

The fact that it uses session to store it is good though, since it persists after updating.

It would be great if you updated this to be a button instead, which when clicked, toggles "time_ordered". 

*Note:* The controller is already set up to take care of this, but to get more consistent results, you should change this line in the controller:

```ruby
session[:missing_manufacturer_time_order] = params[:time_ordered] if params[:time_ordered].present?
```

To this:

```ruby
session[:missing_manufacturer_time_order] = ActiveRecord::Type::Boolean.new.type_cast_from_database(params[:time_ordered]) if params[:time_ordered].present?
```

Robertgervais:
#175

PR #730, opened 2019-04-30
Dedupe registration (seth work)

sethherr:
Some work on #715


sethherr:
Post chat additions:

- I cleaned up a little bit
- There are some (currently existing) issues with component creation - I opened #731 to deal with those issues, which will in turn fix some of the things I added to the v2 specs

PR #731, opened 2019-04-30
Fix issues with component creator

sethherr:


sethherr:
Also, remove `Ctype.unknown` because other things have `.other` (e.g. `Manufacturer`) and nothing else has `.unknown` - and there isn't a good reason to have both.

sethherr:
I think `ComponentCreator` could eventually be entirely folded into the Component model file - I don't think that's true of all the Bike Creator services

PR #740, opened 2019-05-02
Fix failed bikes page exception

### LINKED ISSUE #714: NoMethodError in Admin::FailedBikes#index
Steps to reproduce:
1. Navigate to http://localhost:3001/admin/failed_bikes
2. Observe the following 500 error:

> Showing app/views/admin/failed_bikes/index.html.haml where line #33 raised:
```
undefined method `each_key' for ["Manufacturer can't be blank"]:Array
Did you mean?  each_entry
```

```haml
          %td
            - if b_param.bike_errors.present?
              = b_param.bike_errors.each_key { |f| f.to_s.humanize}   # <----------
            - if b_param.errors.present?
              = b_param.errors.each_key { |f| f.to_s.humanize}
            / - if b_param.params.creation_organization_id && b_param.params.creation_organization_id
```

jmromer:
Fixes #714:

- Replaces `each_key` invocation on Array with `each`
- Eliminates N+1 query by preloading associated creators
- Paginates failed bike output (otherwise the page loads extremely slowly)

**Screenshot**

Screen Shot 2019-05-02 at 2 19 20 PM

### DIFF
--- a/app/controllers/admin/failed_bikes_controller.rb
+++ b/app/controllers/admin/failed_bikes_controller.rb
@@ -5,6 +5,8 @@ class Admin::FailedBikesController < Admin::BaseController
     page = params.fetch(:page, 1)
     per_page = params.fetch(:per_page, 25)
 
+    @b_params_total_count = BParam.where("created_bike_id IS NOT NULL").count
+
     @b_params =
       BParam
         .where("created_bike_id IS NOT NULL")
--- a/app/controllers/admin/failed_bikes_controller.rb
+++ b/app/controllers/admin/failed_bikes_controller.rb
@@ -5,10 +5,9 @@ class Admin::FailedBikesController < Admin::BaseController
     page = params.fetch(:page, 1)
     per_page = params.fetch(:per_page, 25)
 
-    @b_params_total_count = BParam.where("created_bike_id IS NOT NULL").count
-
     @b_params =
       BParam
+        .includes(:creator)
         .where("created_bike_id IS NOT NULL")
         .order("created_at desc")
         .page(page)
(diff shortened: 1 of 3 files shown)


app/views/admin/failed_bikes/index.html.haml
```diff
@@ -1,16 +1,13 @@
-
 %h1
   Manage Failed Bikes
 
 %h4
-  = @b_params.count
+  = number_with_delimiter(@b_params_total_count, delimiter: ',')
```
sethherr: You can get this with `@b_params.total_count`, I think?

app/controllers/admin/failed_bikes_controller.rb
```diff
@@ -2,7 +2,17 @@ class Admin::FailedBikesController < Admin::BaseController
   layout "new_admin"
 
   def index
-    @b_params = BParam.where("created_bike_id IS NOT NULL").order("created_at desc")
+    page = params.fetch(:page, 1)
```
sethherr: OH SNAP! I didn't know you could pass a nil value to fetch, that's awesome 🙂

app/views/admin/failed_bikes/index.html.haml
```diff
@@ -21,7 +18,7 @@
       %th
         Errors
     %tbody
-      - @b_params.each do |b_param|
+      - @b_params.includes(:creator).each do |b_param|
```
sethherr: Why not do this `includes` in the controller? That's what we've done in other places in the app. If there's a good reason to do it in the view instead, let's do it (and add an issue for moving the other ones into the views)

app/views/admin/failed_bikes/index.html.haml
jmromer: ah good catch thanks 👍

PR #749, opened 2019-05-06
Add match cases request specs

jmromer:
- Adds request-spec test cases for matching criteria

### DIFF
--- a/app/services/bike_finder.rb
+++ b/app/services/bike_finder.rb
@@ -10,36 +10,25 @@ module BikeFinder
   # - any email associated with any owner (via user.user_emails)
   #
   # Return a Bike object, or nil
-  def self.find_matching(serial:, owner_email:, current_user: nil)
-    owner_email = EmailNormalizer.normalize(owner_email)
-    serial = SerialNormalizer.new(serial: serial).normalized
+  def self.find_matching(serial:, owner_email:)
+    email = EmailNormalizer.normalize(owner_email)
 
     candidate_user_ids =
       User
         .joins("LEFT JOIN user_emails ON user_emails.user_id = users.id")
-        .where("users.email = ? OR user_emails.email = ?", owner_email, owner_email)
+        .where("users.email = ? OR user_emails.email = ?", email, email)
         .select(:id)
         .uniq
 
-    found_bike = Bike
+    Bike
       .joins("LEFT JOIN ownerships ON bikes.id = ownerships.bike_id")
-      .where(serial_normalized: serial)
+      .where(serial_normalized: SerialNormalizer.new(serial: serial).normalized)
       .where(
         "bikes.owner_email = ? OR ownerships.user_id IN (?) OR ownerships.creator_id IN (?)",
-        owner_email,
+        email,
         candidate_user_ids,
         candidate_user_ids
       )
       .first
-
-    return found_bike if found_bike.present?
-
-    if current_user.present?
-      Bike
-        .includes(:ownerships)
-        .where(serial_normalized: serial, ownerships: { claimed: true })
-        .select { |b| b.creation_organization&.in?(current_user.organizations) }
-        .first
-    end
   end
 end
--- a/app/services/bike_finder.rb
+++ b/app/services/bike_finder.rb
@@ -25,7 +25,7 @@ module BikeFinder
       .where(serial_normalized: SerialNormalizer.new(serial: serial).normalized)
       .where(
         "bikes.owner_email = ? OR ownerships.user_id IN (?) OR ownerships.creator_id IN (?)",
-        owner_email,
+        email,
         candidate_user_ids,
         candidate_user_ids
       )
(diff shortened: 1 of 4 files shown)


app/services/bike_finder.rb
```diff
@@ -25,7 +25,7 @@ def self.find_matching(serial:, owner_email:)
       .where(serial_normalized: SerialNormalizer.new(serial: serial).normalized)
       .where(
         "bikes.owner_email = ? OR ownerships.user_id IN (?) OR ownerships.creator_id IN (?)",
-        owner_email,
+        email,
```
sethherr: Good catch! Why don't any of the `BikeFinder` specs catch this?

spec/requests/api/v3/bikes_request_spec.rb
```diff
@@ -149,7 +149,132 @@
...
+             bike_attrs.to_json,
+             json_headers
+
+        returned_bike = json_result["bike"]
+        expect(response.status).to eq(201)
+        expect(response.status_message).to eq("Created")
+        expect(returned_bike["id"]).to_not eq(bike.id)
+      end
+
+      it "creates a new bike if it is POSTed by a creation org whose member initially created the match" do
```
sethherr: Is this statement accurate? Why does this create another bike?

jmromer: Is the test description a mistranslation of the following? 🤔 

> Bike was created by a user in an organization and has been claimed - creation organization posts a matching bike (what should happen?)

jmromer: > Why does this create another bike?

That's a good catch thanks -- it should be updating the ownership

sethherr: It didn't look like that ownership was claimed, so I was curious.

app/services/bike_finder.rb
jmromer: Updated 👍  b56df9b8

spec/requests/api/v3/bikes_request_spec.rb
jmromer: @sethherr Re the claimed status -- I'm trying to figure out the intent of the existing logic here. Currently we reject updates if a bike is already claimed:

```rb
# app/controllers/api/v2/bikes.rb L175-176 (898bfe65)

@bike = found_bike
authorize_bike_for_user
```

```rb
# app/controllers/api/v2/bikes.rb L83-86 (00000000)

def authorize_bike_for_user(addendum = "")
  return true if @bike.authorize_for_user!(current_user)
  error!("You do not own that #{@bike.type}#{addendum}", 403)
end
```

```rb
# app/models/bike.rb L254-258 (3630a8d5)

def authorize_for_user!(u)
  return authorize_for_user(u) unless claimable_by?(u)
  current_ownership.mark_claimed
  true
end
```

```rb
# app/models/bike.rb L242-245 (aa4f5113)

def claimable_by?(u)
  return false if u.blank? || current_ownership.blank? || current_ownership.claimed?  <-----
  user == u || current_ownership.claimable_by?(u)
end
```

```rb
# app/models/bike.rb L247-251 (3630a8d5)

def authorize_for_user(u)
  return true if u == owner || claimable_by?(u)
  return false if u.blank? || current_ownership&.claimed    <-----
  authorized_by_organization?(u: u)
end
```

```
» claimable_by?(u)
false
» authorize_for_user(u)
false
```

In the preceding, the bike's `creation_organization` is the same as the current user's `organization`, and the bike is claimed:
```
» creation_organization.name
"Organization 1"
» u.organizations.first.name
"Organization 1"
» claimed?
true
```

Do we want to change this behavior? 

WIP implementation at 08badaba

sethherr: The intent of the existing behavior is that, until a user claims a bike, the organization that created it can continue to update it.

This behavior is to take care of these issues:
- A bike shop registers a bike for a user, but wants to add a photo after registering it
- ... Or some extra details about the bike (e.g. a specific component or a description)
- A bike was sent to the wrong email address, and the original owner needs to send it to the correct one

The initial goal was to make it so that the editing was unobtrusive, but cut off access to it once the user was officially in charge of the bike. This seemed like unsurprising behavior - since if you haven't officially claimed the bicycle, it seems reasonable that the person who sent it to you can update it. Once you've claimed it, you don't expect someone else to be able to go in and modify things about the bike.

When building this system, I imagined users would claim bikes within the next day and that organizations wouldn't have a desire to edit registrations once they had finished the initial editing of them. Both of these assumptions turned out to be incorrect - users wait longer and organizations sometimes want to update additional bits later on. I still think the current behaviour makes sense, but it doesn't cover all use cases.

jmromer: I reverted the test and implementation for that last user story — I think we should ship the bug fix and tests on this branch and spec out the remaining story separately to avoid scope creep.

PR #752, opened 2019-05-08
Move graphs to separate JS file

Robertgervais:
So, I know you didn't ask me to do this... but I did it. I'm not gonna be mad if you hate it, but I was trying to clean up the binx_admin file since I'm rooting around in there so much right now.


app/javascript/packs/pages/binx_admin.js
```diff
@@ -2,6 +2,7 @@ import log from "../utils/log";
 import moment from "moment-timezone";
 import LoadFancySelects from "../utils/LoadFancySelects";
 import BinxAdminInvoices from "./binx_admin_invoices.js";
+import BinxGraphs from "./binx_graphs.js"
```
sethherr: I think we're getting enough admin js that we should create an `admin` folder in the pages folder, and put all the admin stuff in there.

sethherr: My thoughts on naming conventions for this: 
- All classes/functions should be prefixed with `BinxAdmin{{Name}}`
- The filenames should be just the final part

e.g. for this, filename would be `graphs.js` and the classname would be `BinxAdminGraphs`

Robertgervais:
Hey, I know this is failing but I do not know *why* because the code blocks it's saying are identical are not 🤷‍♂️

PR #755, opened 2019-05-10
Add ambassador organization type, ambassador dashboard

jmromer:
- Introduces an "ambassador" organization type
- Adds an ambassador dashboard endpoint in the `organized` namespace
- Directs user to the either the ambassador dashboard or the bikes page based on whether the current org is an ambassador org
- Introduces `User.ambassadors` to list all ambassadors, optionally filtering by organization
- Introduces `User#ambassador?` 
- Allows ambassador users to edit bikes (and mark them as recovered)

Screen Shot 2019-05-14 at 9 42 58 AM

Screen Shot 2019-05-14 at 9 42 43 AM

Screen Shot 2019-05-14 at 9 43 29 AM

Screen Shot 2019-05-14 at 11 37 29 AM

### DIFF
--- a/app/controllers/organizations_controller.rb
+++ b/app/controllers/organizations_controller.rb
@@ -27,7 +27,7 @@ class OrganizationsController < ApplicationController
   def create
     @organization = Organization.new(permitted_create_params)
     if @organization.save
-      membership = Membership.create(user_id: current_user.id, role: "admin", organization_id: @organization.id)
+      Membership.create(user_id: current_user.id, role: "admin", organization_id: @organization.id)
       notify_admins("organization_created")
       flash[:success] = "Organization Created successfully!"
       if current_user.present?
@@ -102,7 +102,7 @@ class OrganizationsController < ApplicationController
 
   def permitted_create_params
     approved_kind = params.dig(:organization, :kind)
-    approved_kind = "other" unless Organization.kinds.include?(approved_kind)
+    approved_kind = "other" unless Organization.creatable_kinds.include?(approved_kind)
     params.require(:organization)
           .permit(:name, :website)
           .merge(auto_user_id: current_user.id, kind: approved_kind)
--- a/app/models/membership.rb
+++ b/app/models/membership.rb
@@ -12,6 +12,8 @@ class Membership < ActiveRecord::Base
 
   after_commit :update_relationships
 
+  scope :ambassador_organizations, -> { where(organization: Organization.ambassador).order(:created_at) }
+
   def self.membership_types
     MEMBERSHIP_TYPES
   end
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -9,6 +9,7 @@ class Organization < ActiveRecord::Base
     software: 5,
     property_management: 6,
     other: 7,
+    ambassador: 8,
   }.freeze
 
   acts_as_paranoid
@@ -81,6 +82,11 @@ class Organization < ActiveRecord::Base
     matching_slugs = PaidFeature.matching_slugs(slugs)
     return nil unless matching_slugs.present?
     where("paid_feature_slugs ?& array[:keys]", keys: matching_slugs)
+  end
+
+  # Organization kinds creatable by non-superadmins
+  def self.creatable_kinds
+    kinds.reject { |kind| kind == "ambassador" }
   end
 
   def to_param
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -37,6 +37,7 @@ class User < ActiveRecord::Base
 
   scope :confirmed, -> { where(confirmed: true) }
   scope :unconfirmed, -> { where(confirmed: false) }
+  scope :ambassadors, -> { where(id: Membership.ambassador_organizations.select(:user_id)).order(created_at: :asc) }
 
   validates_uniqueness_of :username, case_sensitive: false
 
@@ -120,6 +121,10 @@ class User < ActiveRecord::Base
   def superuser?; superuser end
 
   def developer?; developer end
+
+  def ambassador?
+    memberships.ambassador_organizations.any?
+  end
 
   def to_param; username end
 
--- a/spec/factories/users.rb
+++ b/spec/factories/users.rb
@@ -19,6 +19,11 @@ FactoryBot.define do
       factory :developer do
         developer { true }
       end
+      factory :user_ambassador do
+        after(:create) do |user, _evaluator|
+          FactoryBot.create(:membership_ambassador, user: user)
+        end
+      end
       factory :organized_user do
         # This factory should not be used directly, it's here to wrap organization
         # Use `organization_member` or `organization_admin`
--- a/spec/requests/api/v3/organizations_request_spec.rb
+++ b/spec/requests/api/v3/organizations_request_spec.rb
@@ -63,7 +63,7 @@ describe "Organization API V3" do
           ENV["ALLOWED_WRITE_ORGANIZATIONS"] = "some-other-uid"
           post url, organization_json, json_headers
           expect_status 401
-          expect_json(error: "Unauthorized. Cannot write organiztions")
+          expect_json(error: "Unauthorized. Cannot write organizations")
         end
       end
     end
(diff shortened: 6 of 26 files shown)


app/controllers/organizations_controller.rb
```diff
@@ -27,7 +27,7 @@ def connect_lightspeed
   def create
     @organization = Organization.new(permitted_create_params)
     if @organization.save
-      membership = Membership.create(user_id: current_user.id, role: "admin", organization_id: @organization.id)
+      _membership = Membership.create(user_id: current_user.id, role: "admin", organization_id: @organization.id)
```
sethherr: I'd prefer just removing this assignment, rather than using an underscore variable.

app/models/user.rb
```diff
@@ -93,6 +93,15 @@ def admin_text_search(n)
       (where("name ILIKE ? OR email ILIKE ?", search_str, search_str) +
        joins(:user_emails).where("user_emails.email ILIKE ?", search_str)).uniq
     end
+
+    def ambassadors(organization: nil)
+      ambassador_orgs = Organization.ambassador
```
sethherr: I'm not sure this is possible - but - if it is, I think I'd prefer defining a scope on `Membership` for ambassador organizations, and then using a scope (rather than a class method) that references that scope.

I'm not *completely* sure the scoped solution is much more performant, but I think it looks a lot cleaner. If it's enormously more performant, I'm open to doing the class method.

app/models/user.rb
```diff
@@ -121,6 +130,18 @@ def superuser?; superuser end
 
   def developer?; developer end
 
+  def ambassador?
+    ambassador_index =
```
sethherr: Another benefit of the scope option means that this could be reduced to `memberships.ambassador_organizations.any?`

spec/requests/api/v3/organizations_request_spec.rb
```diff
@@ -63,7 +63,7 @@
           ENV["ALLOWED_WRITE_ORGANIZATIONS"] = "some-other-uid"
           post url, organization_json, json_headers
           expect_status 401
-          expect_json(error: "Unauthorized. Cannot write organiztions")
+          expect_json(error: "Unauthorized. Cannot write organizations")
```
sethherr: Thanks!

app/models/user.rb
sethherr: Corollary - I'm not sure I see the value to passing in organization here - because you could just use `organization.users` if you wanted access to those ambassadors

db/migrate/20190514155447_create_ambassador_tasks.rb
```diff
@@ -0,0 +1,20 @@
...
+    create_table :ambassador_tasks do |t|
+      t.string :description, null: false, default: ""
+
+      t.timestamps null: false
+    end
+
+    create_table :ambassador_task_assignments do |t|
+      t.belongs_to :user, index: true, null: false
+      t.belongs_to :ambassador_task, index: true, null: false
+      t.boolean :completed, null: false, default: false
```
sethherr: Rather than a boolean completed, I have leaned toward a timestamp (so `completed_at`). If you don't like this, let me know - it just often becomes helpful down the road.

spec/factories/users.rb
```diff
@@ -19,6 +19,11 @@
       factory :developer do
         developer { true }
       end
+      factory :ambassador do
```
sethherr: ... this isn't the naming convention that exists for everyting, but I decided a little while ago that I'd prefer if factories were prefixed with the name of the parent factory - so this would be `user_ambassador`. At some point, it would be great if we updated the existing ones, but that's not a big priority.

It just helps keep things on track. This goes for the other factories as well

app/controllers/organizations_controller.rb
sethherr: Good catch on seeing that it wasn't necessary though!

app/models/user.rb
jmromer: Good catch thanks! Updated

app/models/membership.rb
```diff
@@ -12,6 +12,8 @@ class Membership < ActiveRecord::Base
 
   after_commit :update_relationships
 
+  scope :ambassador_organizations, -> { where(organization: Organization.ambassador).order(:created_at) }
```
sethherr: Oooh, nifty. Does this generate good sql?

app/models/organization.rb
```diff
@@ -83,6 +84,11 @@ def self.with_paid_feature_slugs(slugs)
     where("paid_feature_slugs ?& array[:keys]", keys: matching_slugs)
   end
 
+  # Organization kinds creatable by non-superadmins
+  def self.creatable_kinds
+    kinds.reject { |kind| kind == "ambassador" }
```
sethherr: Just a question (not a request for changing): Is this more performant that `kinds - ["ambassador"]`?

jmromer: Not sure. There's an extra array allocation with the array subtraction approach, but I think it's a negligible difference.

Semantically array subtraction amounts to 

```ruby
kinds.reject {|kind| lhs_arr.any? { |e| kind.eql?(e) } }
```

`reject` is a little more flexible since it accepts a block, array subtraction a little more succinct.

app/models/membership.rb
jmromer: I think so? If I'm missing a better approach, lmk. This is the SQL that's generated:

```sql
SELECT
   "memberships".* 
FROM
   "memberships" 
WHERE
   "memberships"."deleted_at" IS NULL 
   AND "memberships"."organization_id" IN 
   (
      SELECT
         "organizations"."id" 
      FROM
         "organizations" 
      WHERE
         "organizations"."deleted_at" IS NULL 
         AND "organizations"."kind" = $1 
      ORDER BY
         "organizations"."name" ASC
   )
ORDER BY
   "memberships"."created_at" ASC 
```

An alternative:

```ruby
Membership.includes(:organization).where(organizations: {kind: 8})
```
```sql
SELECT
   COUNT(DISTINCT "memberships"."id") 
FROM
   "memberships" 
   LEFT OUTER JOIN
      "organizations" 
      ON "organizations"."id" = "memberships"."organization_id" 
      AND "organizations"."deleted_at" IS NULL 
WHERE
   "memberships"."deleted_at" IS NULL 
   AND "organizations"."kind" = $1 


SELECT
      "memberships"."id" AS t0_r0,
      "memberships"."organization_id" AS t0_r1,
      "memberships"."user_id" AS t0_r2,
      "memberships"."role" AS t0_r3,
      "memberships"."invited_email" AS t0_r4,
      "memberships"."created_at" AS t0_r5,
      "memberships"."updated_at" AS t0_r6,
      "memberships"."deleted_at" AS t0_r7,
      "organizations"."id" AS t1_r0,
      "organizations"."name" AS t1_r1,
      "organizations"."slug" AS t1_r2,
      "organizations"."available_invitation_count" AS t1_r3,
      "organizations"."created_at" AS t1_r4,
      "organizations"."updated_at" AS t1_r5,
      "organizations"."website" AS t1_r6,
      "organizations"."short_name" AS t1_r7,
      "organizations"."show_on_map" AS t1_r8,
      "organizations"."sent_invitation_count" AS t1_r9,
      "organizations"."deleted_at" AS t1_r10,
      "organizations"."is_suspended" AS t1_r11,
      "organizations"."auto_user_id" AS t1_r12,
      "organizations"."access_token" AS t1_r13,
      "organizations"."api_access_approved" AS t1_r14,
      "organizations"."approved" AS t1_r15,
      "organizations"."avatar" AS t1_r16,
      "organizations"."is_paid" AS t1_r17,
      "organizations"."lock_show_on_map" AS t1_r18,
      "organizations"."landing_html" AS t1_r19,
      "organizations"."paid_feature_slugs" AS t1_r20,
      "organizations"."parent_organization_id" AS t1_r21,
      "organizations"."kind" AS t1_r22,
      "organizations"."ascend_name" AS t1_r23,
      "organizations"."registration_field_labels" AS t1_r24 
   FROM
      "memberships" 
      LEFT OUTER JOIN
         "organizations" 
         ON "organizations"."id" = "memberships"."organization_id" 
         AND "organizations"."deleted_at" IS NULL 
   WHERE
      "memberships"."deleted_at" IS NULL 
      AND "organizations"."kind" = $1 
```

sethherr: Cool, I like the readability of your initial approach

PR #756, opened 2019-05-11
Minor revision to bikes edit

Robertgervais:


Robertgervais:
Image 5-10-19 at 10 00 PM

Robertgervais:
Image 5-14-19 at 4 29 PM

app/views/admin/bikes/_bike.html.haml
```diff
@@ -19,18 +22,22 @@
...
-              %td
-                %span.convertTime.preciseTime
-                  = l bike.current_stolen_record.date_stolen, format: :convert_time
-            %tr
-              %td
-                From
-              %td
-                = bike.short_address(bike.current_stolen_record)
+    - if bike.stolen and bike.current_stolen_record.present?
+      .col-lg-6.border.border-danger.mt-auto
```
sethherr: Nice. I like only specifying col width lg here. 👍

sethherr: Also, thanks for explaining what `mt-auto` does to me 🙂

PR #760, opened 2019-05-15
Add ambassador tasks

jmromer:
- Adds `AmbassadorTask` and `AmbassadorTaskAssignment` to model ambassador tasks
- Adds an `Ambassador` decorator
- Adds `Ambassador#progress` to compute a given ambassador's current progress on task assignments
- Adds an admin interface to manage Ambassadors and AmbassadorTasks
- Adds a input field to mark tasks as completed

demo


app/controllers/organized/base_controller.rb
```diff
@@ -25,6 +25,12 @@ def ensure_admin!
       redirect_to organization_bikes_path(organization_id: current_organization.to_param) and return
     end
 
+    def ensure_ambassador!
+      return true if current_user && current_user.ambassador?
```
sethherr: superusers need to be able to view these pages too

app/models/ambassador_task.rb
```diff
@@ -0,0 +1,23 @@
...
+
+  validates :description, uniqueness: true
+
+  def description_html
+    Kramdown::Document.new(description).to_html
+  end
+
+  def assign_to(user)
+    return unless user.ambassador?
+    AmbassadorTaskAssignment.create(ambassador_task: self, user: user)
```
sethherr: nit: you could do `ambassador_task_assignments.create(user: user)`

app/models/ambassador.rb
```diff
@@ -0,0 +1,41 @@
...
+  def completed_tasks_count
+    ambassador_task_assignments.select(&:completed?).count
+  end
+
+  def tasks_count
+    ambassador_task_assignments.count
+  end
+
+  def groups
+    Organization.where(id: memberships.select(:organization_id))
```
sethherr: `select` isn't like sequel here, it will instantiate all the memberships records. I think it would be better with `membership.pluck("DISTINCT organization_id")`

app/models/ambassador_task.rb
```diff
@@ -0,0 +1,23 @@
...
+  has_many :users, through: :ambassador_task_assignments
+
+  validates :description, uniqueness: true
+
+  def description_html
+    Kramdown::Document.new(description).to_html
+  end
+
+  def assign_to(user)
+    return unless user.ambassador?
```
sethherr: This will hide errors. e.g. if I try `assign_to(user)` and that user is not an ambassador, the validation you added will stop the creation, and you'll be able to see why - if you do this, it will instead return nil instead of a task, and you won't see any errors.

app/controllers/organized/base_controller.rb
```diff
@@ -25,6 +25,12 @@ def ensure_admin!
...
+      flash[:error] = "You have to be an ambassador to do that!"
+      redirect_to root_url
```
sethherr: rather than `root_url` here, it would be nice to redirect to `user_root_url`

app/models/ambassador.rb
```diff
@@ -0,0 +1,41 @@
...
+    return 0.0 if ambassador_task_assignments.empty?
+    (completed_tasks_count / tasks_count.to_f).round(2)
+  end
+
+  def progress_count
+    "#{completed_tasks_count}/#{tasks_count}"
+  end
+
+  def completed_tasks_count
+    ambassador_task_assignments.select(&:completed?).count
```
sethherr: Similarly, this isn't like sql and will instantiate all the `ambassador_task_assignments` - if you add a `completed` scope to `ambassador_task_assignment` and do `ambassador_task_assignments.completed.count`, you will be executing only one sql command count

app/models/ambassador_task_assignment.rb
```diff
@@ -0,0 +1,29 @@
...
+
+  validates :user, presence: true
+  validates :ambassador_task, presence: true
+
+  validates :ambassador_task, uniqueness: { scope: :user }
+  validate :associated_user_is_an_ambassador
+
+  delegate :description, :description_html, to: :ambassador_task
+
+  default_scope { order(created_at: :asc) }
```
sethherr: I used to use default scopes regularly. It turns out [they're a pain](https://andycroll.com/ruby/dont-use-default-scope/) and difficult to remove. I would recommend avoiding using them.

app/models/ambassador.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  def tasks_count
+    ambassador_task_assignments.count
+  end
+
+  def groups
```
sethherr: Why use groups rather than organizations here?

sethherr: I think this might be better named `ambassador_organizations` - and then you could just have the method return `organizations.ambassadors` - which is important

app/views/admin/ambassadors/show.html.haml
```diff
@@ -0,0 +1,30 @@
+%h1= "Ambassador: #{@ambassador.username}"
+
+- if @ambassador.avatar
+  - unless @ambassador.avatar.url(:medium) == 'https://files.bikeindex.org/blank.png'
+    = image_tag @ambassador.avatar.url(:medium), class: 'users-show-avatar'
+
+
+%table.table.table-striped.table-hover.table-bordered.table-sm.without-exterior-border
+  %tbody
+    - @ambassador.attributes.slice(*@user_attributes).each_pair do |key, value|
```
sethherr: I think I would prefer this list to be defined here rather than the controller.

app/views/admin/ambassadors/show.html.haml
```diff
@@ -0,0 +1,30 @@
...
+%table.table.table-striped.table-hover.table-bordered.table-sm.without-exterior-border.ambassador-tasks-table
+  %thead.small-header
+    %tr
+      %th Task
+      %th Status
+  %tbody
+    - @ambassador.ambassador_task_assignments.each do |task|
+      %tr
+        %td= task.description_html.html_safe
+        %td= task.status
```
sethherr: Rather than status here, it would be nice to get a completed at date.

There is a handy way of displaying this:

```haml
- if task.completed_at
  %span.convertTime
    = l task.completed_at, format: :convert_time
```

app/views/admin/ambassadors/show.html.haml
```diff
@@ -0,0 +1,30 @@
...
+
+%table.table.table-striped.table-hover.table-bordered.table-sm.without-exterior-border.ambassador-tasks-table
+  %thead.small-header
+    %tr
+      %th Task
+      %th Status
+  %tbody
+    - @ambassador.ambassador_task_assignments.each do |task|
+      %tr
+        %td= task.description_html.html_safe
```
sethherr: I think this points at the fact that ambassador_tasks should have titles in addition to descriptions.

app/models/ambassador.rb
sethherr: Actually, https://github.com/bikeindex/bike_index/pull/760/files#r284909158 is a better solution

app/models/ambassador.rb
```diff
@@ -0,0 +1,41 @@
+class Ambassador < SimpleDelegator
```
sethherr: Why not just put these methods in user?

I can't do `Ambassador.where(created_at: (Time.now - 1.week)..Time.now)` - or `Ambassador.last` - which is stuff I would expect to be able to do.

I think another option would be having it be a concern?

app/views/admin/ambassadors/show.html.haml
```diff
@@ -0,0 +1,30 @@
+%h1= "Ambassador: #{@ambassador.username}"
+
+- if @ambassador.avatar
+  - unless @ambassador.avatar.url(:medium) == 'https://files.bikeindex.org/blank.png'
+    = image_tag @ambassador.avatar.url(:medium), class: 'users-show-avatar'
+
+
+%table.table.table-striped.table-hover.table-bordered.table-sm.without-exterior-border
```
sethherr: If you use `table.table-list` without any extra classes, it will make a definition list like the other definition table lists in admin, and you don't need to bold the first cell

app/models/ambassador_task.rb
sethherr: Something additional this method could handle is making sure we don't assign the same task to the same user multiple times

app/models/ambassador.rb
jmromer: Good call thanks, a decorator isn't quite right here. Refactored to inherited from User instead -- lmk what you think.

sethherr: Ooooh! Clever, yeah, if that works that's nifty!

app/models/ambassador.rb
```diff
@@ -0,0 +1,24 @@
+class Ambassador < User
```
sethherr: This is clever 🙂

app/views/admin/ambassadors/show.html.haml
sethherr: I still feel the same way about this - I feel like this is a view concern, and would prefer to define the `USER_ATTRIBUTES` here, rather than in the controller.

But this is a nit, and isn't actually important, and this is better than nothing!

I think in the future, we should create a partial for admin user display, like we have for bike display, and include that partial in the admin user edit, here and any other relevant admin places.

db/migrate/20190517161246_add_title_to_ambassador_tasks.rb
```diff
@@ -0,0 +1,9 @@
+class AddTitleToAmbassadorTasks < ActiveRecord::Migration
+  def change
+    change_table :ambassador_tasks do |t|
+      t.column :title, :string, null: false, default: ""
+    end
+
+    add_index :ambassador_tasks, :title, unique: true
```
sethherr: Why add this index? Is it just to make creating/updating faster by validating uniqueness at the db level?

What happens if an ambassador_task is created without a title?

app/views/admin/ambassadors/show.html.haml
jmromer: Ah, I misread your last (slightly ambiguous antecedent to "this", my bad). 
Added a decorator in df9185c7

db/migrate/20190517161246_add_title_to_ambassador_tasks.rb
jmromer: My default heuristic for applying data integrity constraints is something like:

1. Add a db-level constraint if one is available
2. Add an app-level constraint if friendlier user-facing errors are needed or if (1) isn't available

There's a presence validator in `AmbassadorTask` that'll keep blank tasks from being created, but if you force it through you can persist one task with a blank title.

```rb
# app/models/ambassador_task.rb L5 (379470e9)

validates :title, presence: true, uniqueness: true
```

Duplicate titles (including duplicate blank ones) will raise an exception if they make it through to the database:

```
2.5.1 (bikeindex)[22] » AmbassadorTask.create
   (0.1ms)  BEGIN
  AmbassadorTask Exists (0.2ms)  SELECT  1 AS one FROM "ambassador_tasks" WHERE "ambassador_tasks"."title" = '' LIMIT 1
   (0.1ms)  ROLLBACK
```

```
2.5.1 (bikeindex)[23] » dupe_task.save(validate: false)
   (0.1ms)  BEGIN
  SQL (0.6ms)  INSERT INTO "ambassador_tasks" ("title", "created_at", "updated_at") VALUES ($1, $2, $3) RETURNING "id"  [["title", "One"], ["created_at", "2019-05-18 01:38:31.581295"], ["updated_at", "2019-05-18 01:38:31.581295"]]
   (0.1ms)  ROLLBACK
ActiveRecord::RecordNotUnique: PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_ambassador_tasks_on_title"
DETAIL:  Key (title)=(One) already exists.
: INSERT INTO "ambassador_tasks" ("title", "created_at", "updated_at") VALUES ($1, $2, $3) RETURNING "id"
```

app/views/admin/ambassadors/show.html.haml
sethherr: Yeah, sorry, I definitely didn't explain myself very well 🙂

Issue #713, opened 2019-04-23
Admin Bikes table multidelete is broken


sethherr:
Since the revision of `/admin/bikes`, the multi delete functionality does not work.

I'm not entirely sure it needs to be fixed - it potentially could be removed - but it either needs to be fixed or removed, not left in a broken state

Robertgervais:
Closed with #786

PR #786, opened 2019-05-23
Multi delete bike table fix

Robertgervais:


Robertgervais:
Let me know if you need a photo. i have a screencap but it's FULL of emails, will slack to you if need be.

Robertgervais:
In response to https://github.com/bikeindex/bike_index/issues/713

PR #787, opened 2019-05-23
Restrict ambassador access to non-ambassador endpoints

jmromer:
- Adds `before_action`s prohibiting access to `organized` endpoints for ambassador organizations (exceptions for ambassador controllers and `Organized::BaseController#root`

### DIFF
--- a/app/controllers/organized/admin_controller.rb
+++ b/app/controllers/organized/admin_controller.rb
@@ -2,5 +2,6 @@ module Organized
   class AdminController < Organized::BaseController
     before_filter :ensure_admin!
     skip_before_filter :ensure_member!
+    before_filter :ensure_not_ambassador_organization!
   end
 end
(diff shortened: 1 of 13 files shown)


sethherr:
We just had an ambassador request multi-serial search functionality - it would be nice to add multi-serial searching and give access to it.

app/controllers/organized/admin_controller.rb
```diff
@@ -2,5 +2,6 @@ module Organized
   class AdminController < Organized::BaseController
     before_filter :ensure_admin!
     skip_before_filter :ensure_member!
+    before_filter :ensure_not_ambassador_organization!
```
sethherr: This isn't inherited from base controller?

spec/controllers/organized/bulk_imports_controller_spec.rb
```diff
@@ -1,7 +1,6 @@
 require "spec_helper"
 
 describe Organized::BulkImportsController, type: :controller do
-  let(:root_path) { organization_bikes_path(organization_id: organization.to_param) }
```
sethherr: Good catch removing this, makes spec suite a lot cleaner.

app/controllers/organized/admin_controller.rb
jmromer: Ah thanks! That's vestigial.

PR #788, opened 2019-05-23
Fix admin subnavs to be more in line with bikes index subnav

Robertgervais:


app/views/admin/bikes/missing_manufacturer.html.haml
```diff
@@ -3,10 +3,13 @@
     %h1
       Multi manufacturer update
   .col-md-6
-    -if session[:missing_manufacturer_time_order]
-      = link_to "Normal view", missing_manufacturer_admin_bikes_path(reset_view: true), class: "btn btn-primary"
-    -else
-      = link_to "Time ordered view", missing_manufacturer_admin_bikes_path(time_ordered: true), class: "btn btn-primary"
+    %ul
+      -if session[:missing_manufacturer_time_order]
```
sethherr: I'd prefer if both of these showed all the time, and the one that was currently being viewed ("alphabetical ordering" or "creation ordering") had the active class.

app/views/admin/bikes/missing_manufacturer.html.haml
```diff
@@ -3,10 +3,13 @@
...
+        %li.nav-item
+          = link_to "Normal view", missing_manufacturer_admin_bikes_path(reset_view: true), class: "btn btn-primary"
```
sethherr: class should be `nav-link`,  not `btn`

app/views/admin/graphs/tables.html.haml
```diff
@@ -2,7 +2,9 @@
   .col-md-6
     %h1
       Bike Index records tables
-      %small
+  .col-md-6
```
sethherr: It would be great if the admin graphs index nav list was used here too. Probably using shared partial

app/views/admin/bulk_imports/show.html.haml
```diff
@@ -5,11 +5,12 @@
...
         \##{@bulk_import.id}
   .col-sm-8
-    %p
-      View #{link_to "all bulk imports", admin_bulk_imports_path}
-      - if @bulk_import.organization_id.present?
-        %em
-          \- View #{link_to "all #{@bulk_import.organization.name} imports", admin_bulk_imports_path(organization_id: @bulk_import.organization.to_param)}
+    %ul
+      %li.nav-item
+        View #{link_to "all bulk imports", admin_bulk_imports_path}
```
sethherr: Does this look weird if you add the `nav-link` class?

app/views/admin/graphs/_graphs_subnav.html.haml
```diff
@@ -0,0 +1,16 @@
+- location_count ||= false
```
sethherr: Good solution

app/views/admin/bulk_imports/show.html.haml
Robertgervais: Mmm, not really. It looks pretty much the same just a little slantier. It kinda seems like nav-link almost cancels out %em

UNADJUSTEDNONRAW_mini_47

I honestly think it might look better without the `%em` but obviously I'll defer to you. It looks like the two links are different for no actual reason.

sethherr: Yeah, no need for `%em`

PR #790, opened 2019-05-24
Ambassadors recovery flow

jmromer:
- Adds a "recover bike" link to the organized bike access panel for ambassador orgs
- Adds the letter_opener gem to view html emails in browser
- Introduces `BikeDecorator#creation_organization_name`

demo

Email:
Screen Shot 2019-05-24 at 3 34 57 PM

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -126,8 +126,8 @@ group :production do
 end
 
 group :development do
-  # gem 'test-unit', '~> 3.0'
   gem "bullet"
+  gem "letter_opener"
   gem "rerun"
 end
 
--- a/app/controllers/bikes/recovery_controller.rb
+++ b/app/controllers/bikes/recovery_controller.rb
@@ -24,8 +24,14 @@ module Bikes
     private
 
     def permitted_params
-      params.require(:stolen_record).permit(:date_recovered, :timezone, :recovered_description,
-                                            :index_helped_recovery, :can_share_recovery)
+      params.require(:stolen_record).permit(
+        :date_recovered,
+        :timezone,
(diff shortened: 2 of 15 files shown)


Gemfile
```diff
@@ -126,8 +126,8 @@ group :production do
 end
 
 group :development do
-  # gem 'test-unit', '~> 3.0'
   gem "bullet"
+  gem "letter_opener"
```
sethherr: You can use the mailer preview to view emails!

sethherr:
Sorry, I meant to mark "request for changes"

jmromer:
> I think we need to record who recovered the bike, if current_user is present

👍 Added in 3058d41 and 2ba7e43

app/controllers/bikes/recovery_controller.rb
```diff
@@ -24,8 +24,14 @@ def update
     private
 
     def permitted_params
-      params.require(:stolen_record).permit(:date_recovered, :timezone, :recovered_description,
-                                            :index_helped_recovery, :can_share_recovery)
+      params.require(:stolen_record).permit(
+        :date_recovered,
+        :timezone,
+        :recovering_user_id,
```
sethherr: I'd prefer if `recovering_user_id` was taken automatically in the controller, it could even be merged in here `merge(recovering_user_id: current_user&.id)` instead of put in a form in the view.

I don't _think_ anyone will be malicious about this, but why not protect from it just in case.

jmromer: Good call — updated in 0fe013e

PR #792, opened 2019-05-24
Set attributes for ambassador organizations

jmromer:
- Adds a `before_save` that sets sensible defaults for ambassador organization records
- Adds client-side logic to toggle related input fields based on "ambassador" being selected as the organization type

demo

demo2

### DIFF
--- a/app/javascript/packs/pages/admin/organization_form.js
+++ b/app/javascript/packs/pages/admin/organization_form.js
@@ -14,7 +14,7 @@ class OrganizationForm {
   }
 
   setEventListeners () {
-    this.$form.on("click", "#js-organization-type input.form-check-input", e => {
+    this.$form.on("change", "#js-organization-type input.form-check-input", e => {
       this.toggleAmbassadorFields(e.target);
     });
   }
--- a/app/javascript/packs/pages/admin/organization_form.js
+++ b/app/javascript/packs/pages/admin/organization_form.js
@@ -13,8 +13,14 @@ class OrganizationForm {
     this.setEventListeners();
   }
 
+  setEventListeners () {
+    this.$form.on("click", "#js-organization-type input.form-check-input", e => {
+      this.toggleAmbassadorFields(e.target);
+    });
+  }
+
   toggleAmbassadorFields(target) {
-    const inputFieldIds = [
+    const inputFields = [
       'organization_ascend_name',
       'organization_website',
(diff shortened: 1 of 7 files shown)


jmromer:
@sethherr These CodeClimate failures seem spurious to me but I'm open to suggestions if you have any.

sethherr:
Feel free to ignore codeclimate suggestions if they aren't useful

app/javascript/packs/pages/initializer.js
```diff
@@ -7,6 +7,7 @@ import BinxMapping from "./binx_mapping.js";
 import BinxAppOrgExport from "./binx_org_export.js";
 import BinxAppOrgMessages from "./binx_org_messages.js";
 import BinxAdmin from "./admin/binx_admin.js";
+import OrganizationForm from "./admin/organization_form.js";
```
sethherr: I'd prefer if this was imported and initialized in `binx_admin.js`

app/javascript/packs/pages/admin/organization_form.js
```diff
@@ -0,0 +1,81 @@
...
+          .find("input:checked")
+          .first()
+
+    this.toggleAmbassadorFields($selectedType);
+
+    this.setEventListeners();
+  }
+
+  setEventListeners () {
+    this.$form.on("click", "#js-organization-type input.form-check-input", e => {
```
sethherr: Shouldn't you bind to the change of the input, rather than the click?

app/models/organization.rb
```diff
@@ -254,4 +255,18 @@ def generate_access_token
...
+  def set_ambassador_organization_defaults
+    return unless kind == "ambassador"
+    self.show_on_map = false
+    self.lock_show_on_map = false
+    self.api_access_approved = false
+    self.approved = true
+    self.website = nil
+    self.ascend_name = nil
+    self.parent_organization_id = nil
+    self.locations = locations.reload.limit(1) if locations.count > 1
```
sethherr: I'm scared of this. 

I can see a scenario where we decide we need multiple locations and try to add them, and it takes a while to figure out why they aren't saving (or why a new one is replacing another one, or whatever happens).

I realize I told you to do this initially, but I think figuring out a way to lock it on the frontend will be good enough. And actually, you can just ignore the locations entirely because it's not clear what we're going to do with the locations.

PR #796, opened 2019-05-24
Ambassador landing page updates

sethherr:


sethherr:
Reasons behind the additions I made
- Added some additional explanation
- "Tasks" felt like it had negative connotations, so switch to calling them activities
- If you have checked a task off, eventually it should stop showing up above the fold
- Show the ambassador tasks even if you're a superuser, just for context

Issue #803, opened 2019-05-28
Switch to admin/user for ambassador details


sethherr:
- Add current `ambassador#show` details to `admin/user#edit`
- Add ambassador selection to `admin/users#index`
- Display ambassador tasks completed in `admin/users#index` table

sethherr:
We still haven't removed the ambassador controller and views, so this is still open. Talking with @jmromer about it

PR #804, opened 2019-05-29
admin users index interface

sethherr:
Fixes part of #803


app/views/admin/users/_table.html.haml
```diff
@@ -13,39 +19,56 @@
...
+                - user.memberships.each do |membership|
+                  - next if membership.organization.ambassador?
+                  - if first_org_rendered
+                    <small class="less-strong">|</small>
+                  - else
+                    - first_org_rendered = true
+                  = link_to membership.organization.name.gsub(" ", "\u00a0"), admin_users_path(organization_id: membership.organization_id), class: "small"
+            %td
+              - if user.ambassador?
+                - ambassador = Ambassador.find(user.id)
```
sethherr: This is pretty ugly. I'm sure there is a better way and I'm just being lazy. @jmromer, ideas?

jmromer: Any thoughts on this approach? https://github.com/bikeindex/bike_index/pull/806

PR #805, opened 2019-05-29
Revise users edit page

Robertgervais:


Robertgervais:
Image 5-28-19 at 8 35 PM

Robertgervais:
I added this vert scroll bike box (honestly I love vertical scroll now), because i was looking for something to fill that white space to the right and it seemed like the best thing.

Robertgervais:
Image 5-29-19 at 4 50 PM


Fake user

Robertgervais:
#803

app/views/admin/users/edit.html.haml
```diff
@@ -24,84 +22,129 @@
...
+          %tr
+            %th Task
+            %th Completed at
+        %tbody
+          - ambassador.ambassador_task_assignments.each do |task|
+            %tr
+              %td
+                %strong= task.title
+                %br
+                = task.description_html.html_safe
```
sethherr: I don't think you need the description. That will make things overlong.

app/views/admin/ambassadors/_ambassador.html.haml
```diff
@@ -0,0 +1,31 @@
+%h1= "Ambassador: #{@ambassador.username}"
```
sethherr: What's  this file for?

app/views/admin/users/edit.html.haml
```diff
@@ -24,84 +22,129 @@
...
-      .form.form-horizontal
-        %h3
-          User attributes
-        .control-group.padded
-          %lable.control-label
+        %tr
+          %td
             Created
-          .controls.convertTime
+          %td.convertTime
```
sethherr: Nice! I hadn't tried putting the convertTime on `td`s, it makes sense and simplifies things though!

app/views/admin/users/edit.html.haml
```diff
@@ -24,84 +22,129 @@
...
-      .form.form-horizontal
-        .control-group
-          = f.label :confirmed, class: "control-label"
-          .controls
-            = f.check_box :confirmed, disabled: @user.confirmed
-        .control-group
-          = f.label :superuser, class: "control-label"
-          .controls
-            = f.check_box :superuser
+      = ambassador.avatar
```
sethherr: avatar should only be included if it isn't the default image

Also, this table list can always be displayed. It isn't exclusive to ambassadors.

sethherr: Also, this could be put into the table list

Robertgervais:
Image 5-29-19 at 7 15 PM

Robertgervais:
#175

PR #806, opened 2019-05-29
Display admin ambassador tasks when ambassadors filter selected

jmromer:
In the admin users display, displays the "ambassador tasks" column conditionally if the "ambassadors" search filter is selected.

Screen Shot 2019-05-29 at 8 47 38 AM
Screen Shot 2019-05-29 at 8 47 56 AM
Screen Shot 2019-05-29 at 8 48 04 AM


jmromer:
I don't follow the first point, but the second seems dispositive, so I'll close for now :)

PR #807, opened 2019-05-29
remove codeclimate

sethherr:
- We aren't really using the PR features (eg ["Feel free to ignore codeclimate suggestions if they aren't useful"](https://github.com/bikeindex/bike_index/pull/792#issuecomment-495812808))
- I just looked at the codeclimate page and didn't get any value (even from its test coverage view)
- I think it's vanity metrics, not meaningful, and just a distraction

Please let me know if you see a reason to keep it! If it is providing any value, it should stick around,

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -109,9 +109,6 @@ workflows:
   commit:
     jobs:
       - build
-      - upload-coverage:
-          requires:
-            - build
       - deploy_production:
           requires:
             - build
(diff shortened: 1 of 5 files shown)


jmromer:
@sethherr https://chrome.google.com/webstore/detail/code-climate/phgahogocbnfilkegjdpohgkkjgahjgk?hl=en

sethherr:
Looking at the coverage when I'm looking at the files seems like enough of a win to keep it for now.

I'm closing. We can definitely update if things change.

PR #812, opened 2019-05-30
Refresh homepage, landing pages design

jmromer:
Updates the design and layout for the homepage and landing pages

- Refactors to consistify some naming "from advocacy organization" to "community groups" and from "shop/shops" to "bike_shops"

Accessible from http://85479dc0.ngrok.io

## Demos

demo
mobile


app/views/welcome/index.html.haml
```diff
@@ -1,44 +1,52 @@
...
-                .form-group.col-md-6
-                  = f.label :email, class: 'sr-only'
-                  = f.email_field :email, placeholder: 'email', class: 'form-control'
-                .form-group.col-md-6
-                  = f.label :password, class: 'sr-only'
-                  = f.password_field :password, placeholder: 'password', class: 'form-control'
+      .root-landing-header-text
+        %h2
+          Over
+          = number_with_delimiter(@recovery_count)
```
sethherr: Should be `Counts.recoveries_value` - and `@recovery_count` can be fully removed.

app/views/welcome/index.html.haml
```diff
@@ -1,44 +1,52 @@
...
+          Bikes Recovered
+        %h1.upcase
+          Bike registration that works
+        %h2
+          = link_to new_organization_path do
+            Find out how &raquo;
```
sethherr: This is actually targeting consumers (not organizations).

Currently, it should scroll to "BUT HOW DOES IT WORK?" on the homepage itself.

spec/requests/landing_pages_request_spec.rb
```diff
@@ -0,0 +1,39 @@
+require "spec_helper"
+
+RSpec.describe LandingPagesController, type: :request do
```
sethherr: Yay!! 🎉

spec/spec_helper.rb
```diff
@@ -17,7 +17,6 @@
 ENV["RAILS_ENV"] ||= "test"
 require File.expand_path("../../config/environment", __FILE__)
 require "rspec/rails"
-# require 'rspec/autorun'
```
sethherr: Nice. No idea why this was hanging around.

app/views/welcome/index.html.haml
jmromer: `@recovery_count` is cached with a daily refresh -- let me know if you want to get rid of that

sethherr: Yeah, getting rid of it is fine. The `Counts.recoveries_value` is cached (calculated and stored in redis) via [update_counts_worker](https://github.com/bikeindex/bike_index/blob/master/app/workers/update_counts_worker.rb)

app/views/shared/_footer_revised.html.haml
```diff
@@ -2,31 +2,44 @@
...
-            %li
-              = active_link "Sign up a shop or organization", new_organization_path
-            %li
-              = active_link "BikeHub", "https://bikehub.com"
-        %nav.col-md-3.pull-right
-          %h4
-            = active_link "Dev and Design Resources", dev_and_design_path
+            %li= active_link "Donate", support_bike_index_path
+            %li= active_link "Store", store_path
+            %li= active_link "Ambassadors", ambassadors_current_path
```
sethherr: I think this should link to the ambassadors how to for now, at least until we update the ambassadors page. I think the title "Ambassadors" is still fine

app/assets/stylesheets/revised/pages/landing/_home.scss
```diff
@@ -281,11 +432,13 @@
 }
 
 .root-landing-who {
+  background-color: #f9f9f9;
```
sethherr: I think, because of the white background in the images, it would be great if the background here was white too.

The quick and easy fix I see that makes the alternating background still exist is switching the "how to" to be `#f9f9f9` and then making this and the recoveries white. It looks like you'd have to add some padding to the bottom of the "how to" section.

Of course, if you see something easier go for that instead. I think the alternating background is nice and subtly breaks up the page and so makes sense to continue.

app/assets/stylesheets/revised/pages/landing/_home.scss
```diff
@@ -12,18 +12,163 @@
...
+    background-color: black;
+    color: #fff;
+    font-weight: bold;
+    padding: 10px 15px;
+    text-decoration: none;
+  }
+}
+
+.center-navbar-signup-link-container {
+  display: inline-block;
```
sethherr: Something needs to happen differently when the collapse is opened - eg this should hide, or be position absolute.

Image from iOS

app/assets/stylesheets/revised/pages/landing/_home.scss
```diff
@@ -12,18 +12,163 @@
...
+  display: inline;
+  font-size: 0.8rem;
+  font-size: 12px;
+  height: 37px;
+  line-height: 18px;
+  padding: 8px 15px;
+  text-align: center;
+}
+
+@media screen and (min-width: 501px) {
```
sethherr: The collapse actually shows up at `@media (max-width: 991px)`, so this should probably show up then too

app/views/welcome/index.html.haml
```diff
@@ -1,44 +1,52 @@
...
-                  = f.password_field :password, placeholder: 'password', class: 'form-control'
+      .root-landing-header-text
+        %h2
+          Over
+          = number_with_delimiter(Counts.recoveries_value)
+          Bikes Recovered
+        %h1.upcase
+          Bike registration that works
+        %h2
+          = link_to "#how-it-works", class: "scroll-to-ref", data: { offset: -80 } do
```
sethherr: Actually, sorry - just reviewing it with Craig, let's change this to link to have the text "Register now" and link to `new_bike_path`

app/assets/stylesheets/revised/pages/landing/_home.scss
jmromer: Updated in 39746d16

app/assets/stylesheets/revised/pages/landing/_home.scss
jmromer: Updated in 6aa173b5

app/assets/stylesheets/revised/pages/landing/_home.scss
jmromer: Updated in d7e17042

app/assets/stylesheets/revised/pages/landing/_home.scss
```diff
@@ -12,18 +12,177 @@
...
+  text-align: center;
+  width: 75%;
+}
+
+.center-navbar-signup-link {
+	background-color: black;
+	border-radius: 4px;
+	border: 1px solid #3498DB;
+	width: 85px;
+  box-sizing: border-box;
```
sethherr: I believe border box is applied to everything via a psuedo `::before` element, and this isn't necessary

app/assets/stylesheets/revised/pages/landing/_home.scss
```diff
@@ -12,18 +12,177 @@
...
+  background-repeat: no-repeat;
+}
+
+@include media-breakpoint-up(lg) {
+  .bg-image-bike-hero {
+    background-position: bottom;
+  }
+
+  .bg-image-bike-shops {
+    background-position: center top;
```
sethherr: Since these are all different, I'm guessing you checked where the background looked best at different screen sizes, which is above and beyond!

app/assets/stylesheets/revised/pages/landing/_signup.scss
```diff
@@ -41,47 +64,58 @@ $landing-page-targets: "shop", "school", "advocacy", "law_enforcement",
...
 
-      img {
-        display: block;
-        width: 100%;
-        height: auto;
-      }
+.landing-page {
+  @include media-breakpoint-down(lg) {
+    .banner-header {
+      padding-top: 100px !important;
```
sethherr: Why do you need the `!important`s here?

config/locales/en.yml
```diff
@@ -35,10 +35,17 @@ en:
     stolen_howworks: "How it works"
     stolen_merging: "Stolen bike registry"
     landing_pages_show: "%{organization} Bike Registration"
-    landing_pages_for_shops: Bike Index for Shops
-    landing_pages_for_advocacy: Bike Index for Advocacy Organizations
+    landing_pages_ambassadors_current: Bike Index Ambassadors
```
sethherr: **much** better place to put these. Good choice 👍

app/assets/stylesheets/revised/pages/landing/_signup.scss
jmromer: Good nudge thanks -- I couldn't find where in our stylesheets the overridden rules are defined, but I refactored to work around that without the `!important`s

PR #822, opened 2019-05-30
Admin invites revision

Robertgervais:


app/controllers/admin/organization_invitations_controller.rb
```diff
@@ -45,7 +48,7 @@ def create
       end
     else
       flash[:error] = "Oh no! This organization has no more invitations. Email contact@bikeindex.org for help"
-      redirect_to root_url
+      redirect_to new_admin_organization_invitation_url(@organization_invitation.id, organization_id: @organization_invitation.organization.to_param)
```
sethherr: I think this could actually just redirect to the admin organization page

PR #825, opened 2019-05-31
Add more recovery stories

sethherr:
Craig wants more recovery stories, so I added some more to the page.

He also thought it would be impressive if we displayed the total number of pages in the pagination. I don't remember exactly why we switched to window 0, but [I complimented @jmromer on doing it](https://github.com/bikeindex/bike_index/pull/710#discussion_r277741239), so there was definitely a reason.

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -42,7 +42,7 @@ class WelcomeController < ApplicationController
 
   def recovery_stories
     page = params[:page] || 1
-    @per_page = params[:per_page] || 10
+    @per_page = params[:per_page] || 50
 
     @recovery_displays = RecoveryDisplay.page(page).per(@per_page)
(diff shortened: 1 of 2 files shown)


jmromer:
> why we switched to window 0

Does that impact the total number of pages? 🤔 

Window 0 omits the "inner window" of pagination links, to economize on space since they don't wrap nicely (modulo the layout at the time that shipped):
https://github.com/kaminari/kaminari#specifying-the-inner-window-size-4-by-default

jmromer:
`window: 5` doesn't wrap great at mobile breakpoints, but a smaller value might work if the intent is to show the total number of pages:

Screen Shot 2019-05-31 at 12 54 54 PM

sethherr:
Thanks for the suggestion and the look @jmromer. Switched to window 3.

Issue #739, opened 2019-05-02
Improve recovery display proofing flow


sethherr:
Right now there are too many moving parts in the recovery_display flow.

We need to streamline it to improve efficiency and make it easier for more people to share the responsibility for proofing and posting them.

Steps I see for this:

- Add a `recovery_display_status` enum to `StolenRecord` , with values of something like `not_eligible`, `waiting_on_decision`, `displayed` & `not_displayed` (not eligible including recoveries that don't have an image, that we didn't help, or that aren't open to being shared).
- Make the default recoveries index page only show recoveries that are `waiting_on_decision`
- Improve the recovery_display form page
  - add text about our standards for the text (e.g. length, no swears or racism)
  - add a live character count
  - make it default to using the existing bike image
  - remove link column (it isn't used), and remove bike_url field (it's confusing, also isn't used)
  - add the full text of the recovery in a non-editable text block, so you can see the original as you edit the content you're posting
- Make the flow work smoothly - clicking on "skip this recovery" returns you to the `waiting_on_decision` recoveries, clicking on "display this recovery" probably slides down the recovery display form on that page, so you don't have to switch between pages.
- After shipping, migrate all existing stolen records to use the new recovery_display_status attribute. In a separate PR remove `recovery_posted` from stolen record

---

Additional fix that should happen in here - right now I don't think you can update the checkboxes on the recovery edit page. When working on this page, a submit button should be added too.

sethherr:
resolved by @Robertgervais

PR #829, opened 2019-05-31
improve recovery displays workflow

Robertgervais:


Robertgervais:
https://github.com/bikeindex/bike_index/blob/4865f5635b56d6fefcf6bdaa11e86c76430728c2/spec/models/stolen_record_spec.rb

app/controllers/bikes/recovery_controller.rb
```diff
@@ -12,7 +12,9 @@ def edit
     end
 
     def update
-      if @stolen_record.add_recovery_information(permitted_params)
+      update_params = permitted_params.merge(recovering_user: current_user)
```
sethherr: Use the way this is done on master - it's in the permitted params.

app/controllers/bikes/recovery_controller.rb
```diff
@@ -12,7 +12,9 @@ def edit
...
+      if @stolen_record.add_recovery_information(update_params)
+        @stolen_record.update_attributes(recovery_display_status: 3)
```
sethherr: Should be a string

app/controllers/bikes/recovery_controller.rb
sethherr: View the changes done here: https://github.com/bikeindex/bike_index/pull/846/files

app/models/stolen_record.rb
```diff
@@ -168,14 +185,26 @@ def tsv_row(with_article = true, with_stolen_locations: false)
     row
   end
 
+  def calculated_recovery_display_status
+    return "not_eligible" unless can_share_recovery
+    return "not_displayed" if not_displayed?
+    return "displayed" if displayed?
```
sethherr: Are you going to set this to `displayed` somewhere else in the future?

app/models/stolen_record.rb
```diff
@@ -168,14 +185,26 @@ def tsv_row(with_article = true, with_stolen_locations: false)
...
   def add_recovery_information(info = {})
     info = ActiveSupport::HashWithIndifferentAccess.new(info)
     self.date_recovered = TimeParser.parse(info[:date_recovered], info[:timezone]) || Time.now
     update_attributes(current: false,
                       recovered_description: info[:recovered_description],
                       recovering_user_id: info[:recovering_user_id],
                       index_helped_recovery: ("#{info[:index_helped_recovery]}" =~ /t|1/i).present?,
-                      can_share_recovery: ("#{info[:can_share_recovery]}" =~ /t|1/i).present?)
+                      can_share_recovery: ("#{info[:can_share_recovery]}" =~ /t|1/i).present?,
+                      recovery_display_status: calculated_recovery_display_status)
```
sethherr: you don't need to do this, it happens automatically in the callback

app/models/stolen_record.rb
Robertgervais: No, the only place it gets set to displayed, currently, is when the display get's created.

app/models/stolen_record.rb
```diff
@@ -100,7 +109,15 @@ def self.locking_defeat_description_select
     locking_defeat_description.map { |l| [l, l] }
   end
 
-  before_save :set_phone, :fix_date, :titleize_city, :update_tsved_at
+  before_save :set_calculated_attributes
```
sethherr: The new approach I'm taking with putting all callbacks (ie `before_save ...`) up at the top, after the scopes. It makes it easier to find. You don't need to move the method, but I would like it if you moved this up there to the top!

app/models/stolen_record.rb
sethherr: Great! That makes sense.

Robertgervais:
Responded. I took `recovery_display_status` out of `add_recovery_info`, and all of THE TESTS I WROTE are still green, but it still won't let me set enum status in console when I remove this. I know what before_save is supposed to do, and it makes sense not to have it in two places, but without it I can't seem to get the `waiting_on_decision` status set and I'm not sure why.

Robertgervais:
#739

PR #835, opened 2019-06-03
Display UI feedback when marking ambassador activities completed

### LINKED ISSUE #801: Feedback on checking ambassador task complete
Two things that would improve this interaction:

- Display alert when one is clicked (probably using our existing [alert coffee file](app/assets/javascripts/revised/components/alerts.coffee))
- Update the leaderboard number when there is a change

jmromer:
- Updates the organized/ambassador_dashboard UI to display feedback when a 
- Changes the task completion checkbox to a "Mark complete" / "Mark pending" button
- Displays an appropriate flash message on success / failure
- Displays an updated leaderboard when marking an item complete
- Makes the ambassador association subtype-specific in order to return Ambassador objects from association methods and only permit Ambassador instances when persisting assignments: a78e79f

demo

Resolves  #801


app/controllers/organized/ambassador_dashboards_controller.rb
```diff
@@ -11,10 +11,14 @@ def show
           .includes(:ambassador_task_assignments)
           .sort_by { |ambassador| -ambassador.percent_complete }
 
-      @ambassador_task_assignments =
-        current_user
-          .ambassador_task_assignments
-          .task_ordered
+      if current_user.ambassador?
+        current_ambassador = Ambassador.find(current_user.id).decorate
+        @suggested_activities = current_ambassador.suggested_activities.decorate
```
sethherr: Could you put `task_ordered` scope on this?

jmromer: It's set in the decorator:
```rb
# app/decorators/ambassador_decorator.rb L5-10 (7839d064)

def ambassador_task_assignments
  ambassador
    .ambassador_task_assignments
    .includes(:ambassador_task)
    .task_ordered
end
```

app/decorators/ambassador_task_assignment_decorator.rb
```diff
@@ -0,0 +1,18 @@
...
+  def button_to_toggle_completion_status(current_user, current_organization)
+    return unless current_user.ambassador?
+
+    is_complete = ambassador_task_assignment.completed?
+    button_label = is_complete ? "Mark Pending" : "Mark Complete"
+
+    h.button_to(
+      button_label,
+      h.organization_ambassador_task_assignment_url(current_organization, ambassador_task_assignment),
+      method: :put,
```
sethherr: Would you add a bootstrap button class to this too?

app/models/ambassador_task.rb
```diff
@@ -1,6 +1,6 @@
 class AmbassadorTask < ActiveRecord::Base
   has_many :ambassador_task_assignments
-  has_many :users, through: :ambassador_task_assignments
+  has_many :ambassadors, through: :ambassador_task_assignments
```
sethherr: Nice! Makes sense.

app/decorators/ambassador_task_assignment_decorator.rb
jmromer: Good catch, thanks! Updated

Screen Shot 2019-06-03 at 3 16 59 PM

app/models/ambassador_task.rb
```diff
@@ -14,14 +14,14 @@ def description_html
...
-    ambassador_task_assignments.create(user: user)
+  def assign_to(ambassador)
+    ambassador_task_assignments.create(ambassador: ambassador)
   end
 
   # Assigns the task to all ambassadors, if not already assigned
   def ensure_assigned_to_all_ambassadors!
     Ambassador.find_each do |ambassador|
-      next if ambassador_task_assignments.exists?(user: ambassador)
+      next if ambassador_task_assignments.exists?(ambassador: ambassador)
```
sethherr: Does this instantiate every ambassador? aka, will this generate performance issues and should this use pluck & array comparison outside of the `find_each`, and batch generate the tasks - for when we have 100 ambassadors?

spec/models/ambassador_task_assignment_spec.rb
```diff
@@ -5,48 +5,52 @@
...
-    let!(:ambassador_task_assignment2) { FactoryBot.create(:ambassador_task_assignment, user: ambassador, completed_at: Time.now - 1.day) }
-    it "is complete" do
-      ambassador.reload
-      expect(ambassador.ambassador_task_assignments.completed).to eq([ambassador_task_assignment2])
-      expect(ambassador.ambassador_task_assignments.locked_completed.pluck(:id)).to eq([ambassador_task_assignment2.id])
-      expect(ambassador.ambassador_task_assignments.pending_completion.pluck(:id)).to eq([ambassador_task_assignment1.id])
-      ambassador_task_assignment1.update_attributes(completed_at: Time.now - 1.minute)
-      ambassador.reload
-      expect(ambassador.ambassador_task_assignments.pending_completion.pluck(:id)).to eq([ambassador_task_assignment1.id])
+  describe "#completed" do
```
sethherr: nit: I know this isn't TDD doctrine, but for tests like this where you're adding a bunch of objects in the database, in the interest of preventing the test suite from slowing down, I lean towards trying to consolidate and only create the objects once - unless -

- Creating them multiple times makes the tests easier to understand
- The different objects don't overlap much, so there would be little (or no) time savings
- Too many things are tested, so when modifying or adding to anything in the app it breaks the fat tests
- ... or anything else tilts the balance at all

I'd **much** prefer comprehensive, readable tests to incomplete tests. This is mainly a justification for why I wrote just one test block originally.

app/models/ambassador_task.rb
jmromer: We could definitely do away with instantiating models but most of the time is spent in the db — for 100 users this'll take around 15 seconds (extrapolating based on the 29 ambassador records I have locally). I think it would be a good idea to (a) move this logic into a background job or (b) bulk insert

PR #836, opened 2019-06-03
Update omniauth, jazz_fingers

jmromer:
Eliminates deprecation warnings from outdated versions of omniauth, jazz_fingers.

### Before

```
% bin/rails c

W, [2019-06-02T10:15:34.073533 #83787]  WARN -- : You are setting a key that conflicts with a built-in method OmniAuth::AuthHash::InfoHash#name defined at ~/.asdf/installs/ruby/2.5.1/lib/ruby/gems/2.5.0/gems/omniauth-1.3.2/lib/omniauth/auth_hash.rb:34. This can cause unexpected behavior when accessing the key as a property. You can still access the key via the #[] method.
~/.asdf/installs/ruby/2.5.1/lib/ruby/gems/2.5.0/gems/jazz_fingers-5.0.0/lib/jazz_fingers/prompt.rb:40:
warning: method Pry#input_array is deprecated. Use Pry#input_ring instead
Loading development environment (Rails 4.2.11)
2.5.1 (bikeindex)[1] »
```

### After

```
bike_index task-completion-feedback % bin/rails c
Loading development environment (Rails 4.2.11)
2.5.1 (bikeindex)[1] »
```

See: 
https://github.com/omniauth/omniauth/issues/872
https://github.com/plribeiro3000/jazz_fingers/issues/20

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -38,7 +38,7 @@ gem "libv8", "~> 3.16.14.7"
 gem "mini_magick" # a smaller implementation of rmagick, required for rqrcode
 gem "money-rails"
 gem "nokogiri", "~> 1.8.1"
-gem "omniauth", "~> 1.3.2"
+gem "omniauth", "~> 1.6"
 gem "omniauth-facebook"
 gem "omniauth-strava"
 gem "paranoia"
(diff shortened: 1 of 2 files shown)


sethherr:
Why does it appear that the most recent commit didn't run on CircleCI?

Screen Shot 2019-06-03 at 10 47 43 AM

jmromer:
That happens sometimes when rebasing -- I think GH must cache commits, so if you re-order or edit them they can show up out of order on PRs. I rebased again and re-pushed to get CI running on the omniauth commit just in case though 👍

PR #837, opened 2019-06-03
Add admin/ambassador_task_assignments controller

### LINKED ISSUE #802: Admin display ambassador tasks
Add a new `/admin` controller that displays ambassador tasks that have been completed.

This should be sortable by organization, task, user and completion date - a good example of how this has been done in the past is the [bike_codes_controller](https://github.com/bikeindex/bike_index/blob/master/app/controllers/admin/bike_codes_controller.rb).

jmromer:
Adds an admin page for completed ambassador task assignments, sortable by column.

demo

Resolves #802


spec/controllers/admin/ambassador_task_assignments_controller_spec.rb
```diff
@@ -0,0 +1,159 @@
+require "spec_helper"
+
+RSpec.describe Admin::AmbassadorTaskAssignmentsController, type: :request do
```
sethherr: If this is a request spec (which it appears to be?) I think it should be in the request specs folder

app/controllers/admin/ambassador_task_assignments_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+    sort_criterion = params.fetch(:sort, "").to_sym
+    direction = params.fetch(:direction, "").to_sym
+
+    case sort_criterion
+    when :completed_at
+      assignments.reorder(completed_at: direction)
+    when :organization_name
+      assignments
+        .to_a
+        .sort_by!(&:organization_name)
```
sethherr: Why can’t you use a join here, and keep this an active record collection, rather than an array?

spec/controllers/admin/ambassador_task_assignments_controller_spec.rb
jmromer: How would you feel about consolidating both controller and request specs in `spec/controllers` (setting `type: :request` as needed) and refactoring the former to the latter incrementally? It'd be helpful to have both sets of tests in one place.

app/controllers/admin/ambassador_task_assignments_controller.rb
jmromer: It's not the case that we can't. It would be a complex join because of the way the data is associated (through an associated ambassador and ambassador memberships), but it's feasible. I'll update.

spec/controllers/admin/ambassador_task_assignments_controller_spec.rb
sethherr: Let's move the controller specs to "requests", rather than the other way around, since that is what I want to move toward.

This should be in a separate PR.

sethherr: Added an issue for this # #845

app/controllers/admin/ambassador_task_assignments_controller.rb
sethherr: Honestly, with regard to the join on organization name - I don't think it's crucial that it sorts alphabetically. It could be sorted by organization_id. And frankly, you could probably even skip sorting by organization altogether.

The reason I believe having it always be active record associations would be nice is that it's more performant and it will enable us to add additional scopes on. Which we will need to do. For example: searching just for a specific task, just for one user, just for one organization.

app/controllers/admin/ambassador_task_assignments_controller.rb
```diff
@@ -0,0 +1,44 @@
...
+  end
+
+  def sorted_assignments
+    assignments =
+      AmbassadorTaskAssignment
+        .includes(:ambassador, :ambassador_task)
+        .completed
+
+    sort_criterion = params.fetch(:sort, "").to_sym
+    direction = params.fetch(:direction, "").to_sym
```
sethherr: There are already helper methods (in `SortableTable`) for these things, `sort_column` and `sort_direction`

sethherr: Which do two important things:

1. Make sure we only execute the whitelisted list of sequel sorts
2. provide a default

app/controllers/admin/ambassador_task_assignments_controller.rb
jmromer: > It could be sorted by organization_id.
Not if we want the user to be able to sort by the organization's name, which how we're representing orgs in the UI. I'm not sure if there's a better way to represent organizations in the UI, and I don't think it would be intuitive to sort values in the column in a way that's not transparent to the user, but I'm open to suggestions.

The complexity here is added from the fact that a user can have multiple ambassador group memberships, so the following is insufficient:
```
assignments.reorder("organizations.name #{direction}")
```

Making progress though.

> You could probably even skip sorting by organization altogether.
So do we want to drop this feature altogether? I'm taking my cues from #802

> The reason I believe having it always be active record associations would be nice is that it's more performant and it will enable us to add additional scopes on. 

I agree, modulo time-to-deployment :)

sethherr: Write a comment that it would add extra time and doesn't deliver sufficient value 😉 and then ignore it.

spec/factories/memberships.rb
```diff
@@ -8,7 +8,7 @@
 
     factory :membership_ambassador do
       before(:create) do
-        Membership.skip_callback(:create, :before, :assign_ambassador_tasks!)
+        Membership.skip_callback(:create, :after, :assign_ambassador_tasks!)
```
sethherr: Good catch!

spec/requests/admin/ambassador_task_assignments_controller_spec.rb
```diff
@@ -0,0 +1,222 @@
+require "spec_helper"
+
+RSpec.describe Admin::AmbassadorTaskAssignmentsController, type: :request do
```
sethherr: One trend I've seen recently, with request specs, is making multiple requests in a single spec.

I've previously been somewhat skeptical, because it seems like it muddies up the test isolation - but having read a bunch more request specs, I'm not convinced it's actually terrible.

This isn't something you need to do here, just something I've been thinking about

app/controllers/admin/ambassador_task_assignments_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  private
+
+  def sortable_columns
+    %w[completed_at organization_name task_title ambassador_name]
+  end
+
+  def filter_params
+    {}.tap do |h|
```
sethherr: Why use `tap` and not `slice`?

sethherr: Or not just doing the filtering in the query interface in the model?

jmromer: Updated to use `slice`, that's a little more succinct. The query method doesn't need to know that the controller expects filter param keys to be prepended with `search_`

> Connascence of Meaning (CoM) or Connascence of Convention (CoC)
> Connascence of meaning is when multiple components must agree on the meaning of particular values. Returning integers 0 and 1 to represent false and true, respectively, is an example of this form of connascence.
https://en.wikipedia.org/wiki/Connascence#Connascence_of_Meaning_(CoM)_or_Connascence_of_Convention_(CoC)

Issue #842, opened 2019-06-04
Ambassador Decorator bug


sethherr:
Draper::UninferrableDecoratorError: Could not infer a decorator for ActiveRecord::Base.

https://app.honeybadger.io/projects/35931/faults/50329990

PR #843, opened 2019-06-04
Fix ambassador dashboard exception

jmromer:
Fixes a UninferrableDecoratorError  / NoMethodError exception that occurs when accessing the organized ambassador dashboard as a super admin who is not in the ambassador organization.

Fixes https://github.com/bikeindex/bike_index/issues/842
https://app.honeybadger.io/projects/35931/faults/50329990#notice-comments

### DIFF
--- a/spec/requests/organized/ambassador_dashboards_request_spec.rb
+++ b/spec/requests/organized/ambassador_dashboards_request_spec.rb
@@ -1,7 +1,6 @@
 require "spec_helper"
 
 describe "Organized::AmbassadorDashboardsController" do
-  let(:base_url) { "/o/#{organization.to_param}/ambassador_dashboard" }
   # Request specs don't have cookies so we need to stub stuff if we're in request specs
   # This is suboptimal, but hey, it gets us to request specs for now
   before { allow(User).to receive(:from_auth) { user } }
@@ -12,7 +11,7 @@ describe "Organized::AmbassadorDashboardsController" do
     describe "index" do
       let(:organization) { FactoryBot.create(:organization) }
       it "redirects to the user homepage" do
-        get base_url
+        get organization_ambassador_dashboard_path(organization)
         expect(response).to redirect_to(/user_home/)
       end
     end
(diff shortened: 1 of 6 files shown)


spec/requests/organized/ambassador_dashboards_request_spec.rb
```diff
@@ -34,28 +33,66 @@
...
-        expect(assigns(:ambassadors).count).to eq(1)
+        expect(assigns(:ambassadors).count).to eq(3)
         expect(response).to render_template(:show)
       end
     end
 
     describe "resources" do
       it "renders the ambassador resources" do
-        get "#{base_url}/resources"
+        get resources_organization_ambassador_dashboard_path(organization)
```
sethherr: Why switch to routes instead of explicit paths?

sethherr: The benefits I see to using explicit paths is that it adds one extra bit of test coverage.

I can see an argument for using paths instead, in that we generally use them in the code.

jmromer: 1. In the preexisting tests, for consistency with the new tests
2. In the new tests, because url helpers to provide a layer of indirection so that we're not propagating knowledge of the routes throughout the app. (Suppose, for example, we want to refactor underscores to dashes in routes for the sake of SEO.)

I'm not sure I see where the added test coverage would come from, but I'll go ahead ship and iterate since this is a bug fix. Happy to standardize on something and refactor for uniformity.

sethherr: The "extra test coverage" is a little bit of a stretch, because I'm not sure that it actually is useful test coverage (like shoulda matchers) - but by writing out the route as a URL string, we get [routing specs](https://relishapp.com/rspec/rspec-rails/v/3-8/docs/routing-specs).

There are [two routing specs](https://github.com/bikeindex/bike_index/tree/master/spec/routing), both for complicated routing parameters, both were used as unit tests when I was verifying things worked as expected; if we switched fully over to request specs and wrote out URLs, these tests wouldn't be necessary.

Merging this was definitely the right choice. I'm adding additional explanation to the request spec migration issue #845

PR #847, opened 2019-06-05
Mobile topbar improvements

### LINKED ISSUE #818: Fix bug with topbar on mobile
On android, with long organization names, the top navbar gets too tall and make accessing the top parts of the site difficult.

sethherr:
Fix #818


sethherr:
Updated topbar with a name that is too large:

Screen Shot 2019-06-04 at 6 20 18 PM

sethherr:
Going to ship and iterate because this is a bug fix. But yes, makes sense to do this via media queries instead.

PR #857, opened 2019-06-06
Add alert for law enforcement organizations without unstolen

sethherr:


app/models/invoice.rb
```diff
@@ -30,6 +30,10 @@ def self.friendly_find(str)
     where(id: str).first
   end
 
+  def self.feature_slugs
+    all.map(&:feature_slugs).flatten.uniq
```
jmromer: Looks like these model instantiations (`all.map`) could be avoided:

```rb
Invoice.includes(:paid_features).pluck(:feature_slugs).flatten.uniq.compact

SELECT feature_slugs 
FROM "invoices" 
LEFT OUTER JOIN "invoice_paid_features" 
ON "invoice_paid_features"."invoice_id" = "invoices"."id" 
LEFT OUTER JOIN "paid_features" 
ON "paid_features"."id" = "invoice_paid_features"."paid_feature_id"
```

sethherr: Yeah. That’s why I moved the method into invoice, because I figured it would eventually be refactored.

Issue #809, opened 2019-05-30
Separate out user bikes


sethherr:
The Admin user page is getting unwieldy, particularly considering there is pagination for bikes.

Instead of paginating users bikes on the users page, there should be a few bikes rendered, and if there are too many, display the count and link to the bikes index page with a `search_user_id` parameter. Also, add the `search_user_id` functionality to admin bikes controller

PR #858, opened 2019-06-06
Separate large lists of bikes from users edit page

Robertgervais:


Robertgervais:
Image 6-6-19 at 1 23 PM

Robertgervais:
The controller code I added isn't super elegant, welcome to suggestions, works nicely otherwise.

Robertgervais:
#809

app/controllers/admin/bikes_controller.rb
```diff
@@ -142,6 +142,10 @@ def matching_bikes
     return @matching_bikes if defined?(@matching_bikes)
     bikes = Bike.unscoped
     # do example up here because it unscopes
```
sethherr: This comment is now separated from its applicable place. Which is critical in this case, because users won't work with the example query. You will need to skip example if `search_user_id`

app/controllers/admin/users_controller.rb
```diff
@@ -15,15 +15,14 @@ def show
...
-      bikes = @user.rough_approx_bikes
+    if @user.rough_approx_bikes.count
+      bikes = @user.rough_approx_bikes.limit(10)
     else
-      bikes = @user.bikes
+      bikes = @user.bikes.limit(10)
     end
-    @bikes = bikes.reorder(created_at: :desc).page(page).per(per_page)
+    @bikescount = @user.bikes.count
+    @bikes = bikes.reorder(created_at: :desc)
```
sethherr: You can put the limit here rather than twice above.

PR #860, opened 2019-06-07
Create ambassador task assignments on task / membership creation asynchronously

### LINKED ISSUE #839: Refactor AmbassadorTask after_create hook
Refactor AmbassadorTask after_create hook to:
1. Not instantiate models wherever it can be avoided, and
2. Bulk insert assignments, or
3. Create assignments in a background job

jmromer:
- Moves the logic for assigning a newly created `AmbassadorTask` to all existing `Ambassador`s from an AR callback in `AmbassadorTask` to `AmbassadorTaskAssignmentCreator.assign_task_to_all_ambassadors`, which is enqueued upon creation of a new task.

- Moves the logic for assigning existing tasks to a new `Ambassador` from an AR callback in `Membership` to `AmbassadorTaskAssignmentCreator.assign_all_ambassador_tasks_to`, which is enqueued upon creation of a new ambassador membership.

- Assignments are created in bulk (`activerecord-import` added in d1a469d)

```
2.5.1 (bikeindex)[5] » AmbassadorTaskAssignmentCreator.assign_task_to_all_ambassadors(AmbassadorTask.fifth)
...
AmbassadorTaskAssignment Create Many Without Validations Or Callbacks (8.2ms)
INSERT INTO "ambassador_task_assignments"
("user_id","ambassador_task_id","created_at","updated_at")
VALUES
(1,29,'2019-06-07 01:15:01.629558','2019-06-07 01:15:01.629558')
(100,29,'2019-06-07 01:15:01.629558','2019-06-07 01:15:01.629558')
...

RETURNING "id"

#<Struct:ActiveRecord::Import::Result:0x7fa8c1a99bf8
  failed_instances = [],
  ids = ["958", "959", "960", "961", "962", "963", "964", "965", "966"...],
  num_inserts = 1,
  results = []>

```

Resolves #839

### DIFF
--- a/app/services/ambassador_task_assignment_creator.rb
+++ b/app/services/ambassador_task_assignment_creator.rb
@@ -14,17 +14,15 @@ module AmbassadorTaskAssignmentCreator
         .where(ambassador_task_assignments: { ambassador_task_id: ambassador_task.id })
         .select(:id)
 
-    not_assigned_ambassador_ids =
+    new_assignments =
       Ambassador
         .includes(:ambassador_task_assignments)
-        .where.not(id: already_assigned_ambassador_ids)
+        .where
+        .not(id: already_assigned_ambassador_ids)
         .pluck(:id)
+        .map { |ambassador_id| { user_id: ambassador_id, ambassador_task_id: ambassador_task.id } }
 
-    new_assignments =
-      not_assigned_ambassador_ids
-        .map { |ambassador_id| { user_id: ambassador_id, ambassador_task: ambassador_task } }
-
-    AmbassadorTaskAssignment.create(new_assignments)
+    AmbassadorTaskAssignment.import(new_assignments, validate: true)
   end
 
   def self.assign_all_ambassador_tasks_to(ambassador_or_id)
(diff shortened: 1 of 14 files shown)


app/services/ambassador_task_assignment_creator.rb
```diff
@@ -0,0 +1,43 @@
+module AmbassadorTaskAssignmentCreator
```
sethherr: Why create this module instead of putting these directly into the worker?

app/workers/ambassador_membership_after_create_worker.rb
```diff
@@ -0,0 +1,12 @@
+class AmbassadorMembershipAfterCreateWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: "high_priority", backtrace: true
+
+  def perform(membership_id)
+    membership = Membership.includes(:organization).find(membership_id)
+    return unless membership.ambassador?
```
sethherr: Doesn’t the service check this with `Ambassador.exists?`

app/services/ambassador_task_assignment_creator.rb
jmromer: It simplifies testing and REPL-driven development somewhat to be able to execute the service methods separately, but mainly because SRP. Keeping logic out of workers (especially when the workers are life-cycle based as these are) adds flexibility. For example, if you want to combine different sets of business logic in the context of a job, that's a simpler addition when the guts of the logic are elsewhere.

app/workers/ambassador_membership_after_create_worker.rb
jmromer: Yeah, that seems a little paranoid. Removed it in 6195707a6b1

PR #863, opened 2019-06-07
Minor view improvements

sethherr:


app/views/layouts/new_admin.html.haml
```diff
@@ -43,7 +43,7 @@
                 other_links = [
                                 { title: "POS Integration", path: "https://posintegration.bikeindex.org", match_controller: false },
                                 { title: "Ambassador Tasks", path: admin_ambassador_tasks_path, match_controller: true },
-                                { title: "Ambassador Task Assignments", path: admin_ambassador_task_assignments_path, match_controller: true },
+                                { title: "Completed Ambassador Tasks", path: admin_ambassador_task_assignments_path, match_controller: true },
```
sethherr: How do you feel about this @jmromer? I think this naming makes it easier for people who aren't familiar with the code.

jmromer: Makes sense. I went back and forth between the ambassador-facing terminology and the internal naming. `Completed Ambassador Activities` might be good too, for consistency with the former elsewhere.

sethherr: Makes sense, good idea.

I did a quick rename of places I found "Tasks" in admin

PR #864, opened 2019-06-07
Add previous slug for organizations

sethherr:
BikeEastBay changed their name from East Bay Bike Coalition. There are still some links and references around on their site to the previous slug they had (`ebbc`). I suspect we will run into this problem again in the future, so add a way to deal with it.


sethherr:
1. Pretty unlikely. I don't intend to support or encourage this path. I thought of creating an association, but this is way easier and also more performant. The time cost of doing a data migration for a single string column will be lower than the cost of implementing something we don't use.
2. Meh, only developers can see this, I don't think there is a high likelihood of it breaking. If it gets more use, we'll figure it out then.

PR #866, opened 2019-06-08
Optimize test suite, CI config

jmromer:
Optimizes the test suite and CI config for performance

### Local improvements:

Speeds up local test runs by ~95% (given sufficient parallelization — 7:14 to 0:21 on my machine):

- Adds `parallel_tests` to parallelize the test suite (best results found at 1 process per core, details below)
- Adds `spring` to pre-load the application. Also speeds up rake, non-parallel test runs, etc
- Changes database cleaner strategy to default to transactions ([granularly overridable](https://github.com/bikeindex/bike_index/blob/538a1d924e43e18556069580c1cedc34c852aeb7/spec/support/database_cleaner.rb)) [more on [db cleaning strategies](https://stackoverflow.com/questions/11419536/postgresql-truncation-speed/11423886#11423886)]

### CI config changes

- Reduces parallelism to [2 nodes](https://github.com/bikeindex/bike_index/pull/866/commits/538a1d924e43e18556069580c1cedc34c852aeb7), which improves throughput across multiple builds (see below)
- Caches [more granularly](https://github.com/bikeindex/bike_index/pull/866/commits/a89c345ffb4cc57d7c713c4db064196baf19e91a) (separate caches for Ruby and Node dependencies, and for precompiled assets)

### RSpec updates 

In line with upstream deprecations:

- Refactors to remove global render_views setting
- Disables monkey patching
- Disables inferring test type from directory
- Use "rails_helper" instead of "spec_helper" (the latter should be used for any non-rails code)

### Usage

To run the RSpec suite in parallel, set up databases with 

```
bin/rake parallel:setup
```

and run with 

```
bin/rake parallel:spec
```

You can override the default number of processes with the following env vars:

```
export PARALLEL_TEST_PROCESSORS=XX   
```

Set [`export DISABLE_SPRING=0`](https://github.com/grosser/parallel_tests/wiki/Spring) to speed up app boot time when running tests in parallel.

-----------------------------------------------

### Before

```
Finished in 7 minutes 4 seconds (files took 9.43 seconds to load)
1963 examples, 0 failures, 13 pending
Randomized with seed 27223

bundle exec rspec  93.85s user 18.10s system 25% cpu 7:14.45 total
```

[CI: 6:45](https://circleci.com/gh/bikeindex/bike_index/2959)

### After

spring + parallel_tests

One process per processor: `export PARALLEL_TEST_PROCESSORS=18`
Spring enabled: `export DISABLE_SPRING=0`

#### with truncation

Parallelized:

```
Using recorded test runtime
36 processes for 241 specs, ~ 6 specs per process

Running via Spring preloader in process 33075
Running via Spring preloader in process 33076
...

1963 examples, 0 failures, 13 pendings
Took 152 seconds (2:32)

bin/rake parallel:spec  18.42s user 11.79s system 18% cpu 2:43.33 total
```

#### with :deletion

Parallelized:

```
1963 examples, 0 failures, 13 pendings
Took 28 seconds

bin/rake parallel:spec  12.88s user 6.53s system 51% cpu 37.505 total
```

[CI: 6:00](https://circleci.com/gh/bikeindex/bike_index/2961)

#### with :transaction by default

In serial:

```
Finished in 1 minute 45.18 seconds (files took 2.11 seconds to load)
1963 examples, 0 failures, 13 pending

Randomized with seed 64163

bin/rspec --format=progress  0.23s user 0.15s system 0% cpu 1:47.94 total
```

[CI 4:15](https://circleci.com/gh/bikeindex/bike_index/3024)

Parallelized:

```
1964 examples, 0 failures, 13 pendings
Took 22 seconds

bin/rake parallel:spec  13.03s user 6.69s system 62% cpu 31.579 total
```

8c179ba adds garbage collection deferral, which can shave off a few more seconds:

```
1964 examples, 0 failures, 13 pendings

Took 17 seconds
bin/rake parallel:spec  0.23s user 0.15s system 2% cpu 18.953 total
```

#### Before 

Parallel: 5

parallel-5-done

#### After

Parallel: 2

parallel-2-done


bin/parallel_cucumber
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
sethherr: Do we need this? I don't think we have an cucumber

bin/parallel_spinach
```diff
@@ -0,0 +1,29 @@
+#!/usr/bin/env ruby
```
sethherr: Or spinach

spec/support/sidekiq.rb
```diff
@@ -0,0 +1,5 @@
+require "sidekiq/testing"
```
sethherr: Is there a strong reason to move this to a separate file rather than rails_helper? Beyond my desire to consolidate, I've generally seen these included in the rails_helper file.

Also goes for simplecov and vcr.

spec/workers/additional_email_confirmation_worker_spec.rb
```diff
@@ -1,6 +1,6 @@
-require "spec_helper"
+require "rails_helper"
 
-describe AdditionalEmailConfirmationWorker do
+RSpec.describe AdditionalEmailConfirmationWorker, type: :job do
```
sethherr: Just [looked up job specs](https://relishapp.com/rspec/rspec-rails/docs/job-specs/job-spec), because I was concerned it wasn't applicable since we aren't using ActiveJob.

It seems like the `type: job` _is_ targeted at ActiveJob, and adds additional functionality for it - but since this isn't causing any problems, I don't see a reason not to use it.

Ignore this comment 🙂

sethherr:
Do you think it would be reasonable to add some default test parallelization settings to `.env`? It seems like they have a huge benefit, so making it default to them seems reasonable.

jmromer:
> Do you think it would be reasonable to add some default test parallelization settings to .env?

They won't be picked up by the shell if placed there. Luckily the env var is optional and the library's default behavior (2 processes per core) yields virtually optimal performance, and I suspect that value scales better than 1 per core as the test suite grows. I added some documentation in 10ee7ed6

> I have found render_views to be incredibly valuable, since it catches bugs in the views - and makes the controller specs closer to the integration specs I believe are the most useful.

We still render views in controller specs — that's enabled per-controller-spec-suite, in line with the library's updated default. Request specs render views.

bin/parallel_cucumber
jmromer: Oversight, thanks! Updated in https://github.com/bikeindex/bike_index/pull/866/commits/af3c4acce13167d632d03c9c842a4091522391ab https://github.com/bikeindex/bike_index/pull/866/commits/ec2fe3c03a39dc83eacf1841f3a16517486fad0b and https://github.com/bikeindex/bike_index/pull/866/commits/896be8f95d375d018ab51356db4ec7de040d95f4

PR #870, opened 2019-06-10
Add easy path for disabling spring locally

sethherr:
I know people love spring!

I've spent a non-trivial chunk of time dealing with various Spring issues (previously in other projects), which makes me cautious. Spring also doesn't seem to speed up my personal development process of guard running tests for the files I'm changing (by much).

This PR adds a rake task to toggle spring disabled. I decided to do the same thing Rails 5 does for toggling caching in development, so I also added a rake task for `dev:cache` here as well, since it's nice to have, and why wait till rails 5!

### DIFF
--- a/bin/spring
+++ b/bin/spring
@@ -3,15 +3,17 @@
 # This file loads spring without using Bundler, in order to be fast.
 # It gets overwritten when you run the `spring binstub` command.
 
+ENV["DISABLE_SPRING"] = "1" if File.exists?(File.join(__dir__, "../.offspring"))
+
 unless defined?(Spring)
-  require 'rubygems'
-  require 'bundler'
+  require "rubygems"
+  require "bundler"
 
   lockfile = Bundler::LockfileParser.new(Bundler.default_lockfile.read)
   spring = lockfile.specs.detect { |spec| spec.name == "spring" }
   if spring
     Gem.use_paths Gem.dir, Bundler.bundle_path.to_s, *Gem.path
-    gem 'spring', spring.version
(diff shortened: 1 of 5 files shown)


jmromer:
Sweet thanks! I'll check this out. I'm definitely seeing preloading help with parallel tests, but it's only appreciable when running the full suite.

I'm definitely also reminded of why I learned to hate Spring the last time I worked in Rails: https://github.com/rails/spring/issues/396

My workaround so far has been to run the server (`bin/spring server`) in the foreground in one tmux pane, monitor its status in another (`watch -n 2 'bin/spring status'`), and force-kill any lingering processes when ending my session in a tmuxinator hook: 

```shell
ps aux | grep -E 'spring .+ bike_index' | grep -v 'grep' | awk '{ print $2 }' | xargs kill -9 2>/dev/null
```

Screen Shot 2019-06-10 at 8 25 57 PM

bin/spring
```diff
@@ -3,15 +3,18 @@
 # This file loads spring without using Bundler, in order to be fast.
 # It gets overwritten when you run the `spring binstub` command.
 
+# Run rake dev:spring to toggle spring
+ENV["DISABLE_SPRING"] = "1" if File.exists?(File.join(__dir__, "..", "tmp", "spring-off.txt"))
+
```
jmromer: Wouldn't setting the DISABLE_SPRING ENV var be enough to disable it? 🤔

sethherr: But I want to be able to enter consoles without having to set than environmental variable, and run migrations, etc.

PR #868, opened 2019-06-10
Add automatically generated unstolen notification messages

jmromer:
Enables unstolen notifications for ambassador orgs, pre-populates the message based on the current user:

Screen Shot 2019-06-10 at 5 17 42 PM

```haml
-# app/views/bikes/_organized_access_panel.html.haml L140-151 (b9998cf1)

.collapse#new_stolen_message{ class: ("in" if @contact_owner_open) }
  - @stolen_notification ||= StolenNotification.new(bike: @bike, sender: current_user)
  = form_for @stolen_notification do |f|
    = f.hidden_field :bike_id, value: @bike.id
    = f.text_field :reference_url, placeholder: 'Link url', class: "form-control additional-field"
    - if current_user.ambassador?
      = f.text_area :message, required: true, value: @stolen_notification.default_message, rows: 6, class: "form-control"
    - else
      = f.text_area :message, required: true, placeholder: "Where did you see this #{@bike.type}? If you're contacting about an ad you found online, include a link in the field below", rows: 6, class: "form-control"

    .send-message
      = f.submit "Send message", class: 'btn btn-primary btn-lg'
```

Grants unstolen notifications permission to ambassador orgs:

```rb
# app/models/organization.rb L160-171 (2495c0ae)

def paid_for?(feature_name)
  features =
    Array(feature_name)
      .map { |name| name.strip.downcase.gsub(/\s/, "_") }

  return false unless features.present? && paid_feature_slugs.is_a?(Array)

  features.all? do |feature|
    paid_feature_slugs.include?(feature) ||
    (ambassador? && feature == "unstolen_notifications")
  end
end
```

Screen Shot 2019-06-10 at 3 43 14 PM

Resolves https://github.com/bikeindex/bike_index/issues/800

Related:

https://github.com/bikeindex/bike_index/pull/857
https://github.com/bikeindex/bike_index/pull/582

PR #871, opened 2019-06-10
Make Message kinds variable plural

sethherr:
Follow up to #868 - specifically to [my message about plural variables](https://github.com/bikeindex/bike_index/pull/868/files#r292229758)

I thought, from a cursory look around the code, that we were only passing singular paid features to `paid_for?` - however, this `@kind` was actually an array some of the time.

This refactors it to _always_ be an array, which makes more sense to me.

### DIFF
--- a/app/controllers/organized/messages_controller.rb
+++ b/app/controllers/organized/messages_controller.rb
@@ -38,9 +38,10 @@ module Organized
     end
 
     def ensure_permitted_message_kind!
-      @kind = params[:organization_message].present? ? params[:organization_message][:kind_slug] : params[:kind]
-      @kind ||= current_organization.message_kinds
-      return true if current_organization.paid_for?(@kind)
+      kind = params[:organization_message].present? ? params[:organization_message][:kind_slug] : params[:kind]
+      @kinds =  Array(kind).flatten if kind.present?
+      @kinds ||= current_organization.message_kinds
+      return true if current_organization.paid_for?(@kinds)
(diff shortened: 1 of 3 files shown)


app/controllers/organized/messages_controller.rb
```diff
@@ -38,9 +38,10 @@ def organization_messages
     end
 
     def ensure_permitted_message_kind!
-      @kind = params[:organization_message].present? ? params[:organization_message][:kind_slug] : params[:kind]
-      @kind ||= current_organization.message_kinds
-      return true if current_organization.paid_for?(@kind)
+      kind = params[:organization_message].present? ? params[:organization_message][:kind_slug] : params[:kind]
+      @kinds = Array(kind).flatten if kind.present?
```
jmromer: `Array()` has nice semantics -- if `kind` isn't present here it'll return an empty array:

```rb
params = {}
Array(params.dig(:organization_message, :kind_slug) || params[:kind])
# []

params = {organization_message: {kind_slug: "slug"} }
Array(params.dig(:organization_message, :kind_slug) || params[:kind])
# [ "slug"]

params = {kind: "kind"}
Array(params.dig(:organization_message, :kind_slug) || params[:kind])
# [ "kind" ]

params = {organization_message: {kind_slug: ["slug", "slug2"]} }
Array(params.dig(:organization_message, :kind_slug) || params[:kind])
# [ "slug", "slug2" ]
```

PR #872, opened 2019-06-11
Add uniqueness constraint on organization names

sethherr:
I think there might need to be more interface around this, eventually.

We ran into a problem where people from an organization had independently created two organizations and weren't able to figure out why they couldn't see each other. This should be harder to do than it is right now.


sethherr:
Looks kinda crappy, doesn't read perfectly - but hey, it's better than nothing and generally people shouldn't be seeing this.

Screen Shot 2019-06-10 at 6 41 35 PM

PR #875, opened 2019-06-11
Update simplecov, simplecov config

jmromer:
Test coverage dropped precipitously with #866 primarily because of the addition of Spring. I didn't realize its preloading would be used even when running tests with `bundle exec rspec`. 

A less significant factor (accounting for about 3pp of the coverage loss) was requiring simplecov later in the initialization process when migrating from using spec_helper to rails_helper.

In both cases, the issue was that SimpleCov needs to be loaded before anything else in the tests or app: 

> In order to track coverage for your whole application simplecov needs to be the first one so that it (and the underlying coverage library) can subsequently track loaded files and their usage.

This patch fixes the sequencing of `require`s, disables spring on CI and, for compatibility with Spring locally, [eager loads](https://github.com/colszowka/simplecov#want-to-use-spring-with-simplecov) the application.

```sh
DISABLE_SPRING=1 COVERAGE=1 bin/parallel_rspec spec 

# Coverage report generated to /coverage. 7862 / 8791 LOC (89.43%) covered.
# 1967 examples, 0 failures, 13 pendings
# Took 90 seconds (1:30)
```

~There may still be an issue with CodeClimate reporting because of the parallelization on CircleCI, though. (Are we paying for CC? The SimpleCov local UI is pretty decent...)~

Screen Shot 2019-06-11 at 9 38 29 AM


jmromer:
> do you think we should switch the default for the spring rake task to be disabled, rather than enabled?

I think that'd be a sensible default — if you're not familiar with spring and end up in a situation where it's causing the app or tests to hang, I can see that being really frustrating.

PR #876, opened 2019-06-11
Switch default to disabled for spring

sethherr:
From @jmromer in [#875](https://github.com/bikeindex/bike_index/pull/875#issuecomment-500925425)

> if you're not familiar with spring and end up in a situation where it's causing the app or tests to hang, I can see that being really frustrating.

### DIFF
--- a/lib/tasks/dev_configuration.rake
+++ b/lib/tasks/dev_configuration.rake
@@ -27,7 +27,7 @@ class RakeDevConfiguration
 
       if File.exist?(file)
         delete_cache_file(file)
-        puts "Spring is no longer enabled (you may need to manually kill the spring processes)"
+        puts "Spring is no longer enabled (you will need to manually kill the spring processes)"
       else
         create_cache_file(file)
         puts "Spring is now enabled."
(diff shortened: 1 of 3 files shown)


lib/tasks/dev_configuration.rake
```diff
@@ -22,15 +22,16 @@ class RakeDevConfiguration
...
 
     def toggle_spring
-      file = "tmp/spring-off.txt"
+      file = "tmp/spring-on.txt"
       FileUtils.mkdir_p("tmp")
 
       if File.exist?(file)
         delete_cache_file(file)
-        puts "Spring is now enabled."
+        puts "Spring is no longer enabled (you will need to manually kill the spring processes)"
```
jmromer: might be useful to go ahead and do that here with a system call

sethherr: That seems scary to me, but if it would be useful to you, I'll add it in!

jmromer: Nah, I'm ok waiting to see if the need arises :)

PR #880, opened 2019-06-12
Update bikes#show UI

jmromer:
Updates the bikes#show UI with premium listings selection behind a feature flag.


app/assets/javascripts/revised/pages/bikes/edit.coffee
```diff
@@ -17,6 +18,3 @@ class BikeIndex.BikesEdit extends BikeIndex
 
   submitBikeEditForm: ->
     $('form.primary-edit-bike-form').submit()
-    setTimeout (-> # Sometimes the page reload isn't triggered. Do it manually
-      location.reload(true)
-    ), 300
```
jmromer: This interferes with server-determined redirects -- I'm testing to see if it's safe to omit.

sethherr: Hmmmm. which server determined redirects?

app/controllers/bikes_controller.rb
```diff
@@ -153,51 +153,107 @@ def create
...
-    }
-    # To make stolen the first key if bike is stolen. using as_json for string keys instead of sym
-    (@bike.stolen ? hash.to_a.rotate(-1).to_h : hash).as_json
+      # When reporting a bike as stolen, redirect to the recovery form.
+      # The recovery form redirects to the stolen form and is handled by
+      # CoffeeScript in pages/bikes/edit_stolen.coffee
+      target_template = (params[:edit_template] == "report_stolen") ?
+        "report_recovered" : params[:edit_template]
+      redirect_to edit_bike_url(@bike, page: target_template) and return
+    end
```
jmromer: Currently the form simply reloads and is rerendered with the recovery form replacing the stolen form and vice versa. This reproduces that behavior by redirecting with the appropriate query param, but it may be more desirable from a UX standpoint to redirect to the base form.

app/assets/javascripts/bike_index.js.coffee
```diff
@@ -106,6 +106,7 @@ $(document).ready ->
   else if document.getElementById('admin-content')
     new BikeIndex.Views.DataTables
     if document.getElementById('bike_edit_root_url')
+      # TODO: Is this dead code?
```
sethherr: Yes! Good catch.

We also have a bunch of legacy backbone that is lying around. I think the admin bikes edit revised coffee is good to delete, I'm waiting to delete the backbone stuff until we've fully finished updating the admin.

app/assets/javascripts/revised/pages/bikes/edit.coffee
jmromer: On master:

```rb
# app/controllers/bikes_controller.rb L172-177 (721391b4)

  else
    flash[:success] = "Bike successfully updated!"
    return if return_to_if_present
    redirect_to edit_bike_url(@bike, page: params[:edit_template]) and return
  end
end
```


This branch:
```rb
# app/controllers/bikes_controller.rb L196-202 (ab3fa27b)

  # When reporting a bike as stolen, redirect to the recovery form.
  # The recovery form redirects to the stolen form and is handled by
  # CoffeeScript in pages/bikes/edit_stolen.coffee
  target_template = (params[:edit_template] == "report_stolen") ?
    "report_recovered" : params[:edit_template]
  redirect_to edit_bike_url(@bike, page: target_template) and return
end
```

app/controllers/bikes_controller.rb
sethherr: I think each template should be unique and uniquely named. So there is a "report stolen" form and then a "theft details" form.

sethherr: I'm not sure if I answered your question correctly? Let me know if I missed the point.

app/assets/javascripts/revised/pages/bikes/edit.coffee
sethherr: Try commenting it out with a note and we can test this once it's live and see if it's necessary?

app/controllers/bikes_controller.rb
```diff
@@ -153,51 +153,107 @@ def create
...
+    @bike_templates = bike_templates
+    @edit_templates = @theft_templates.merge(@bike_templates)
   end
 
   protected
 
+  # Determine the appropriate edit template to use.
+  def target_edit_template(requested_page, valid_pages)
+    default_page = (@bike.stolen? ? :theft_details : :bike_details)
+    return [true, default_page] if requested_page.blank?
```
sethherr: I would prefer not to use parallel assignment. Can you make this return either a hash or a single value?

app/assets/javascripts/revised/pages/bikes/edit.coffee
jmromer: It's necessary on this branch, to flip between the stolen and recovery forms and vice versa.

app/controllers/bikes_controller.rb
jmromer: > I think each template should be unique and uniquely named. 

That's the setup on this branch. The behavior I'm maintaining from master is that when you submit the "report stolen" form, you then see the "report recovery" form.

app/controllers/bikes_controller.rb
jmromer: Updated in 94e14ef

app/controllers/bikes_controller.rb
jmromer: >  the "report stolen" form simply reloads and is rerendered with the recovery form replacing the stolen form and vice versa. 

I expressed that poorly but hopefully the GIF disambiguates -- I meant to say that when the user marks a bike as stolen, the next thing they see is the "mark as recovered" form.

sethherr: Ahhh. Yes, I understand now. 

I think the behavior you suggested, of redirecting to to the base form, would be an improvement. Your choice of when/whether to make that update.

jmromer: Cool, that simplifies the implementation a bit too -- updated in f7c5fb4

PR #886, opened 2019-06-13
Update guard, rufo, Guardfile, CodeClimate config

jmromer:
- Adds a CodeClimate config to reduce the number of spurious / neurotic warnings
- Updates guard, rufo, rb-fsevent
- Moves eslint to CC 
- Disables rubocop-performance for now since CC doesn't seem to support it
- Updates Guardfile to test an inheritance tree when base classes are modified

@sethherr I should have asked about this earlier but if the Guardfile change interferes with your workflow, let me know — i'll make it a `.sample` file and have git ignore the actual one.

Screen Shot 2019-06-12 at 8 46 57 PM


sethherr:
Yeah, Guardfile change interferes with my process.

jmromer:
> whenever I save, it reruns the whole spec file

My bad, that's probably me throwing spaghetti at the wall—I've been trying to get it to pick up changes more quickly. Restored the previous state and added a `.sample` file in 9e8aafc

sethherr:
Two issues this PR introduced:

1. It seems like `bin/setup` includes a good process for setting up a new development environment. If that is the case, it should be document in the "Running Bike Index locally" section of the readme
2. The old Guardfile is removed from this repository

I think we should have both Guardfiles (the old one and the new one, from sample) in this repo. I'm happy if the process I use requires invoking guard with a manual call for a file. However, I think any updates made to either should be committed to this repository. I think people maintaining separate local Guardfiles is a bad pattern. I have opened a PR that defaults to the new and renames my old one Guardfile_legacy - #888

PR #887, opened 2019-06-13
Use beginning of day for week_creation_chart count

sethherr:


app/models/counts.rb
```diff
@@ -45,7 +49,8 @@ def assign_recoveries
     end
 
     def assign_week_creation_chart
-      assign_for("week_creation_chart", Bike.unscoped.where(created_at: (Time.now - 7.days)..Time.now).group_by_day(:created_at).count.to_json)
+      Time.zone = TimeParser::DEFAULT_TIMEZONE # Just to make sure
+      assign_for("week_creation_chart", Bike.unscoped.where(created_at: beginning_of_week..Time.now).group_by_day(:created_at).count.to_json)
```
jmromer: Defaulting to `Time.current` obviates the need for this

sethherr: hahahaha. You got me!

sethherr: Updated in c5f09030b44970e762149d07e74585a5b6ea7bde

jmromer: There's still a `Time.now` on line 52. I'm 99.9% sure it doesn't make a difference given (1) the TZ we set on the server, (2) the TZ set in the Rails app, and (3) the figures being computed here, but in general it takes some cognitive load off to not have to reason through it every time :)

sethherr: Yeah yeah yeah, #892

PR #888, opened 2019-06-13
Maintain legacy Guardfile

sethherr:
Follow up from #886 

This uses the process used by the old Guardfile_frontend, which had a process for running it documented in the readme (removed in [remove live reload](https://github.com/bikeindex/bike_index/commit/55af21cb4be60b3e967237278b91247b90996bb7)):

> If you're working on frontend development, you can use [Guard::LiveReload](https://github.com/guard/guard-livereload) to reload the frontend as you work with `bundle exec guard -G Guardfile_frontend`

Basically, we're maintaining both Guardfiles in the repository.


sethherr:
Running with non-parallel on legacy, in bikes_controller_spec.rb with a fail thrown in one test:

```ruby
# First test run:
Finished in 11.1 seconds (files took 7 seconds to load)
82 examples, 1 failure, 1 pending

# Re-saving controller_spec, rerun
Finished in 0.31851 seconds (files took 6.99 seconds to load)
1 example, 1 failure
```

Parallel run: 

```ruby
# First test run
Finished in 14.13 seconds (files took 7.64 seconds to load)
82 examples, 1 failure, 1 pending

# Re-saving controller_spec, rerun
Finished in 11.76 seconds (files took 7.28 seconds to load)
82 examples, 1 failure, 1 pending
```

If we're going to go the parallel testing route, that's fine. But we need to do a better job setting up the defaults - I'm still getting `1 processes for 1 specs, ~ 1 specs per process` - which means that this is not a net improvement without configuration. I would prefer to have it use the legacy, non-parallel setup by default, for now, and require running guard with a specified Guardfile for those using parallelization.

sethherr:
If we are switching to parallelization we need to:

1. default parallel count 2 (or greater)
2. override `rake db:test:prepare` to do all the parallel stuff
3. fix the rerun issue that causes `All examples were filtered out; ignoring {:focus=>true}`

jmromer:
I'm a little confused by this — the point of not versioning the Guardfile is so that each can vary independently. I'm cool with changing the defaults in the sample file though.

jmromer:
> I'm still getting `1 processes for 1 specs, ~ 1 specs per process`

From the looks of it, the `bin/parallel_rspec` runner doesn't accept specific line numbers, so it's running for the whole spec file. With two files:

```
bike_index premium-listings % bin/parallel_rspec spec/controllers/news_controller_spec.rb spec/controllers/bike_codes_controller_spec.rb

2 processes for 2 specs, ~ 1 specs per process
```

I would prefer to have it use the legacy, non-parallel setup by default, for now, and require running guard with a specified Guardfile for those using parallelization.

Why not just modify the `Guardfile` in the repo now? 

(Makes sense to update `Guardfile.sample` to more sensible defaults though)

jmromer:
> fix the rerun issue that causes All examples were filtered out; ignoring {:focus=>true}

That's caused by this setting in `spec_helper.rb`:

```rb
# spec/spec_helper.rb L28-33 (db52a54b)

# These two settings work together to allow you to limit a spec run
# to individual examples or groups you care about by tagging them with
# `:focus` metadata. When nothing is tagged with `:focus`, all examples
# get run.
config.filter_run :focus
config.run_all_when_everything_filtered = true
```
I don't think there's an impact but I don't mind disabling them. Less noise in the output at least.

With:

```
% bin/parallel_rspec spec/controllers/news_controller_spec.rb
1 processes for 1 specs, ~ 1 specs per process
Run options: include {:focus=>true}
All examples were filtered out; ignoring {:focus=>true}
Randomized with seed 61809
```

Without

```
% bin/parallel_rspec spec/controllers/news_controller_spec.rb
Using recorded test runtime
1 processes for 1 specs, ~ 1 specs per process
Randomized with seed 61809
```

sethherr:
> I'm a little confused by this — the point of not versioning the Guardfile is so that each can vary independently. I'm cool with changing the defaults in the sample file though.

Is there a reason your Guardfile would be different than everyone else's who is using parallelization? It seems like you're just doing the things that make sense if you're running parallel.

jmromer:
> Is there a reason your Guardfile would be different than everyone else's who is using parallelization? It seems like you're just doing the things that make sense if you're running parallel.

Suppose someone (future one-of-us, e.g.) wants to customize further (e.g., wants to use other plugins, or not work in "red-green-refactor" mode?) 🤔 

But:

I think I misread "process" in the following (I thought it referred to the `bundle exec` vs `bin/` invocation, rather than using the `-G` flag to specify a guardfile):

> This uses the process used by the old Guardfile_frontend, which had a process for running it documented in the readme (removed in remove live reload):

If we want to maintain a `Guardfile`, the least-effort approach seems to be rename `Guardfile.sample` to `Guardfile` and add something like `Guardfile.local` to .gitignore — that might be more intuitive for new developers when setting up the project too (I didn't realize we weren't using `bin/setup`)

sethherr:
My specific response to this is motivated by a concern about there being workflow optimizations which people can't see and don't benefit from.

I'd prefer to have your workflow committed to the repository, because I suspect at some point I'm going to switch over to paralleling and I'd like to benefit from any optimization work you do 🙂 

I'll simplify the readme to reference `bin/setup`, seems like it's a [good practice](https://thoughtbot.com/blog/bin-setup)

also - it looks like `bundle exec` and `bin` invoke the same thing, but `bin` is the current approved way of doing things (since 2013 🙄), so I will stop saying bundle install.

PR #893, opened 2019-06-14
Improve parallel testing instructions

sethherr:


README.markdown
```diff
@@ -81,38 +76,14 @@ We use [RSpec](https://github.com/rspec/rspec) and
...
 
-```shell-script
-bin/rake parallel:prepare
-```
+- Run `bin/rake parallel:prepare` to synchronize the test db schema after migrations (rather than `db:test:prepare`).
+- Run specific files or test directories with `bin/parallel_rspec <FILES_OR_FOLDERS>`
+- Run Guard with parallelism `bin/guard -G Guardfile_parallel`
 
-See the [`parallel_tests`][] docs for more.
+By default, `parallel_tests` spawns one process per CPU in your computer (which is a sensible default). You can optionally set the
```
sethherr: Do you use a different number @jmromer ? If not, can we just remove this sentence?

sethherr:
I reset my database and ran `bin/setup` and then `bin/rake parallel:spec` and `bin/rake parallel:spec` and everything seemed to work correctly.

Will this just work now? Are there other configuration options that need to be described? Or is this batteries included enough? If there are additions to the Guardile or anything else you do @jmromer, please add them!

README.markdown
jmromer: I think that's fine to delete — pointing to the canonical docs should be enough.

bin/setup
```diff
@@ -15,10 +15,13 @@ Dir.chdir APP_ROOT do
...
   system "redis-server &"
 
-  puts "\n== Preparing database =="
-  system "bin/rake db:setup db:test:prepare"
+  puts "\n== Preparing and seeding database =="
+  system "bin/rake db:setup"
   system "bin/rake seed_test_users_and_bikes"
 
+  puts "\n== Preparing test database (including parallelism) =="
+  system "bin/rake parallel:prepare"
```
jmromer: I think we'll want `bin/rake parallel:setup` for the first run. It can be time-consuming, though -- should we set up for parallelism eagerly?

> Create additional database(s)
> `rake parallel:create`
> Copy development schema (repeat after migrations)
> `rake parallel:prepare`
> Setup environment from scratch (create db and loads schema, useful for CI)
> `rake parallel:setup`
https://github.com/grosser/parallel_tests

sethherr: I haven’t successfully run `parallel:setup` - it errors. Is it still doing things that need to happen despite failing to finish?

jmromer: Sorry, typo — I meant `parallel:create`. I think `parallel:prepare` will fail if the additional test dbs don't exist yet. I'll test it though

jmromer: Disregard my last -- just repeated setup and it works fine 👍

PR #903, opened 2019-06-16
Fix new recovery display page(Minor Bug)

Robertgervais:


Robertgervais:
Hey, trying to create a new recovery display wasn't getting any info from stolen_record. Just needed to pass the right params. Fixxo.

PR #904, opened 2019-06-16
Fix overflow issues on mobile

sethherr:
Screen Shot 2019-06-17 at 11 32 18 AM


app/decorators/ambassador_task_assignment_decorator.rb
```diff
@@ -8,7 +8,7 @@ def button_to_toggle_completion_status(current_user, current_organization)
     return unless current_user.ambassador?
 
     is_complete = ambassador_task_assignment.completed?
-    button_label = is_complete ? "Mark Pending" : "Mark Complete"
+    button_label = is_complete ? "Pending" : "Complete"
```
jmromer: Just want to flag the potential for confusion / friction here—for an activity that's been completed, the button will say "Pending". (user: "Is pending the current state, or the target state?") 

We may also want to take this opportunity to move this logic into a Helper method since we're moving away from Draper.

sethherr: Good point, re: confusion. I’m going to revert that and see if I can just fix it via css. If I can’t, I’ll do something with spans or something and move the logic into a helper

PR #905, opened 2019-06-17
Add Character Counter to recovery display form

Robertgervais:


Robertgervais:
I'll put a shot up in a second. Also, you're right on all accounts, I think I just wanted to see if that JS was okay before I continued on. I think I'd like to style it as "x/max count" and maybe it'll change colors once it's hit the max? Something like that.

Robertgervais:
Image 6-17-19 at 3 40 PM

PR #906, opened 2019-06-17
Revise admin ownerships edit

Robertgervais:


Robertgervais:
Image 6-16-19 at 8 48 PM

Robertgervais:
Image 6-16-19 at 8 54 PM

app/controllers/admin/ownerships_controller.rb
```diff
@@ -2,6 +2,7 @@ class Admin::OwnershipsController < Admin::BaseController
   before_filter :find_ownership
 
   def edit
+    render layout: "new_admin"
```
sethherr: Is this controller wide?

Robertgervais: No but it should be, i'll fix.

Robertgervais:
#175

Issue #908, opened 2019-06-17
Fix ambassador leaderboard counts


sethherr:
![F14704FB-A189-48E6-B1F7-C504B9BF60D8](https://user-images.githubusercontent.com/1235441/59621894-17a34f80-90e5-11e9-990c-59b0c3a419f5.png)

jmromer:
@sethherr Looks like a data issue -- bulk insert seems to be creating duplicate assignments. 

Deduping with the following

```rb
keep_ids = AmbassadorTaskAssignment.group('user_id, ambassador_task_id').pluck('max(id)')

AmbassadorTaskAssignment.where.not(id: keep_ids).find_in_batches(batch_size: 100) {|id_batch| AmbassadorTaskAssignment.where(id: id_batch).delete_all }
```

debugging the after_create jobs...

jmromer:
```rb
AmbassadorTaskAssignment.group('user_id, ambassador_task_id').pluck('count(id)').uniq
# => [1]
```

PR #909, opened 2019-06-17
Add db-level uniqueness constraint for task assignments

jmromer:
Adds a database-level uniqueness constraint to task assignments. Working on determining why dupes are being created, I'll add a fix for that in a follow-up PR.

Related: #908


sethherr:
You could remove activerecord import and go back to `AmbassadorTask.find_each { |task| task.assign_to(user) }`

jmromer:
@sethherr Yes, that's the plan. A uniqueness index would be good to add as well.

PR #910, opened 2019-06-17
Use find_each to insert task assignments

### LINKED ISSUE #908: Fix ambassador leaderboard counts

jmromer:
I'm still unclear on why these insertions aren't idempotent when run from Sidekiq jobs -- the queries look good, after re-checking them — but as an expedient fix this patch reverts to iterative insertion as a safeguard. Model-layer validation should prevent any dupes from being created.

Closes #908


sethherr:
Can we remove activerecord-import, since I don't believe we're using it anymore?

PR #915, opened 2019-06-19
Add Bike Index Alert purchasing back-end

jmromer:
- Adds customer-facing alerts listings (looked awkward up top, but happy to change it back if we definitely want to place it there, see screenshots)
- Adds purchasing logic for purchasing alerts
- Add charge creation endpoint

Screen Shot 2019-06-19 at 7 39 14 PM
Screen Shot 2019-06-19 at 7 39 34 PM

listing above:

Screen Shot 2019-06-19 at 12 47 36 PM

PR #916, opened 2019-06-19
Add timecop. flipper group, separated request spec helpers

jmromer:
- Adds [timecop](https://github.com/travisjeffery/timecop) gem for freezing time
- Adds a flipper superusers group
- Separates helpers for request and controller specs, and extracts json-request-related helpers 

Extracted from #915

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -164,6 +164,7 @@ group :test do
   gem "factory_bot_rails"
   gem "rspec-sidekiq"
   gem "simplecov", require: false
+  gem "timecop"
   gem "vcr" # Stub external HTTP requests
   gem "webmock" # mocking for VCR
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -569,6 +569,7 @@ GEM
     thor (0.19.1)
     thread_safe (0.3.6)
     tilt (2.0.8)
+    timecop (0.9.1)
     twitter (6.2.0)
       addressable (~> 2.3)
       buftok (~> 0.2.0)
@@ -738,6 +739,7 @@ DEPENDENCIES
   swagger-ui_rails
   therubyracer
   thor (= 0.19.1)
+  timecop
   twitter
   uglifier
   unicorn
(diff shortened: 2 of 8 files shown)


Gemfile
```diff
@@ -164,6 +164,7 @@ group :test do
   gem "factory_bot_rails"
   gem "rspec-sidekiq"
   gem "simplecov", require: false
+  gem "timecop"
```
sethherr: What are you adding timecop for? I'd prefer to avoid timecop, if at all possible, either by using `be_within(5.seconds).of xxxx` or [using the built in rails helpers](https://andycroll.com/ruby/replace-timecop-with-rails-time-helpers-in-rspec/) (full disclosure, I haven't tried the helpers).

That said, maybe there is a need for it, in which case, go for it.

jmromer: This is the test case that prompted it:

```rb
# spec/models/theft_alert_spec.rb L19-32 (3989a5d4)

it "sets begin and end times, flips the status" do
  theft_alert = FactoryBot.create(:theft_alert_paid)
  duration = theft_alert.theft_alert_plan.duration_days
  now = nil

  Timecop.freeze do
    now = Time.current
    theft_alert.begin!(facebook_post_url: "https://facebook.com")
  end

  expect(theft_alert.begin_at).to eq(now.beginning_of_day)
  expect(theft_alert.end_at).to eq(now.end_of_day + duration.days)
  expect(theft_alert.status).to eq("active")
end
```

sethherr: I don't understand why you need to freeze time for that?

Separately - I don't understand why you would do a start time of beginning of day instead of the  current time.

jmromer: Vestigial -- removed in https://github.com/bikeindex/bike_index/pull/918/commits/6ede7e8528bd971c7f3612896b9dd35dd6539c58

> Separately - I don't understand why you would do a start time of beginning of day instead of the current time.

If the time meaningful, let's not truncate it

PR #918, opened 2019-06-19
Add TheftAlert, TheftAlertPlan models

jmromer:
- Adds `TheftAlert` and `TheftAlertPlan` models (with a temporary seed method on the latter)
- Adds admin pages to manage these

Screen Shot 2019-06-19 at 12 24 38 PM
Screen Shot 2019-06-19 at 12 24 50 PM

Extracted from #915 to reduce diff size

### DIFF
--- a/app/controllers/admin/theft_alerts_controller.rb
+++ b/app/controllers/admin/theft_alerts_controller.rb
@@ -1,0 +1,48 @@
+class Admin::TheftAlertsController < Admin::BaseController
+  layout "new_admin"
+
+  def index
+    @theft_alerts = TheftAlert.includes(:theft_alert_plan).all
+  end
+
+  def edit
+    @theft_alert = TheftAlert.find(params[:id])
+
+    case state_transition
+    when "begin"
+      render :edit
+    else
+      flash[:error] = "Invalid state transition."
+      redirect_to admin_theft_alerts_path
+    end
+  end
+
+  def update
+    @theft_alert = TheftAlert.find(params[:id])
+    @theft_alert.public_send("#{state_transition}!", theft_alert_params)
+
+    if @theft_alert.errors.present?
+      flash[:error] = @theft_alert.errors.full_messages.first
+      redirect_to edit_admin_theft_alert_path(@theft_alert, params: { state_transition: state_transition })
+    else
+      flash[:success] = "Success!"
+      redirect_to admin_theft_alerts_path
+    end
+  end
+
+  private
+
+  def theft_alert_params
+    params.require(:theft_alert).permit(:facebook_post_url)
+  end
+
+  def state_transition
+    return @state_transition if defined?(@state_transition)
+
+    valid_state_transitions = %w[begin end reset]
+    state_transition = params[:state_transition]
+    return unless state_transition.in?(valid_state_transitions)
+
+    @state_transition = state_transition
+  end
+end
--- a/app/controllers/admin/theft_alerts_controller.rb
+++ b/app/controllers/admin/theft_alerts_controller.rb
@@ -19,7 +19,13 @@ class Admin::TheftAlertsController < Admin::BaseController
 
   def update
     @theft_alert = TheftAlert.find(params[:id])
-    @theft_alert.public_send("#{state_transition}!", theft_alert_params)
+
+    case state_transition
+    when "begin"
+      @theft_alert.begin!(facebook_post_url: theft_alert_params[:facebook_post_url])
+    when "end", "reset"
+      @theft_alert.public_send("#{state_transition}!")
+    end
 
     if @theft_alert.errors.present?
       flash[:error] = @theft_alert.errors.to_a
--- a/app/controllers/admin/theft_alerts_controller.rb
+++ b/app/controllers/admin/theft_alerts_controller.rb
@@ -22,7 +22,7 @@ class Admin::TheftAlertsController < Admin::BaseController
     @theft_alert.public_send("#{state_transition}!", theft_alert_params)
 
     if @theft_alert.errors.present?
-      flash[:error] = @theft_alert.errors.full_messages.first
+      flash[:error] = @theft_alert.errors.to_a
       redirect_to edit_admin_theft_alert_path(@theft_alert, params: { state_transition: state_transition })
     else
       flash[:success] = "Success!"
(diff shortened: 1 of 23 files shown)


app/controllers/admin/theft_alerts_controller.rb
```diff
@@ -0,0 +1,48 @@
...
+      redirect_to admin_theft_alerts_path
+    end
+  end
+
+  def update
+    @theft_alert = TheftAlert.find(params[:id])
+    @theft_alert.public_send("#{state_transition}!", theft_alert_params)
+
+    if @theft_alert.errors.present?
+      flash[:error] = @theft_alert.errors.full_messages.first
```
sethherr: Why not all the messages?

app/models/theft_alert.rb
```diff
@@ -0,0 +1,79 @@
...
+  belongs_to :creator,
+             class_name: "User",
+             foreign_key: :user_id
+
+  scope :due_for_expiration, -> { active.where('"theft_alerts"."end_at" <= ?', Time.current) }
+
+  delegate :bike, to: :stolen_record
+
+  def begin!(kwargs)
+    start_time = Time.current.beginning_of_day
```
sethherr: I don't understand why we'd do `beginning_of_day` here rather than the current time. What is the intention of that?

app/models/theft_alert.rb
```diff
@@ -0,0 +1,79 @@
...
+  belongs_to :payment
+  belongs_to :creator,
+             class_name: "User",
+             foreign_key: :user_id
+
+  scope :due_for_expiration, -> { active.where('"theft_alerts"."end_at" <= ?', Time.current) }
+
+  delegate :bike, to: :stolen_record
+
+  def begin!(kwargs)
```
sethherr: I'd prefer if the expected keys were included here. ie 

```ruby
def begin!(facebook_post_url: "")
```

app/models/theft_alert.rb
```diff
@@ -0,0 +1,79 @@
...
+      status: "active",
+      facebook_post_url: kwargs.fetch(:facebook_post_url, ""),
+      begin_at: start_time,
+      end_at: end_time.end_of_day,
+    }
+
+    update(attrs) if valid_state?(attrs)
+  end
+
+  def end!(*)
```
sethherr: I'd prefer explicit args here as well (ditto for `reset!`)

app/models/theft_alert.rb
```diff
@@ -0,0 +1,79 @@
...
+      end_at: nil,
+    }
+
+    update(attrs) if valid_state?(attrs)
+  end
+
+  private
+
+  # Ensure fields have expected values for the target state's status
+  def valid_state?(target_state)
```
sethherr: I think I'd prefer this to either:

Rather than passing a hash of target state, assign the attrs to the `TheftAlert` and use a [custom validator](https://guides.rubyonrails.org/active_record_validations.html#custom-methods) like `validate :record_in_valid_state` - which will block save if in a state is invalid. This has the advantage of following rails conventions, and enabling things like re-rendering a form with the values that were input, while displaying the error.

spec/factories/theft_alerts.rb
```diff
@@ -0,0 +1,56 @@
...
+    end
+
+    trait :begun do
+      status { "active" }
+      begin_at { Time.current.beginning_of_day }
+      sequence(:facebook_post_url) do |n|
+        "https://facebook.com/user.#{creator.id}/posts/#{n}"
+      end
+      before(:create) do |theft_alert, _evaluator|
+        plan = create(:theft_alert_plan)
```
sethherr: I'd prefer if, for both this an `ended`, you used a plan that would be assignable during factory creation (so that you can create multiple theft alerts referencing the same plan, etc)

I also don't believe you need to put this in a `before(:create)` block, I think you can just do:

```ruby
end_at { begin_at + theft_alert_plan.duration_days.days) } # optionally adding .end_of_day if necessary
```

app/models/theft_alert.rb
jmromer: I'd prefer the same, but I ran into some friction with that approach. I'll circle back to trace through the execution order and refactor.

sethherr: If you want to deal with refactoring it later that's fine.

spec/factories/theft_alerts.rb
jmromer: > I'd prefer if, for both this an ended, you used a plan that would be assignable during factory creation (so that you can create multiple theft alerts referencing the same plan, etc)

That's already the case -- it's a belongs_to

> I also don't believe you need to put this in a before(:create) block, I think you can just do:

That's not the case (the plan must be set for the record to be persisted), but if there's a workaround, let me know.

app/models/theft_alert.rb
```diff
@@ -0,0 +1,76 @@
...
+            presence: true
+
+  belongs_to :stolen_record
+  belongs_to :theft_alert_plan
+  belongs_to :payment
+  belongs_to :creator,
+             class_name: "User",
+             foreign_key: :user_id
+
+  scope :due_for_expiration, -> { active.where('"theft_alerts"."end_at" <= ?', Time.current) }
```
sethherr: nit: I think you can just do `active.where("end_at <= ?", Time.current)`

jmromer: You can, but I usually default to explicitness in scopes to aid composability — this sort of exception can be a hassle to debug when it's deep inside a query:

```rb
Bike.joins(:manufacturer).where('created_at < ?', Time.current)

# #<ActiveRecord::StatementInvalid: PG::AmbiguousColumn: ERROR:  column reference "created_at" is ambiguous

Bike.joins(:manufacturer).where('"bikes"."created_at" < ?', Time.current)

# SELECT COUNT(*) FROM "bikes" INNER JOIN "manufacturers" ON "manufacturers"."id" = "bikes"."manufacturer_id" WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND ("bikes"."created_at" < '2019-06-19 23:55:12.317688')  [["example", "f"], ["hidden", "f"]]
#  Bike Load (736.8ms)  SELECT "bikes".* FROM "bikes" INNER JOIN "manufacturers" ON "manufacturers"."id" = "bikes"."
```

PR #921, opened 2019-06-20
Add theft alert purchasing background jobs

jmromer:
TODO

- [ ] Send an email to admins about a theft alert being purchased
- [ ] Sidekiq job to deactivate any alerts that have passed


app/controllers/theft_alerts_controller.rb
```diff
@@ -0,0 +1,42 @@
+class TheftAlertsController < ApplicationController
+  before_action :ensure_user_allowed_to_create_theft_alert
+
+  def create
```
sethherr: I know this might be early - but something I'm seeing here that I would love if there was a way to avoid is the creation of both the alert and the purchase in the same controller action.

I would love to create the theft alerts by the time they've gone to the purchase page. Two reasons for this:

1. It reduces the complexity of the controller action. eg it wouldn't be necessary to rescue multiple types of errors (`ActiveRecord::RecordInvalid, ActiveRecord::RecordNotFound => e`)
2. We get additional visibility into the funnel

app/models/stolen_record.rb
```diff
@@ -28,6 +28,7 @@ def self.old_attr_accessible
   belongs_to :state
   belongs_to :creation_organization, class_name: "Organization"
   belongs_to :recovering_user, class_name: "User"
+  has_many :theft_alerts, dependent: :destroy
```
sethherr: I would prefer not to `dependent destroy` this, because it's tied to payments (even though it will leave orphan records)

app/controllers/theft_alerts_controller.rb
sethherr: sorry, I realize I should've said this on the other PR. I just didn't catch it

jmromer: > so it's already created by the time they've gone to the purchase page

Can you elaborate? I'm interpreting "purchase page" as the `bikes#edit?alert` template, which is creating some ambiguity for me—this sounds either like we want to create the TheftAlert before the user chooses a TheftAlertPlan or like we want to add our own payment UI to separate payment info submission from plan selection.

> a way to avoid is the creation of both the alert and the purchase in the same controller action.

The current implementation persists the theft alert even if the Stripe payment fails—we could push farther in that direction by moving the payment processing into a background job, if that addresses the concern.

jmromer: > sorry, I realize I should've said this on the other PR. I just didn't catch it

No prob! I think working in multiple passes is usually better / lower-risk

PR #924, opened 2019-06-20
Add alert purchasing background jobs, update plan selection UI

jmromer:
- [x] Send an email to admins about a theft alert being purchased
- [x] Sidekiq job to deactivate any alerts that have passed 
- [x] Update UI

Continued from #921, which was accidentally closed automatically

demo

### DIFF
--- a/app/controllers/theft_alerts_controller.rb
+++ b/app/controllers/theft_alerts_controller.rb
@@ -20,15 +20,15 @@ class TheftAlertsController < ApplicationController
   rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotFound => e
     Honeybadger.notify(e)
     flash[:error] = "We were unable to process your order. Please contact support."
+    redirect_to edit_bike_url(@bike, params: { page: :alert_purchase })
   rescue Stripe::StripeError => e
     Honeybadger.notify(e)
-    flash[:error] = "Your order is pending, but we were unable to complete payment. Please contact support."
     TheftAlertPurchaseNotificationWorker.perform_async(theft_alert.id)
+    flash[:error] = "Your order is pending, but we were unable to complete payment. Please contact support to complete your purchase."
+    redirect_to edit_bike_url(@bike, params: { page: :alert_purchase })
   else
-    flash[:success] = "Success! Your order is pending."
     TheftAlertPurchaseNotificationWorker.perform_async(theft_alert.id)
-  ensure
-    redirect_to edit_bike_url(@bike, params: { page: :alert })
+    redirect_to edit_bike_url(@bike, params: { page: :alert_purchase_confirmation })
   end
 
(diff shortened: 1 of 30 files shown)


sethherr:
re @jmromer comment https://github.com/bikeindex/bike_index/pull/921#discussion_r295942429

> Can you elaborate? I'm interpreting "purchase page" as the bikes#edit?alert template, which is creating some ambiguity for me—this sounds either like we want to create the TheftAlert before the user chooses a TheftAlertPlan or like we want to add our own payment UI to separate payment info submission from plan selection.

I'm seeing this as, you click on a theft alert and you get sent to a new page, which displays your selected alert and opens the stripe payment form on page load. Here is a rough implementation concept for this

- The theft alert page has a "chosen" state for an alert and a non-chosen state. This happens if a user has an active alert that is not-purchased.
- If there is a chosen plan, the other plans are reduced opacity. The chosen plan's button text is "Purchase alert" - and it opens the stripe form
- The other, non-chosen plans still have "Choose plan" buttons - and clicking on them updates/erases any other non-purchased alerts

app/controllers/theft_alerts_controller.rb
```diff
@@ -20,13 +20,15 @@ def create
   rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotFound => e
```
sethherr: Is it possible to split these different actions into different controller actions?

jmromer: If by actions you mean (1) creating the theft alert and (2) creating a payment: We can't redirect a POST request, but I'm open to hearing suggestions. Payment creation could and definitely should go in a background job.

sethherr: That is what I mean by 2 actions. I imagine, with the current flow, that you would create (or update an existing inactive) theft alert before rendering the `alert_purchase` template.

w/r/t background processing of payments - just running through this with stripes test cards for failures, it seems like stripe manages all the payment verification before actually sending us a response (e.g. if the card is declined, it shows it in the form before reloading the page). So I'm not sure what backgrounding the payment creation would get us?

sethherr: After seeing how smooth stripe's process is for handling problems with payments now, I'm not concerned about splitting this into multiple different actions.

This little article I think covered my opinions about using the bang method for save (though I don't think the bang method should be removed, there are many uses for it) - https://thoughtbot.com/blog/save-bang-your-head-active-record-will-drive-you-mad - I think the `save!` in the theft controller is because theft alerts aren't using a traditional active record validation?

I still think it would be great to get the theft alert created before making the payment, in a separate action.

But **this isn't high priority**, I assume we're going to extend this process with additional steps for writing text in the alert, etc, and we can do the refactoring then.

PR #930, opened 2019-06-21
Require accepting vendor terms for organization view

sethherr:
Also:

- Switch to using contexts rather than methods for the request spec `logged_in_as...`
- Make redirects to `user_root_url` more consistent

### DIFF
--- a/spec/support/request_spec_helpers.rb
+++ b/spec/support/request_spec_helpers.rb
@@ -18,19 +18,19 @@ module RequestSpecHelpers
 
   RSpec.shared_context :request_spec_logged_in_as_organization_admin do
     let(:current_organization) { FactoryBot.create(:organization) }
-    let(:current_user) { FactoryBot.create(:organization_admin, organization: organization) }
+    let(:current_user) { FactoryBot.create(:organization_admin, organization: current_organization) }
     before { log_in(current_user) }
   end
 
   RSpec.shared_context :request_spec_logged_in_as_organization_member do
     let(:current_organization) { FactoryBot.create(:organization) }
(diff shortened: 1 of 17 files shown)


spec/support/request_spec_helpers.rb
```diff
@@ -9,26 +9,31 @@ def log_in(user = nil)
     allow(User).to receive(:from_auth) { user }
   end
 
-  def log_in_as_superuser
-    user = FactoryBot.create(:admin)
-    log_in user
+  RSpec.shared_context :request_spec_logged_in_as_user do
+    let(:user) { FactoryBot.create(:user) }
```
jmromer: I'm -0 on replacing `log_in_as_superuser` and etc with shared contexts — the latter are available globally and the test suite generally relies too heavily on RSpec's DSL as it is. That's a good way to create confusion down the line. 

That said, if we do want to move in this direction, I'd suggest naming `user` and `organization` less opaquely (hence the naming in the previous version, which matches the pattern that you typically sees in implementation code and testing helpers provided by them (see: Devise's `current_user`). 

The `attr_readers` defined in this file are also not needed with this change.

sethherr: Switching to `current_user` and `current_organization` and removing the `attr_readers`s.

I personally like the `let` and context style and would prefer not to use instance variables in specs, so I want to maintain that functionality. If there is a way to reduce the scope of the contexts so they are only available in request specs, I'm absolutely on board.

jmromer: > would prefer not to use instance variables in specs

Agreed — just to clarify, that's not being proposed here.

PR #931, opened 2019-06-21
Add JS for new admin layout/News views

Robertgervais:


app/javascript/stylesheets/_admin.scss
```diff
@@ -176,3 +176,9 @@
     }
   }
 }
+
+.uppy {
```
sethherr: Please put this below the import statement, rather than over here

app/javascript/packs/pages/admin/blogs.js
```diff
@@ -0,0 +1,78 @@
+import log from "../../utils/log";
+
+// required for uppy file upload
+import 'es6-promise/auto'
+import 'whatwg-fetch'
```
sethherr: Remove duplicates

app/javascript/packs/pages/admin/blogs.js
```diff
@@ -0,0 +1,159 @@
...
+        });
+        this.closest('li').remove()
+      })
+    },
+
+    publicImageTemplate(image) {
+      const alt = image.name
+      const src = image.image.url
+      const id = image.id
+      const template =
```
sethherr: You can just return the string, rather than assigning to a variable

Robertgervais:
Okay, so I did add just an itty bit of JS to check if the delete button/primary image button need to be hidden. Normally I'd try and run a ruby conditional to see if a class should be added, but since I can't get any ruby to run in the template I wasn't sure how else to hide it. Let me know if you have a better suggestion! Also, I'm pretty sure that for now this is just running on blogs and custom layouts, in the future we can use a different way of checking. I was thinking basically, `if not location not blogs`, but right now it's just checking the href for `organizations`

PR #941, opened 2019-06-24
Add notes field to theft alerts

sethherr:
So that we can track updates and changes to the add, and any additional information

### DIFF
--- a/app/controllers/admin/theft_alerts_controller.rb
+++ b/app/controllers/admin/theft_alerts_controller.rb
@@ -1,6 +1,8 @@
 class Admin::TheftAlertsController < Admin::BaseController
   layout "new_admin"
 
+  before_action :find_theft_alert, only: [:edit, :update]
+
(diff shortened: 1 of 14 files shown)


jmromer:
@sethherr Added f3c77db -- introduces an `update_details` transition, which allows updating the recognized attributes (here, the facebook url and notes fields) without changing the status.

jmromer:
>  I feel like explicitly allowing admins to update the state would be easier and less prone to surprises

I'll read *state* as `status` here since the state transitions are exposed via the UI. 

There's more development work involved with abstracting away `status` updates, so I can definitely see that as a cost, but the intention there is to preclude surprises in the sense of inconsistent states that might be reached by allowing direct access to the components of each state (the known unknowns being more salient to me than unknown unknowns).

Happy to simplify that if you think it's overkill though—I can also see the case for allowing more granular control since it's an admin panel feature.

PR #943, opened 2019-06-24
Remove TheftAlert state management logic

jmromer:
Exposes more granular control via the admin panel

Screen Shot 2019-06-24 at 6 07 36 PM
Screen Shot 2019-06-24 at 6 07 51 PM


app/views/admin/theft_alerts/edit.html.haml
```diff
@@ -6,7 +6,18 @@
...
     html: { class: "m-0" } do |f|
-    = hidden_field_tag :state_transition, params.fetch(:state_transition, "update_details")
+    .form-group
+      = f.label :status, class: "form-label"
+      = f.select :status, options_for_select(TheftAlert.statuses.keys), class: "form-control"
+    .form-group
+      = f.label :theft_alert_plan_id, class: "form-label"
+      = f.select :theft_alert_plan_id, options_for_select(TheftAlertPlan.active.map{|e| [e.title_string, e.id]}), class: "form-control"
+    .form-group
+      = f.label :begin_at, class: "form-label"
```
sethherr: A few requests:

1. Can you automatically set the `begin_at` when the alert is switched to begin, and set the `end_at` based on that?
2. Can you hide the times and only show them after the alert is enabled?
3. Can you use `datetime_local_field`, include a timezone hidden field, and use `TimeParser` to set the time?

app/views/admin/theft_alerts/edit.html.haml
```diff
@@ -6,7 +6,18 @@
     url: admin_theft_alert_path(@theft_alert),
     method: :patch,
     html: { class: "m-0" } do |f|
-    = hidden_field_tag :state_transition, params.fetch(:state_transition, "update_details")
+    .form-group
+      = f.label :status, class: "form-label"
+      = f.select :status, options_for_select(TheftAlert.statuses.keys), class: "form-control"
+    .form-group
+      = f.label :theft_alert_plan_id, class: "form-label"
```
sethherr: Do you think there is a reason admins need to be able to edit this?

jmromer: A scenario in which someone requests a plan change (potentially with accompanying changes to start/end time fields) is more plausible than changes being required to, say, `creator` or `stolen_record`.

sethherr: Fair. I can see the use of it for that. I also don't see a reason to enable editing creator or stolen_record

app/views/admin/theft_alerts/edit.html.haml
jmromer: Updated in  e476085 and 0ecd791

jmromer:
👍 Added in 30d0034c

PR #950, opened 2019-06-26
Forward locale param when provided, set locale in around_action

jmromer:
- Adds [middleware](https://github.com/ruby-i18n/i18n/blob/master/lib/i18n/middleware.rb) from a more recent (but incompatible, alas) version of i18n to make i18n configuration thread-safe. This resolved the locale-leaking bug locally.
- Refactors locale setting to use an `around_action` to more cleanly isolate I18n locale per-request
- Implements `default_url_options` to set the current locale param, if one is provided, in any Rails-generated links
- Updates the coffeescript setting the bike search query params on the welcome page to not strip the locale param in the link
- Updates the blog post link on the home page logo so it retains the locale param
- Adds a link to the feature flag dashboard to the admin dashboard dropdown menu

### DIFF
--- a/app/assets/javascripts/init.coffee
+++ b/app/assets/javascripts/init.coffee
@@ -160,9 +160,22 @@ window.updateSearchBikesHeaderLink = ->
       distance = null
     else
       localStorage.setItem('distance', distance)
+
   distance ||= 100
-  url = "/bikes?stolenness=proximity&location=#{location}&distance=#{distance}"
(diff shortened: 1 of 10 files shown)


jmromer:
> What makes it incompatible? Rails version?

Yup -- it's locked to 0.7 currently:

```
    rails (= 4.2.11) was resolved to 4.2.11, which depends on
      activesupport (= 4.2.11) was resolved to 4.2.11, which depends on
        i18n (~> 0.7)
```

PR #951, opened 2019-06-27
Introduce MoneyHelper view helper module

jmromer:
Introduces `MoneyHelper` to format currency values and convert between localized currencies, to enable currency conversion in addition to localized currency formatting (see below).

#### TODO

- `Counts.week_creation_chart` seems to have a dupe definition, should we delete one?
- The `protected` access control in `Counts` is a no-op. Is the intent for it to apply to the methods defined below it? If so, I'll move those into the `class << self`.

Screen Shot 2019-06-27 at 1 12 00 PM
Screen Shot 2019-06-27 at 1 12 09 PM

### DIFF
--- a/app/helpers/money_helper.rb
+++ b/app/helpers/money_helper.rb
@@ -1,6 +1,7 @@
 module MoneyHelper
   include MoneyRails::ActionViewExtension
 
+  # TODO: Add a bank implementation that fetches conversion rate values
   Money.default_bank.add_rate(:USD, :EUR, 0.88)
 
   def default_currency
--- a/app/helpers/money_helper.rb
+++ b/app/helpers/money_helper.rb
@@ -1,0 +1,17 @@
+module MoneyHelper
+  include MoneyRails::ActionViewExtension
+
+  Money.default_bank.add_rate(:USD, :EUR, 0.88)
+
+  def default_currency
+    t("number.currency.alphabetic_code")
+  end
+
+  def money_usd(dollars, exchange_to: default_currency)
+    Money.new(dollars * 100, :USD).exchange_to(exchange_to)
+  end
+
+  def as_currency(dollars_usd, exchange_to: default_currency)
+    money_without_cents_and_with_symbol(money_usd(dollars_usd, exchange_to: exchange_to))
+  end
+end
--- a/app/models/counts.rb
+++ b/app/models/counts.rb
@@ -29,9 +29,6 @@ class Counts
 
     def organizations; retrieve_for("organizations") end
 
-    # TODO: Dupe method definition
-    def week_creation_chart; retrieve_for("week_creation_chart") end
-
     def week_creation_chart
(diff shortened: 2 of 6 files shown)


app/models/counts.rb
```diff
@@ -27,11 +27,12 @@ def total_bikes; retrieve_for("total_bikes") end
 
     def stolen_bikes; retrieve_for("stolen_bikes") end
 
-    def week_creation_chart; retrieve_for("week_creation_chart") end
-
     def organizations; retrieve_for("organizations") end
 
-    def week_creation_chart;
+    # TODO: Dupe method definition
+    def week_creation_chart; retrieve_for("week_creation_chart") end
```
sethherr: Yeah, remove it!

app/helpers/money_helper.rb
```diff
@@ -0,0 +1,17 @@
+module MoneyHelper
+  include MoneyRails::ActionViewExtension
+
+  Money.default_bank.add_rate(:USD, :EUR, 0.88)
```
sethherr: Should we make this not static? Is there a standard way, e.g. Stripe?

app/models/counts.rb
```diff
@@ -63,6 +64,8 @@ def assign_recoveries_value
     def recoveries_value; retrieve_for("recoveries_value") end
   end
 
+  # TODO: This `protected` has no effect. Do we want to move the methods below
```
sethherr: Ha, good to know. No reason not to move them into the self then

app/helpers/money_helper.rb
```diff
@@ -0,0 +1,17 @@
+module MoneyHelper
```
sethherr: Does it make sense to put this in helpers? I think we're using helpers as Rails helpers, which are view things. Is this just for frontend, or are we going to use it in other places?

app/helpers/money_helper.rb
jmromer: I think so—I'll add a TODO comment. There's a few different [backends available](https://github.com/RubyMoney/money#implementations) for live conversions, I'll take a look at them for the next PR. This seems adequate for now for the figures reported by `Counts` though.

app/helpers/money_helper.rb
jmromer: I think so since it's meant to abstract away from calls to `money_without_cents_and_with_symbol`, which is a view helper method, and it accesses the localized currency string. `money_usd` could be private, but I'm not sure there's straightforward to do in a view helper module.

app/helpers/money_helper.rb
sethherr: Sounds good. I'm all for not blocking on this, just curious if there was a good way. From a cursory look at [Stripe's Currency Conversion](https://stripe.com/docs/currencies/conversions) it seems like they aren't the solution.

I think we can punt on this for now, and deal with it once we're finished language things.

sethherr: Added an issue to track this #952

Issue #963, opened 2019-07-01
i18n Externalizations tracking


jmromer:
### Files to externalize

- Excludes admin templates / mailers and API documentation
- Includes currently unused templates

### templates
<details>
<summary>listing</summary>

<details>
<summary><code>welcome</code></summary>

- [x] `_root_organization_type_list.html.haml`
- [x] `choose_registration.html.haml`
- [x] `goodbye.html.haml`
- [x] `index.html.haml`
- [x] `recovery_stories.html.haml`
- [x] `update_browser.html.haml`
- [x] `user_home.html.haml`

</details>

<details>
<summary><code>layouts</code></summary>

- [x] `_grid.html.erb`
- [x] `_messages.html.haml`
- [x] `_messages_u.html.haml`
- [x] `_revised_messages.html.haml`
- [x] `application.html.haml`
- [x] `application_revised.html.haml`
- [x] `application_revised_bikehub.html.haml`
- [x] `email.html.haml`
- [x] `embed_layout.html.haml`
- [x] `embed_user_layout.html.haml`
- [x] `payments_layout.html.haml`
- [x] `reg_embed.html.haml`

</details>

<details>
<summary><code>shared</code></summary>

- [x] `_alert.html.haml`
- [x] `_charge.html.erb`
- [x] `_content_footer_ad.html.haml`
- [x] `_content_skeleton.html.haml`
- [x] `_donate_modal.html.haml`
- [x] `_edit_bike_skeleton.html.haml`
- [x] `_email_bike_box.html.haml`
- [x] `_faq.html.haml`
- [x] `_footer.html.haml`
- [x] `_footer_revised.html.haml`
- [x] `_footer_updated.html.haml`
- [x] `_form_well_footer_save.html.haml`
- [x] `_header.html.haml`
- [x] `_header_nav.html.haml`
- [x] `_modal.html.haml`
- [x] `_organized_skeleton.html.haml`
- [x] `_period_select.html.haml`
- [x] `_share_this.html.erb`
- [x] `_shops_map.html.erb`
- [x] `_top_user_nav.html.haml`
- [x] `_trash_icon.html.haml`

</details>

<details>
<summary><code>sessions</code></summary>

- [x] `new.html.haml`

</details>

<details>
<summary><code>landing_pages</code></summary>

- [x] `_for_organizations_photo_list.html.haml`
- [x] `_next_steps_contact.html.haml`
- [x] `_packages_list.html.haml`
[...]

PR #964, opened 2019-07-01
Externalize welcome#index 

jmromer:
- Adds welcome#index translations to en.yml
- Updates config to raise error when translations are missing (in dev and test environments only)

Screen Shot 2019-07-01 at 10 46 36 AM

Screen Shot 2019-07-01 at 10 46 24 AM

Related:  #963


jmromer:
> Does it make sense to include the full path in the translation blocks, to make for easier searching?

My pre-theoretic intuition is that the added verbosity wouldn't add *that* much of a searchability benefit...? I worry a bit too about full paths encouraging re-use across templates. I'll give it some more thought though.

PR #967, opened 2019-07-02
Externalize layout strings, Add cache helper

jmromer:
- Adds a CacheHelper#cache method, which adds the current locale to cache keys. Resolves an issue where a cached fragment in one locale is displayed when a different locale is selected.

Externalizes strings for the following templates in `app/views/layouts`:

-   _grid.html.erb
-   _messages.html.haml
-   _messages_u.html.haml
-   _revised_messages.html.haml
-   application.html.haml
-   application_revised.html.haml
-   application_revised_bikehub.html.haml
-   email.html.haml
-   embed_layout.html.haml
-   embed_user_layout.html.haml
-   payments_layout.html.haml
-   reg_embed.html.haml


app/views/layouts/_revised_messages.html.haml
```diff
@@ -1,20 +1,21 @@
...
-/   We primarily use 3 flash types - success, error, and notice
-/   success - No actionable info. Basically, Bike Index did what you asked. Auto faded
-/   error   - Messages because you did something wrong. Not faded
-/   notice  - Messages that don't fit in the above categories. Not faded, may have actionable info
+-# Note:
+-# -----
+-# We primarily use 3 flash types - success, error, and notice
+-# success - No actionable info. Basically, Bike Index did what you asked. Auto faded
+-# error   - Messages because you did something wrong. Not faded
+-# notice  - Messages that don't fit in the above categories. Not faded, may have actionable info
```
sethherr: Why switch comment types?

app/helpers/cache_helper.rb
```diff
@@ -0,0 +1,9 @@
+# frozen_string_literal: true
+
+module CacheHelper
+  # Override ActionView `cache` helper, adding the current locale to the cache
+  # key.
+  def cache(key = {}, options = {}, &block)
+    super([key, locale: I18n.locale], options, &block)
```
sethherr: Something this makes me think of: admin.

- We're caching in admin
- Admin pages shouldn't be translated (at least for the time being)
- Your `I18n` locale should always be english for admin pages (at least for the time being)

app/views/layouts/_revised_messages.html.haml
jmromer: In Haml the slash renders as an HTML comment so it ends up rendered for the client

sethherr: Oooh. Because it won't be sent to the client. Nifty! Good addition 👍

app/helpers/cache_helper.rb
jmromer: That was my assumption -- I'm not working on the admin templates.

app/views/layouts/application_revised_bikehub.html.haml
```diff
@@ -8,39 +8,41 @@
...
           %a.bikehub-nav-item{ href: new_session_path(partner: sign_in_partner), class: (signup_page ? "" : "active") }
-            Log in
+            = t(".log_in")
     .container.bikehub-explanation.mb-4
       .row
         .col-sm-12
           %h1
-            BikeHub
-            = signup_page ? "Sign up" : "Log In"
+            = t(".bikehub")
```
sethherr: Proper name, shouldn't be translated.

app/helpers/cache_helper.rb
sethherr: Fair. The only reason this is pertinent to this PR is with this caching addition, if someone is in admin but has a different language header set, it will not use the correct cache.

jmromer: I'm foreseeing some (potential) duplicated caching (in practice not so much), am I missing something else though? 🤔 

In particular: If I have Dutch set as my locale and then switch to admin, cached admin HTML fragments will be duplicated under `locale/nl/`.

sethherr: No, this is the only thing. It just seems like something that would be easy to avoid by checking if `controller_namespace == "admin"`

`controller_namespace` is defined in [controller_helpers.rb](https://github.com/bikeindex/bike_index/blob/master/app/controllers/concerns/controller_helpers.rb#L83)

app/views/layouts/application_revised_bikehub.html.haml
jmromer: That's a useful smell -- generally it's a good idea to have entries be complete semantic units as much as possible

app/views/layouts/embed_layout.html.haml
```diff
@@ -23,18 +23,18 @@
           .alert.alert-error
             %a.close{"data-dismiss" => "alert"} ×
             %h4
-              Please fix the following 
-              = pluralize(@b_param.bike_errors.count, "error")
+              = t(".please_fix_the_following")
+              = pluralize(@b_param.bike.errors.count, t(".error"))
```
sethherr: Oops, typo. Fixed in f6551a4efe046d4979dd547aeb3953544c6a62e7

jmromer: Damn, missed that. Thanks!

PR #969, opened 2019-07-02
Add ParamsNormalizer for boolean normalization

sethherr:
To enable replacing all the places we're doing:

```ruby
ActiveRecord::Type::Boolean.new.type_cast_from_database(whatever)
```


sethherr:
if you think this is reasonable @jmromer, I'll replace all the places this is done

PR #971, opened 2019-07-02
Externalize strings from shared partials

jmromer:
Externalize strings from the following templates in `app/views/shared`:

-   _alert.html.haml
-   _charge.html.erb
-   _content_footer_ad.html.haml (removed since empty)
-   _content_skeleton.html.haml
-   _donate_modal.html.haml
-   _edit_bike_skeleton.html.haml
-   _email_bike_box.html.haml
-   _faq.html.haml
-   _footer.html.haml
-   _footer_revised.html.haml
-   _footer_updated.html.haml
-   _form_well_footer_save.html.haml
-   _header.html.haml
-   _header_nav.html.haml
-   _modal.html.haml
-   _organized_skeleton.html.haml
-   _period_select.html.haml
-   _share_this.html.erb
-   _shops_map.html.erb
-   _top_user_nav.html.haml (removed since not used)
-   _trash_icon.html.haml

Related: #963


app/views/feedbacks/index.html.haml
```diff
@@ -3,15 +3,15 @@
 
 = render 'shared/faq'
 
-= render '/shared/content_footer_ad'
```
jmromer: @sethherr Is this safe to delete? Doesn't seem to be in use anymore, but if we want to keep the empty partial around, I'll revert the commit.

sethherr: Yes. You can delete this. Good catch.

PR #975, opened 2019-07-03
Set locale to English when requesting admin pages

### LINKED ISSUE #968: Localization in Admin
- Admin pages shouldn't be translated
- Your I18n locale should always be english for admin pages

jmromer:
Sets locale to english when requesting admin pages, since these will not be localized.

Resolves #968

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -92,7 +92,7 @@ class ApplicationController < ActionController::Base
 
   # TODO: Remove logging in #requested_locale when feature flag is removed
   def set_locale
-    if !Flipper.enabled?(:localization, current_user)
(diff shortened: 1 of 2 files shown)


app/controllers/application_controller.rb
```diff
@@ -92,10 +92,14 @@ def requested_locale
 
   # TODO: Remove logging in #requested_locale when feature flag is removed
   def set_locale
-    if Flipper.enabled?(:localization, current_user)
-      I18n.with_locale(requested_locale) { yield }
-    else
+    if !Flipper.enabled?(:localization, current_user)
```
sethherr: Could you make this an unless? I just think it reads better.

PR #991, opened 2019-07-08
Externalize users/ templates

jmromer:
Externalizes the following `users` templates:

-   _edit_password.html.haml
-   _edit_root.html.haml
-   _edit_sharing.html.haml
-   accept_terms.html.haml
-   accept_vendor_terms.html.haml
-   confirm_error_404.html.haml
-   confirm_error_bad_token.html.haml
-   edit.html.haml
-   globalid.html.haml
-   new.html.haml
-   organization_show.html.haml
-   password_reset.html.haml
-   please_confirm_email.html.haml
-   request_password_reset.html.haml
-   show.html.haml
-   update_password.html.haml


app/views/users/organization_show.html.haml
```diff
@@ -5,77 +5,59 @@
     = @user.email
```
sethherr: What renders this template (app/views/users/organization_show.html.haml) - and do we still need it?

jmromer: I can't see it rendered it from anywhere, and it hasn't been modified since 2013, so probably safe to delete 👍

PR #995, opened 2019-07-09
remove organization invitations

sethherr:
Cleaning up after #992


config/locales/en.yml
```diff
@@ -4,57 +4,57 @@ en:
   language: English
   date:
     abbr_day_names:
-    - Sun
```
sethherr: @jmromer - I'm not sure if these changes are good? If they break things, I'll revert them.

This is just what my prettier did 🤷‍♂️

PR #997, opened 2019-07-09
Usability improvements for EPS

sethherr:


sethherr:
still to do: add class to default_bike_search_path everywhere

PR #1001, opened 2019-07-10
Externalize bikes templates

jmromer:
Externalizes strings for the following templates:

- [x] `app/views/bikes/_bike.html.haml`
- [x] `app/views/bikes/_bike_org_table.html.haml`
- [x] `app/views/bikes/_bike_show_overlays.html.haml`
- [x] `app/views/bikes/_bike_table.html.haml`
- [x] `app/views/bikes/_bike_u.html.haml`
- [x] `app/views/bikes/_edit_primary_menu.html.haml`
- [x] `app/views/bikes/_organized_access_panel.html.haml`
- [x] `app/views/bikes/_stolen_checklist.html.haml`
- [x] `app/views/bikes/_stolen_map.html.erb`
- [x] `app/views/bikes/_theft_alerts_table.html.haml`
- [x] `app/views/bikes/bike_fields/_revised_colors.html.haml`
- [x] `app/views/bikes/bike_fields/_revised_component_fields.html.haml`
- [x] `app/views/bikes/bike_fields/_stolen_location.html.haml`
- [x] `app/views/bikes/edit_accessories.html.haml`
- [x] `app/views/bikes/edit_alert.html.haml`
- [x] `app/views/bikes/edit_alert_purchase.html.haml`
- [x] `app/views/bikes/edit_alert_purchase_confirmation.html.haml`
- [x] `app/views/bikes/edit_bike_details.html.haml`
- [x] `app/views/bikes/edit_drivetrain.html.haml`
- [x] `app/views/bikes/edit_groups.html.haml`
- [x] `app/views/bikes/edit_ownership.html.haml`
- [x] `app/views/bikes/edit_photos.html.haml`
- [x] `app/views/bikes/edit_publicize.html.haml`
- [x] `app/views/bikes/edit_remove.html.haml`
- [x] `app/views/bikes/edit_report_recovered.html.haml`
- [x] `app/views/bikes/edit_report_stolen.html.haml`
- [x] `app/views/bikes/edit_theft_details.html.haml`
- [x] `app/views/bikes/index.html.haml`
- [x] `app/views/bikes/new.html.haml`
- [x] `app/views/bikes/pdf.html.haml`
- [x] `app/views/bikes/scanned.html.haml`
- [x] `app/views/bikes/show.html.haml`
- [x] `app/views/bikes/spokecard.html.haml`


app/controllers/application_controller.rb
```diff
@@ -63,6 +63,10 @@ def locale_from_request_params
     requested_locale if I18n.available_locales.include?(requested_locale)
   end
 
+  def t(key)
+    I18n.t(key, scope: [:controllers, controller_name])
```
sethherr: You should include the `controller_namespace` from `ControllerHelpers`

```ruby
[:controllers, controller_namespace, controller_name].compact
```

app/views/bikes/_bike.html.haml
```diff
@@ -6,18 +6,17 @@
...
-        = bike.attr_list_item(bike.serial_display, 'Serial')
-        = bike.attr_list_item(bike.frame_colors.to_sentence, 'Primary colors')
+        = bike.attr_list_item(bike.serial_display, t(".serial"))
+        = bike.attr_list_item(bike.frame_colors.to_sentence, t(".primary_colors"))
       - if bike.current_stolen_record.present?
         %ul.attr-list
           %li
             %span.attr-title
-              = bike.recovered ? "Recovered" : "Stolen"
+              = bike.recovered ? t(".recovered") : t(".stolen")
```
sethherr: We do this a lot. Should we translate this in the model?

app/views/bikes/_bike_org_table.html.haml
```diff
@@ -21,11 +21,11 @@
             - if bike.created_at.today?
               %span.hidden
                 = bike.created_at.strftime("%-m.%-d.%Y %-H.%M")
-              Today at #{bike.created_at.strftime("%-I %p")}
+              = t(".today_at", created_at: bike.created_at.strftime('%-I %p'))
```
sethherr: Anytime you see dates, please switch them to use the `convertTime` path.

app/views/bikes/_bike_org_table.html.haml
```diff
@@ -2,17 +2,17 @@
   %table.table.table-striped.table-bordered
```
sethherr: Are we rendering this template anywhere or is this dead?

app/views/bikes/_bike_table.html.haml
```diff
@@ -1,16 +1,11 @@
 %section.full-screen-table
```
sethherr: I'm not sure this file is used anymore. If it is, please switch to `convertTime` 😉

app/views/bikes/_bike.html.haml
sethherr: Or maybe it should be in a helper?

sethherr: I think this should use the same thing we use for the other model level strings - like the names in `CycleType` and `Color`

app/views/bikes/_organized_access_panel.html.haml
```diff
@@ -8,51 +8,46 @@
...
+                    = link_to t(".edit"), edit_bike_path(@bike), class: "btn btn-success btn-sm float-right"
                   - else
                     %em.small.less-strong
-                      #{passive_organization.short_name} can not edit
+                      = t(".org_cannot_edit", org_name: passive_organization.short_name)
               - if passive_organization.paid_for?("bike_codes") # Always display stickers if org has paid for them
                 %tr
                   %td
-                    Sticker
+                    = t(".sticker")
```
sethherr: Not sure if this is helpful, but this could reference the `BikeCode.kind` enum

app/views/bikes/bike_fields/_revised_colors.html.haml
```diff
@@ -1,38 +1,43 @@
 - classname ||= ''
 - bike ||= @bike
-- primary_color_help_content = "The color of the frame and fork&mdash;not the wheels, cranks, or anything else. You can put a more detailed description in paint description (once you've registered), <em>this</em> is to get a general color to make searching easier"
+
+- primary_color_help_content = t(".primary_color_help_content_html")
```
sethherr: nit: since this is now a translation, rather than inline string, it doesn't need to be defined separately.

Honestly, I just did this because I didn't know how to do multiline haml tags 😅

app/views/bikes/_bike_org_table.html.haml
jmromer: Just to clarify -- are you referring to a translation key or the css class? I'm seeing `time.format.convert_time` and `convertTime`, respectively

sethherr: If you put a `time.format.convert_time` inside of something with a class of `convertTime`, js (moment) converts the time to the user's time zone.

This is the best solution to time localization that I've seen or come up with.

sethherr: FYI: the localization js also converts time that is in timestamp format, which is useful because our API returns timestamps

jmromer: Cool. Only on occurrence in this batch, in app/views/bikes/pdf.html.haml

app/views/bikes/_bike.html.haml
jmromer: A helper makes sense, I'll circle back to that. I'll do some research re CycleType and Color.

app/views/bikes/edit_drivetrain.html.haml
```diff
@@ -1,96 +1,96 @@
...
-                  #{side.titleize} wheel
+                  = t(".sidetitleize_wheel", sidetitleize: side.titleize)
                 .form-well-input.fancy-select-placeholder.unfancy
                   .standard-diams{ id: "#{side}_standard"}
-                    // ignore this select in dirty forms
+                    -# ignore this select in dirty forms
                     %select.form-control.dirtyignore
                       %option{ value: '' }
-                        Choose diameter
+                        = t(".choose_diameter")
```
sethherr: I think you can remove this translation section, because it will be dealt with in the helper functionality.

app/views/bikes/_bike_org_table.html.haml
sethherr: Ooooh. well, in that case we can't use the `convertTime` path here, because we can't do js in the pdf.

... But we don't need to do this whole today time thing, since the pdf is only generated once. We can just use the `short_with_hour` format

app/views/bikes/edit_theft_details.html.haml
```diff
@@ -39,13 +41,13 @@
...
 
-            = render partial: '/bikes/bike_fields/stolen_location', locals: { srecord: srecord, stolen_type: stolen_type }
+            = render partial: 'bikes/bike_fields/stolen_location', locals: { srecord: srecord, stolen_type: stolen_type }
 
             .form-group.row.unnested-field
               = srecord.label :estimated_value, class: 'form-well-label'
               .form-well-input.input-group{ style: 'max-height: 2.5rem;'}
                 .input-group-addon
-                  $
+                  = t("number.currency.format.unit")
```
sethherr: If we're accepting things in non-usd format, we need to store the unit they're being submitted in :/

sethherr: I think it makes sense to punt on dealing with that for now, and just make everyone submit in USD for now

app/views/bikes/new.html.haml
```diff
@@ -13,111 +14,114 @@
...
           - if @bike.stolen
             = fields_for @stolen_record do |srecord|
-              = render partial: '/bikes/bike_fields/stolen_location', locals: { srecord: srecord }
+              = render partial: 'bikes/bike_fields/stolen_location', locals: { srecord: srecord }
 
           - if @organization&.include_field_reg_address?(current_user)
             .related-fields
               .form-group.row
-                - label_address = @organization.registration_field_label("reg_address") || "Mailing address"
+                - label_address = @organization.registration_field_label("reg_address") || t(".mailing_address")
```
sethherr: nit: could be moved to a multi-line haml tag

app/views/bikes/edit_theft_details.html.haml
jmromer: Cool. Just a head's up that pricing plans are also not localized yet -- right now we're localizing the currency format but not the actual price (since I'm assuming we'll want to establish price points independently). Alternatively, we could just convert currency into local currency.

sethherr: It looks like Stripe will convert when we send the charge to them? 

We might want to just always display USD for now.

app/views/bikes/show.html.haml
```diff
@@ -3,51 +3,52 @@
...
   .row
-    / Bike Photo!
+    -# Bike Photo!
     .col-md-8
       .bike-photos.horizontal-thumbnails
         - if @bike.public_images.any?
           #selected-photo{ class: "image-holder #{"just1photo" if @bike.public_images.count == 1}" }
             .current-photo{ id: "image#{@bike.public_images.first.id}"}
-              = image_tag(@bike.public_images.first.image_url(:large), alt: "#{@bike.type.titleize} photo", id: "i|#{@bike.public_images.first.listing_order}", data: { action: 'zoom', original: @bike.public_images.first.image_url })
+              = image_tag(@bike.public_images.first.image_url(:large), alt: t(".bike_photo", bike_type: @bike.type.titleize), id: t(".ibikepublic_imagesfirstlisting_order", bikepublic_imagesfirstlisting_order: @bike.public_images.first.listing_order), data: { action: t(".zoom"), original: @bike.public_images.first.image_url })
```
sethherr: this `id` shouldn't have been modified. Fixing in #1016

jmromer: Thanks! Looks like `data` was a spurious translation too

PR #1010, opened 2019-07-10
Add Magic Link sign in method

sethherr:
Screen Shot 2019-07-12 at 8 36 10 PM


app/mailers/customer_mailer.rb
```diff
@@ -22,6 +22,12 @@ def password_reset_email(user)
     mail(to: @user.email)
   end
 
+  def magic_login_link_email(user)
+    @user = user
+    @url = "#{ENV["BASE_URL"]}/session/magic_link?token=#{@user.magic_link_token}"
```
jmromer: Would `magic_link_session_url` not work here?

app/services/security_tokenizer.rb
```diff
@@ -0,0 +1,12 @@
+class SecurityTokenizer
+  def self.new_token(time = nil)
+    t = (time || Time.current).to_i
+    "#{t}-" + SecureRandom.hex + Digest::MD5.hexdigest("#{SecureRandom.hex}-#{t}")
+  end
+
+  def self.token_time(str)
+    t = str.to_s.split("-")[0]
+    t = (t.present? && t.to_i > 1427848192) ? t.to_i : 1364777722
```
jmromer: What do these [magic numbers](http://wiki.c2.com/?MagicNumber) signify?

app/mailers/customer_mailer.rb
sethherr: I forget why I switched to doing this for the token emails, but I did, so I'm hesitant to switch back.

app/services/security_tokenizer.rb
sethherr: Good call, good link

app/mailers/customer_mailer.rb
jmromer: A consideration: If nothing else, this approach strips the locale query param when present.

sethherr: Well... it's only for these emails. And since these emails are sent from worker processes, I can't see how the locale query would be preserved?

I see your point though, and adding in the locale query will be important in the emails.

I'm going to ship this with the string URL and open a PR using the url helpers, to ship after, so that we aren't making too many changes at once.

jmromer: Ah true. We'd have to pass into the worker and then maybe invoke the mailer in a `with_locale` block... their might be a less intrusive way to do that second bit, though.

sethherr: We might want to assign the user's language when they submit the new user form (from their header or request or whatever).

And then we can consistently translate based on the user

PR #999, opened 2019-07-09
Externalize organized templates

jmromer:
Externalizes strings from the following organized templates:

-   ambassador_dashboards/getting_started.html.haml
-   ambassador_dashboards/resources.html.haml
-   ambassador_dashboards/show.html.haml
-   bikes/_list.html.haml
-   bikes/_search.html.haml
-   bikes/incompletes.html.haml
-   bikes/index.html.haml
-   bikes/multi_serial_search.html.haml
-   bikes/new.html.haml
-   bikes/recoveries.html.haml
-   bulk_imports/index.html.haml
-   bulk_imports/new.html.haml
-   bulk_imports/show.html.haml
-   exports/index.html.haml
-   exports/new.html.haml
-   exports/show.html.haml
-   manage/_location_fields.html.haml
-   manage/index.html.haml
-   manage/locations.html.haml

(There's a few more in this batch, but it's turning out to be more involved than anticipated, so I'll PR it in smaller chunks to keep the diffs relatively small.)

Also adds some tooling to format / audit translation files:

> This gem analyses code statically for key usages, such as I18n.t('some.key'), in order to:
> 
> - Report keys that are missing or unused.
> - Pre-fill missing keys, optionally from Google Translate or DeepL Pro.
> - Remove unused keys.
>
> Thus addressing the two main problems of i18n gem design:
> 
> - Missing keys only blow up at runtime.
> - Keys no longer in use may accumulate and introduce overhead, without you knowing it.

https://github.com/glebm/i18n-tasks

PR #1014, opened 2019-07-10
Organization exports fix

sethherr:


app/views/organized/exports/show.html.haml
```diff
@@ -45,7 +45,9 @@
 
             .collapse#show_exported_bikes
               %ul.exported-link-list
-                %li= bike_ids.map { |id| link_to(id, bike_path(id)) }.join(", ").html_safe
+                - bike_ids.each_with_index do |id, i|
+                  %li
+                    #{link_to id, bike_path(id)}#{i == (@export.exported_bike_ids.count - 1) ? "" : ","}
```
sethherr: @jmromer I'm not sure why you removed this in #999 ?

First, by switching to a map, it put everything in one `li`, which is a bug. And this `each_with_index` is a simple thing to add commas after the links. It's pretty hacky, but I don't see a reason to drop it?

Other than those actual functional issues, while I'm a fan of `map` in general, my impulse is to avoid using `.html_safe` whenever possible, because I think it's a good habit. I'm open to having this opinion changed though.

jmromer: > First, by switching to a map, it put everything in one li, which is a bug

The regression here is the misplaced `li`, not the introduction of the `map` — these are (strictly speaking) independent of each other, although to the extent the one impacts the other I'd argue that the added complexity of the hackier approach increases the risk of, e.g., misplacing an `li`, by undermining the legibility of the code here. 

(The catalyst for refactoring here in the first place is having to prep for the externalization script, although I think I may have figured out a less intrusive workaround at this point, so I should be able to leave these comma-adding loops alone moving forward.)  

> my impulse is to avoid using .html_safe

I'd assume that "avoid" here is subject to the constraint of not introducing hacky code, which I'd take as an overriding concern. 

That said, I'm close to indifferent between the two approaches if you want to revert to the loop. In my reasoning here I'm also taking as given the requirement of commas and the use of `li`s here.

sethherr: Correct, the missing `li` being the regression, not the map. 

I understood the catalyst. And it’s impressive so much has been updated with so few issues!

“Avoid” is pretty strong in this context, because `html_safe` opens up security issues. If there is any user modifiable content that gets wrapped in html_safe, we have a problem and the cognitive load of trying to think of ways in which someone could manage to modify content is large enough to overweigh fairly ugly stuff.

I think the truly non-hacky solution is done with `map` and something to do with content tags so you don’t need `html_safe`, then joining with commas. But I don’t know exactly how to do that, so I’m going to merge this.

jmromer: [`safe_join`](https://apidock.com/rails/ActionView/Helpers/OutputSafetyHelper/safe_join) is another approach:

```rb
links = [helper.link_to('link-one', ''), helper.link_to('link-two', '')]

links.join(", ".html_safe).html_safe?
# false

joined_str = helper.safe_join(links, ", ")
joined_str.html_safe?
# true
```

sethherr: Oh HELL yeah. 💚

PR #1015, opened 2019-07-11
Access ip address in consistent manner

sethherr:
Follow up to #1013

Things this does:

- Switch to using `forwarded_ip_address` for what is appended to our logging payload
- Always fallback in the same way, in all the places ip addresses are checked
- Move the `forwarded_ip_address` and ip appending to payload into controller helpers, which is included in controllers that don't inherit from `ApplicationController` (ie Doorkeeper controllers)
- Pass `forwarded_ip_address` to user and test that it is passed to user

### DIFF
--- a/app/controllers/api/base.rb
+++ b/app/controllers/api/base.rb
@@ -3,14 +3,10 @@ require "grape_logging"
 module GrapeLogging
   module Loggers
     class BinxLogger < GrapeLogging::Loggers::Base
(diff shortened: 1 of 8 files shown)


sethherr:
The issue is  in grape where we need to handle logging, `request` is passed as an argument, not globally available. 

~I’m absolutely interested if you’ve got a nice way to take care of this~

I see a way, I’ll extract it

PR #1019, opened 2019-07-11
Fix js TimeParser 

sethherr:
#1018 broken things because of a typo, also because we were calling localizeTimes in other places in the code.

I reverted that commit and will fix it here


app/javascript/packs/utils/time_parser.js
```diff
@@ -0,0 +1,85 @@
...
+        let text = $this.text().trim();
+        let time = null;
+        $this.removeClass("convertTime");
+        $this.addClass("convertedTime"); // So we can style it
+        if (!(text.length > 0)) {
+          return;
+        }
+        // If time is only a number, parse as a timestamp
+        // Otherwise, parse as ISO_8601 which is what convert_time strftimes into
+        if (/^\d+$/.test(text)) {
```
sethherr: Turns out I was wrong, or moment changed or something. It didn't actually parse timestamps prior to adding this.

PR #1020, opened 2019-07-11
Bump Ruby to 2.5.5

jmromer:
- Bumps the patch version to silence warnings from i18n-tasks.
~- Kills `FILTERED_PARAMS` setting in `Api::Base` to silence a warning~

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -12,7 +12,7 @@ jobs:
       TZ: /usr/share/zoneinfo/America/Chicago
 
     docker:
-      - image: circleci/ruby:2.5.1-stretch-node
+      - image: circleci/ruby:2.5.5-stretch-node
         environment:
           PGHOST: 127.0.0.1
           PGUSER: root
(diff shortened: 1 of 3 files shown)


app/controllers/api/base.rb
```diff
@@ -8,8 +8,6 @@ def parameters(request, _)
       end
     end
   end
-  # Need to add file parameters to this so it doesn't try (and fail) to make images json
-  FILTERED_PARAMS = Rails.application.config.filter_parameters + [:file]
```
sethherr: **We need this**.

Without it, we 500 when uploading a file because it tries to log the file as JSON - and errors, since the file is binary.

Potential solution: inline this, replacing the line in this file that references `FILTERED_PARAMS`

```ruby
GrapeLogging::Loggers::FilterParameters.new(GrapeLogging::FILTERED_PARAMS)
```

_p.s. sorry there aren't tests for this, I don't know how to test it_

jmromer: Looks like it was only a no-op for me because of Spring (the app being initialized already). I'll just remove the change for now.

PR #1027, opened 2019-07-12
Externalize legal, about pages

jmromer:
Externalizes strings from the following templates:

-  app/views/info/_terms_text.html.erb
-  app/views/info/_vendor_terms_text.html.erb
-  app/views/info/privacy.html.erb
-  app/views/info/about.html.haml

### DIFF
--- a/app/assets/stylesheets/revised/pages/info/about.scss
+++ b/app/assets/stylesheets/revised/pages/info/about.scss
@@ -60,14 +60,6 @@ $hover-about-social-color: lighten($gray-light, 15%);
     }
   }
 
-  .the-alumni {
-    display: none;
-  }
-
-  .about-alumni-list {
-    display: none;
-  }
-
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/revised/pages/info/about.scss
```diff
@@ -114,4 +90,14 @@ $hover-about-social-color: lighten($gray-light, 15%);
       }
     }
   }
+
+  .about-alumni-list {
+    list-style-position: outside;
```
jmromer: This fixes the positioning of alumni images, but that listing is still hidden via a `d-none` on the `ul`

sethherr: That was the intention 🙂 - hidden, but not removed. Thanks for fixing them even though they're invisible!

PR #1034, opened 2019-07-15
Use url helpers in token emails

sethherr:
[@jmromer asked why we weren't using the url paths](https://github.com/bikeindex/bike_index/pull/1010#discussion_r303441100) for are token emails.

I said:

> I forget why I switched to doing this for the token emails, but I did, so I'm hesitant to switch back.

From PR #1010


sethherr:
I'm not sure why, but locally with `letter_opener`, the link in the email doesn't have the token in it.

I just tested both the passwordless sign in and the reset password emails on production, and they both send the token - so I'm not reverting this change. It's just something to pay attention to.

PR #1035, opened 2019-07-15
Translation fixes

sethherr:
- Remove duplicated translation key
- Fix email translation key (so it doesn't show up unescaped, and because it doesn't need to be)

### DIFF
--- a/app/views/organized/users/new.html.haml
+++ b/app/views/organized/users/new.html.haml
@@ -31,7 +31,7 @@
     .col-sm-4#multipleEmailInviteField{ class: display_multi_invite ? "" : "currently-hidden" }
       = text_area_tag :multiple_emails_invited,
         params[:multiple_emails_invited],
-        placeholder: "example@example.com\nexample2@example.com\nexample3@example.com",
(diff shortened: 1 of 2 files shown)


jmromer:
Does `example` scan in Dutch?

jmromer:
Also, what's the duplicated key removed? I can't tell from the diff. 
I haven't normalized the translation file yet but that's inbound

sethherr:
The entirety of `info: terms_text:` is duplicated.

Touché on the example translation. Switched to `email@gmail.com` which I think is sufficiently universal.

jmromer:
Cool, just checking. 

Re the dupe key -- the changes in en.yml look like they'll be time-consuming to merge. Can we drop those for now? I'm anticipating the normalization script to catch duplicate keys.

sethherr:
I think the diff makes it look worse than it is.

Really, I just deleted line 2190 to 2319 (or 2189-2318, since I deleted the earlier key)

jmromer:
Really not a good use of time. This is what you're looking for:
```
git rebase --onto b574d2ffcae^ b574d2ffcae
```

PR #1036, opened 2019-07-15
Externalize remaining landing pages

jmromer:
These were missed because of `html.haml` extension was missing. 
This patch normalizes on that extension and externalizes strings from those templates.

Also kills the duplicated `terms_text` key pointed out in https://github.com/bikeindex/bike_index/pull/1035


sethherr:
For reviewing purposes, it would be great to switch the names first.

I'm opening a PR for the remaining templates

PR #1039, opened 2019-07-16
Generate model attr translation keys, normalize translation file

jmromer:
- Generates model translation keys
- Normalizes en.yml
- Removes duplicate and unused keys
- Re-enables i18n spec 
- Re-generates nl.yml with `TRME` flag
- Begins translating enums and colors etc (continued in follow-up PR https://github.com/bikeindex/bike_index/pull/1040)

Related: #https://github.com/bikeindex/bike_index/issues/963


app/models/color.rb
```diff
@@ -13,6 +13,17 @@ def self.black
...
 
+  def self.name_translation(name)
+    I18n.t(
+      name.to_s.downcase.gsub(/[^[:alnum:]]+/, "_"),
+      scope: [:activerecord, :options_for_select, self.name.underscore],
+    )
+  end
+
+  def self.select_options
+    all.pluck(:id, :name).map { |id, name| [name_translation(name), id] }
```
sethherr: So, assigning the translation is global? This looks crazy to me because the language isn't passed.

Is there any way, without making things hideous, to pass the language into this method and all other model methods?

sethherr: Even if we don't do it in the view.

I feel like it would be nice for testing and diagnostics purposes

jmromer: > So, assigning the translation is global? This looks crazy to me because the language isn't passed.

Can you elaborate? I'm not sure I follow. The locale is set per-request, not passed into [`i18n#translate`](https://rdoc.info/github/svenfuchs/i18n/master/I18n/Base:translate):

```haml
-# app/views/welcome/index.html.haml L42-44 (ebd2366d)

.root-landing-explanation-wrap
  %h4= t(".alert_the_community")
  %p= t(".if_your_bike_goes_missing")
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/ebd2366d/app/views/welcome/index.html.haml#L42-L44)]</sup>

sethherr: e.g. if I'm in console and want to see what the `select_options` in german. I understand if this isn't possible... but, if there is a way to do it without it being too terrible, I would appreciate it.

And I've rephrased "this looks crazy" (which was poor phrasing on my part). I think this is an elegant solution, it just has an extra bit of implicitness that it would be great to be able to make more explicit.

I don't feel the same way about things in the controller because it seems less surprising, somehow. And also because I don't run controller methods in console.

sethherr: > Even if we don't do it in the view

By which I mean that, if the default argument for this pulls the current i18n language, that would be super slick - so in the views you don't need to specify the language, because it's implicit.

It's also possible this is just because I'm less familiar with translation stuff, and I'll get over this.

jmromer: > e.g. if I'm in console and want to see what the select_options are german.

Ah gotcha thanks -- `I18n.with_locale` accomplishes what you have in mind:

```ruby
I18n.with_locale(:nl) { Color.select_options.first }
# [ "TRME Black", 1 ]

Color.select_options.first
# [ "Black", 1 ]
```

But the locale can also be parametrized, so e872e4e3 adds that too 👍

sethherr: AMAZING!

Thanks for the note about `I18n.with_locale(:nl)`, that's good to know.

PR #1040, opened 2019-07-16
Add attribute model name translations

jmromer:
Add translations for attribute strings stored in the database:

- [x] Color
- [x]  Ctype
- [x] Country (https://github.com/onomojo/i18n-country-translations)
- [x]  StolenRecord
- [x]  ~Cgroup~ (not referenced outside of admin)
- [x]  ~WheelSize~ (using un-localized)
- [x]  ~AmbassadorTask~ (to be implemented in a separate PR -- some DB migration involved)
- [x] Refactor / consolidate `Selectable`
- [x] Add locale tests for Selectable methods


app/models/color.rb
```diff
@@ -13,17 +13,14 @@ def self.black
...
 
-  def self.name_translation(name, locale: nil)
-    I18n.t(
-      name.to_s.downcase.gsub(/[^[:alnum:]]+/, "_"),
-      scope: [:activerecord, :select_options, self.name.underscore],
-      locale: locale,
-    )
-  end
+  def self.select_options
+    normalize = ->(value) { value.to_s.downcase.gsub(/[^[:alnum:]]+/, "_") }
```
sethherr: What's the impact on load times for this? Do we need to do caching?

jmromer: Negligible, and caching introduces some complexity with testing since this is a class method.

PR #1046, opened 2019-07-17
Automatically generate users

sethherr:
Because law enforcement wants the ability to add users who don't have passwords, who never see passwords, because passwords are bad or whatever.

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -755,7 +755,7 @@ en:
         sail: Sail
     errors:
       messages:
-        record_invalid: "Validation failed: %{errors}"
+        record_invalid: 'Validation failed: %{errors}'
         restrict_dependent_destroy:
           has_many: Cannot delete record because dependent %{record} exist
           has_one: Cannot delete record because a dependent %{record} exists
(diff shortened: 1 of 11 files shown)


config/locales/en.yml
```diff
@@ -755,7 +755,7 @@ en:
         sail: Sail
     errors:
       messages:
-        record_invalid: 'Validation failed: %{errors}'
+        record_invalid: "Validation failed: %{errors}"
```
sethherr: @jmromer - there are a bunch of changes to this file which are caused by my [PrettyYAML](https://github.com/aukaost/SublimePrettyYAML) editor plugin. Obviously this is unacceptable.

Which of these options would you prefer:

1. I don't use pretty yaml on this file
2. We figure out how to format the translation file in a way that follows along with the formatting from my pretty yaml

I assume option 1 is the preferred option. The changes look nice to me, but they broke the translation keys.

sethherr: I just normalized the file with `i18n-tasks normalize`, this output should pass.

jmromer: That'll work! There's a spec to ensure that translation files conform with what `i18n-tasks normalize` would produce. Not sure how you feel about pre-commit hooks, but that's a possibility too if this starts to be a hassle.

sethherr: This actually causes sublime to get real slow and is a bit of a pain - and because of the length of the file, and the number of these files we will eventually have, I think it will create too many unnecessary changes.

I've reverted the change and ignore locale files in prettier.

jmromer: That's been my experience with pre-commit hooks too—they can add a lot of friction to local development, so I'm of two minds about them.

PR #1050, opened 2019-07-17
Error if token is blank for emails requiring token

sethherr:
Looks like https://github.com/bikeindex/bike_index/pull/1034#issuecomment-512509125 was a valid bug.

Investigating fixes - I believe it isn't the fault of the url helpers, but instead something to do with the database


sethherr:
That didn't resolve it, it appears to be confined to users generated by the automatic user creation process from #1046

sethherr:
Turns out, I was emailing the wrong link. This issue definitely had nothing to do with url_helpers

PR #1052, opened 2019-07-17
Prepare translations rake task

sethherr:
It would be really cool if we still got the colors from the `i18n-tasks`

### DIFF
--- a/lib/tasks/rake_tasks.rake
+++ b/lib/tasks/rake_tasks.rake
@@ -20,7 +20,9 @@ end
 
 desc "Prepare translations for committing to master"
 task prepare_translations: :environment do
-  `i18n-tasks add-missing -v 'TRME %{value}' nl`
-  `i18n-tasks normalize`
-  `i18n-tasks health`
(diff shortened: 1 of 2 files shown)


sethherr:
Read this on executing commands in ruby: https://blog.bigbinary.com/2012/10/18/backtick-system-exec-in-ruby.html, and so tried this:

```ruby
[
    "i18n-tasks add-missing -v 'TRME %{value}' nl",
    "i18n-tasks normalize",
    "i18n-tasks health",
  ].each do |cmd|
    Open3.popen3(cmd) do |_stdin, _stdout, _stderr, _wait_thr|
      while line = stderr.gets
        puts line
      end
    end
  end
```

It didn't work. I give up, it doesn't matter that much.

jmromer:
@sethherr Hope you don't mind -- added e4aa3ea9, that retains the output color 👍 

Screen Shot 2019-07-17 at 7 15 05 PM

sethherr:
Thanks!

PR #1056, opened 2019-07-18
Deliver emails localized

jmromer:
- [x] ~Externalize mailer templates~ (extracted to follow-on PR)
- [x] ~Pass locale to workers for mailers~ (Lots of changes required for this approach, can revisit if needed)
- [x] Set preferred locale for user on signup and use in mailers

### DIFF
--- a/config/locales/customer_mailer.en.yml
+++ b/config/locales/customer_mailer.en.yml
@@ -12,3 +12,9 @@ en:
       subject: Sign in to Bike Index
     additional_email_confirmation:
       subject: Confirm your additional email
+    recovered_from_link:
+      subject: Your %{biketype} has been marked recovered!
+    stolen_notification_email:
+      subject: Stolen bike contact
+    updated_terms_email:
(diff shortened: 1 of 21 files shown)


config/locales/customer_mailer.en.yml
```diff
@@ -1,14 +1,21 @@
+---
 en:
   customer_mailer:
-    welcome_email:
-      subject: Welcome to Bike Index!
+    additional_email_confirmation:
```
sethherr: Why are just the subjects in this file?

Does it make more sense to put this stuff in `en.yml`?

jmromer: > Does it make more sense to put this stuff in en.yml?

That's my inclination — I'm not sure why the mailer localization files are broken out like this. I'm following suit for now but I want to try to consolidate them into en.yml / nl.yml.

sethherr: They were put there a while ago by me, because I was just doing random things.

PR #1057, opened 2019-07-18
Move Sidekiq options to ApplicationWorker

jmromer:
- Adds `ApplicationWorker` to set default sidekiq options
- Extracted from #1056. Taking a different approach to locale-setting there but we may want to keep this patch since this was suggested on a past PR.


app/workers/application_worker.rb
```diff
@@ -0,0 +1,6 @@
+class ApplicationWorker
+  include Sidekiq::Worker
+  sidekiq_options queue: "low_priority"
```
sethherr: Good choice of a default queue!

PR #1061, opened 2019-07-19
Styling and rendering public images for admin blogs

Robertgervais:


Robertgervais:
UNADJUSTEDNONRAW_thumb_6f

app/javascript/packs/pages/admin/blogs.js
```diff
@@ -65,24 +66,87 @@ function BinxAdminBlogs() {
...
+      uppy.on('upload-success', (file, response) => {
+        this.appendPublicImage(response.body.public_image)
+      })
+    },
+
+    publicImageDelete() {
+      $('ul#public_images').on('click', ".image-delete-button", function(e) {
+        e.preventDefault();
+        const id = $(".image-delete-button").closest(".row").find("input").val()
+        console.log(id)
```
sethherr: Oop, needs to be removed

app/controllers/public_images_controller.rb
```diff
@@ -56,7 +56,11 @@ def destroy
     if params[:page].present?
       redirect_to edit_bike_url(imageable_id, page: params[:page]) and return
     elsif imageable_type == "Blog"
-      redirect_to edit_admin_news_url(@imageable.title_slug) and return
+      if request.xhr?
+        render json: {success: true} and return
```
sethherr: Why is this returning something different from the json returned above? Why check for xhr instead of json format?

Robertgervais: mmm, because before I added this my page was, for some reason trying to hit a weird route. It would delete the image and then instead of redirecting the way I wanted it to, it would redirect to DELETE `admin/news/"slug"`. So instead of hitting the regular redirect I just have it removed on the front with JS if it was JUST added.

app/javascript/packs/pages/admin/blogs.js
```diff
@@ -65,24 +66,84 @@ function BinxAdminBlogs() {
         submitOnSuccess: false,
         triggerUploadOnSubmit: false
       })
+      uppy.on('upload-success', (file, response) => {
+        this.publicImageTemplate(response.body.public_image)
```
sethherr: nit: I'd like the appending to the ul to happen here, and `publicImageTemplate` just return the string to render.

app/javascript/stylesheets/_admin.scss
```diff
@@ -176,3 +176,19 @@
...
   }
 }
+
+.admin-public-image {
+  img {
+    max-width: 100%;
+    width: 85px;
+    height: auto;
+  }
+  max-height: 165px;
```
sethherr: Should this be on the `img` tag?

Robertgervais: Hey, this actually is for the public image card. I was trying to shrink them a little so that they took up a lil less space. If you don't care I can just remove it, but it's supposed to be there.

sethherr: I generally think that setting fixed heights on things is problematic.

sethherr: Ahhh. I understand now. It is already set. I don't know what to do here, we can figure it out later

PR #1065, opened 2019-07-23
Fix missing interpolation argument in bikes/show

jmromer:
Fixes an exception from a missing interpolation key.

```yml
# config/locales/en.yml L1283-1285 (0e96aa29)

please_email_support_html: >-
  Please email %{support_email_link} if you don't understand why you're seeing
  this message or if you need an exception made.
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/0e96aa29/config/locales/en.yml#L1283-L1285)]</sup>

Resolves https://app.honeybadger.io/projects/35931/faults/52442699

Also fixes a NoMethodError from `.manufacturer` returning `nil':

https://app.honeybadger.io/projects/35931/faults/52395156

### DIFF
--- a/app/views/bikes/show.html.haml
+++ b/app/views/bikes/show.html.haml
@@ -84,7 +84,8 @@
 
           %li
             %span.attr-title #{t(".manufacturer")}:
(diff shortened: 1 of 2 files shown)


jmromer:
Nope! Thanks for the review :)

PR #1071, opened 2019-07-24
Upgrade fog

sethherr:
- Upgrade `fog-aws` (unblocking nokogiri upgrade for Rails 5 upgrade)
- Add an `ApplicationUploader` for all the other uploaders to inherit from (cleaning up some of the carrierwave stuff)
- remove legacy utf magic comments because Ruby 2 made them unnecessary

### DIFF
--- a/config/initializers/carrierwave.rb
+++ b/config/initializers/carrierwave.rb
@@ -37,7 +37,7 @@ CarrierWave.configure do |config|
 
   if Rails.env.production?
     config.fog_provider "fog/aws"
-    config.asset_host = "https://files.bikeindex.org" if Rails.env.production?
+    config.asset_host = "https://files.bikeindex.org"
     config.fog_credentials = {
       provider: "AWS",
       aws_access_key_id: ENV["S3_ACCESS_KEY"],
--- a/config/initializers/carrierwave.rb
+++ b/config/initializers/carrierwave.rb
@@ -37,7 +37,7 @@ CarrierWave.configure do |config|
 
   if Rails.env.production?
     config.fog_provider "fog/aws"
-    config.asset_host = ENV["ASSET_HOST"]
+    config.asset_host = ENV["S3_ASSET_HOST"]
     config.fog_credentials = {
       provider: "AWS",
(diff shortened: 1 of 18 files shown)


config/initializers/carrierwave.rb
```diff
@@ -47,3 +30,24 @@ def strip
...
   end
 end
+
+# Additional carrierwave configurations
+CarrierWave.configure do |config|
+  config.cache_dir = "#{Rails.root}/tmp/uploads"
+
+  if Rails.env.production?
+    config.fog_provider "fog/aws"
+    config.asset_host = "https://files.bikeindex.org" if Rails.env.production?
```
jmromer: Any objection to moving the asset host string to an env var? 

Also, branching more granularly would be a little more legible here (e.g. the `if Rails.env.production?` is duplicated on L40)

sethherr: What would be the purpose of putting it in an env var? I don't think it's private information, and we don't need to modify it, so I don't know why it would be useful to do - if it does make something better, let's definitely do it!

w/r/t branching more granularly - I've fixed the duplicated `if Rails.env.production?` - what are you suggesting otherwise?

jmromer: Sorry, brain fart — I meant to say *less* granularly, e.g. 

```ruby
if Rails.env.production?
  CarrierWave.configure # ...
else
  CarrierWave.configure # ...
end
```

instead of branching within the `configure` block for each config method. Minor quibble though. Possibly I'm hyper-sensitive to granular conditional branching from having to de-tangle a lot of those to externalize strings from templates recently :)

Re. storing configuration in the environment --

That's one of the 12-factor principles. I don't think it's critical here, I bring it up more from a "broken windows" perspective:

- https://12factor.net/config
- https://github.com/search?q=ENV%5B%22ASSET_HOST%22%5D&type=Code

Also non-blocking though 👍

sethherr: I understand your sensitivity! I think the reduction in duplication is worth it here, but I may eat my words some day.

w/r/t storing config - good point. Extracted

PR #1072, opened 2019-07-24
Fix minor test warnings

sethherr:
Warnings visible from a recent circle run (e.g. [the update_thor ci](https://circleci.com/gh/bikeindex/bike_index/5391))

- [x] Duplicated link key
- [x] `axlsx` - `constant ::Fixnum is deprecated`. Issue discussed https://github.com/randym/axlsx/issues/522 (now using using pre release version, since that's the only thing [released in the past few years](https://rubygems.org/gems/axlsx))
- [x] `app/views/organized_mailer/organization_invitation.text.haml:1: warning: found = in conditional, should be ==`
- [x] `DEPRECATION WARNING: Calling URL helpers with string keys controller, action is deprecated. Use symbols instead. (called from block in _app_views_admin_graphs_index_html_haml___2475475577798723556_70079177928800 at /home/circleci/bikeindex/bike_index/app/views/admin/graphs/index.html.haml:43)`

### DIFF
--- a/app/views/organized_mailer/organization_invitation.text.haml
+++ b/app/views/organized_mailer/organization_invitation.text.haml
@@ -1,4 +1,4 @@
-- if @new_user = true
+- if @new_user == true
   = t(".text.hello_welcome_to_bike_index")
 - else
   = t(".text.hello")
--- a/Gemfile
+++ b/Gemfile
@@ -66,7 +66,7 @@ gem "unicorn" # Use Puma as the app server
 gem "unicorn-worker-killer"
 
 # Making other files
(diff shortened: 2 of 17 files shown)


app/views/organized_mailer/organization_invitation.text.haml
```diff
@@ -1,4 +1,4 @@
-- if @new_user = true
+- if @new_user == true
```
jmromer: Yikes, thanks for catching this 🙏

sethherr: I can't take credit. It's all from the spec output!

Now that we have less garbage in there, I'm going to check in there periodically, since there are useful things 🙂

PR #1073, opened 2019-07-24
Include sync in prepare_translations task

sethherr:
Include `translation:sync` in `bin/rake prepare_translations` task.

I think this is useful because running `translation:sync` radically alters the normalized output (since it overrides any changes the normalization makes to non-english translations) - so I think it should be run any time you run the prepare_translation task. Is there a time when we wouldn't want prepare_translations to include sync (provided the API key is available?)

### DIFF
--- a/lib/tasks/rake_tasks.rake
+++ b/lib/tasks/rake_tasks.rake
@@ -25,4 +25,8 @@ task prepare_translations: :environment do
   i18n_tasks.start(["add-missing", "-v", "'TRME %{value}'"])
   i18n_tasks.start(["normalize"])
   i18n_tasks.start(["health"])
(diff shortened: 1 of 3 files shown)


lib/tasks/rake_tasks.rake
```diff
@@ -22,7 +22,10 @@ desc "Prepare translations for committing to master"
 task prepare_translations: :environment do
   require "i18n/tasks/cli"
   i18n_tasks = I18n::Tasks::CLI.new
-  i18n_tasks.start(["add-missing", "-v", "'TRME %{value}'"])
```
sethherr: I think this isn't necessary, since syncing deletes any changes?

PR #1075, opened 2019-07-24
Admin custom layouts revision

Robertgervais:


app/views/admin/organizations/custom_layouts/_public_image.html.haml
```diff
@@ -0,0 +1,18 @@
+.card.bg-light.admin-public-image
```
sethherr: Is there a reason this needs to be a different template than the blogs template?

If the only reason is that this shouldn't include the delete button, can you just hide the delete button via css?

app/javascript/packs/pages/admin/custom_layouts.js
```diff
@@ -0,0 +1,97 @@
+import log from "../../utils/log";
```
sethherr: What can we do to make a `BinxAdminImageUploader` class which is shared by blogs and custom layouts, so that we don't need all this nearly identical code?

app/views/admin/organizations/custom_layouts/_public_image.html.haml
Robertgervais: I didn't really want to link this to public_image template in news, since it felt sort of weird. With `public` it was a repo where a lot of things looked, so that made sense. What if I made a public_image_revised and added it to the public folder? Would that work? It would be really easy to make it the same template, I just want to put it somewhere that makes sense.

app/javascript/packs/pages/admin/custom_layouts.js
Robertgervais: I can make this happen. Would you rather I used some conditionals for the template, or just use css to hide some of the things I'm not including in this one?

sethherr: I think css will be better, because it's simpler and will keep the js cleaner.

app/views/admin/organizations/custom_layouts/_public_image.html.haml
sethherr: What about `_admin_public_image`? I think both the blog and this should reference the same partial.

Robertgervais: Yeah, I was thinking about it and this was pretty much what I came to, also. This is what I'll go with!

PR #1077, opened 2019-07-25
Improve bike authorize_and_claim_for_user naming

sethherr:
Rename `Bike` method `authorize_for_user!` to `authorize_and_claim_for_user`, to follow Ruby conventions.

Reasoning: I just started to write another method in bike, and was about to make it a bang method, since it also modifies state - however, I feel like bang methods should raise errors if the fail, since that seems to be a convention.

historical context: I _think_ I originally made `authorize_for_user!` a bang method to distinguish it from the similarly named `authorized_for_user?` - but I think renaming does a better job explaining what is going on and distinguishing it.

### DIFF
--- a/app/controllers/api/v2/bikes.rb
+++ b/app/controllers/api/v2/bikes.rb
@@ -81,7 +81,7 @@ module API
         end
 
         def authorize_bike_for_user(addendum = "")
-          return true if @bike.authorize_for_user!(current_user)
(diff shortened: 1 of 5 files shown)


sethherr:
Haha, AKA, it's ruby, there aren't any rules!

I think that provides another heuristic  - bang methods should only exist if there is a non-bang method.

PR #1078, opened 2019-07-25
Add impounding functionality

sethherr:
Enable organizations to mark bikes impounded, for bike depots (#1068)


sethherr:
Reference used for custom model validation errors with localization (i18n): https://stackoverflow.com/questions/17935597/ruby-on-rails-i18n-want-to-translate-custom-messages-in-models

app/models/impound_record.rb
```diff
@@ -0,0 +1,27 @@
+class ImpoundRecord < ActiveRecord::Base
+  belongs_to :bike
+  belongs_to :user
+  belongs_to :organization
+
+  validates_presence_of :bike_id, :user_id
+  validates_uniqueness_of :bike_id, if: :current?, conditions: -> { current }
```
jmromer: I notice a lot of places in the code seem to prefer accessing / mutating the foreign key rather than the associated record. Is there a reason for that? 

For the presence validation here, validating `:bike` and `:user` would ensure the presence of the associated record

db/migrate/20190725141309_create_impound_records.rb
```diff
@@ -0,0 +1,12 @@
+class CreateImpoundRecords < ActiveRecord::Migration
+  def change
+    create_table :impound_records do |t|
+      t.references :bike
+      t.references :user
+      t.references :organization
```
jmromer: No foreign key indexes?

sethherr: Adding. Good call

app/models/impound_record.rb
sethherr: The reasons are because stuff sometimes ends up deleted, and sometimes stubbing things in tests.

I don't have strong feelings about this, it's just what I default to.

PR #1079, opened 2019-07-25
Automate Bike Code creation

sethherr:
I made a mistake adding codes manually - so I decided it was time to add some more automation to bike code creation.

Now if you have a `BikeCodeBatch`, you can create the next bike_codes by specifying how many you want to create.

```ruby
BikeCodeBatch.find(9).create_codes(500)
```

n.b. BikeCodes are actually stickers, see #1080

### DIFF
--- a/app/models/bike_code_batch.rb
+++ b/app/models/bike_code_batch.rb
@@ -2,4 +2,25 @@ class BikeCodeBatch < ActiveRecord::Base
   belongs_to :user # Creator of the batch
   belongs_to :organization
   has_many :bike_codes
+
+  def max_code_integer; bike_codes.maximum(:code_integer) || 0 end
+
+  def create_codes(number_to_create, initial_code_integer: nil, kind: "sticker")
+    raise "Prefix required to create sequential codes!" unless prefix.present?
+    initial_code_integer ||= max_code_integer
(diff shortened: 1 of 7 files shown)


app/models/bike_code_batch.rb
```diff
@@ -2,4 +2,28 @@ class BikeCodeBatch < ActiveRecord::Base
...
+    initial_code_integer ||= max_code_integer
+    initial_code_integer += 1 if bike_codes.where(code_integer: initial_code_integer).present?
+    clength = code_number_length_or_default # Assign so it isn't recalculated in loop
+    number_to_create.times do |i|
+      code_integer_with_padding = (i + initial_code_integer).to_s.rjust(clength, "0")
+      bike_codes.create!(organization: organization,
+                         kind: kind,
+                         code: prefix + code_integer_with_padding)
+    end
+    update_attributes(updated_at: Time.current) # Bump
```
jmromer: nit: [`#touch`](https://apidock.com/rails/ActiveRecord/Persistence/touch) also does this

sethherr: Good suggestion!

Note: (as per your reference link) `touch` only runs these callbacks:

>  the `after_touch`, `after_commit` and `after_rollback`

Which means that manually calling `update_attributes(updated_at: Time.current)` is important in many places, for triggering before_save callbacks.

PR #1082, opened 2019-07-25
Add additional org child functionality

sethherr:


app/controllers/organized/base_controller.rb
```diff
@@ -6,14 +6,6 @@ class BaseController < ApplicationController
 
     layout "application_revised"
 
-    def root
```
sethherr: I think it makes sense to move this out of base_controller, because all the other organized controllers inherit from this controller, and we probably don't want them all to have a `root` method?

app/models/organization.rb
```diff
@@ -270,14 +273,17 @@ def suspended?
...
-    end while self.class.exists?(access_token: access_token)
+  def update_associations
+    parent_organization&.update_attributes(updated_at: Time.now)
   end
 
   private
 
+  def calculated_paid_feature_slugs
+    slugs = current_invoices.feature_slugs
+    child_ids.any? ? slugs + ["child_organizations"] : slugs
```
sethherr: Doing this because we're going to be checking if they have child organizations on every organized page render, and I'm concerned about performance.

PR #1085, opened 2019-07-25
Switch base controller over to new_admin layout

Robertgervais:


Robertgervais:
I tried changing "new_admin" to "admin" but this new i18n stuff raised all kind of errors about a missing template, so i'm leaving it `new_admin` for now. I erased all traces of `admin`. I'm not entirely sure what you meant about any references to the admin template, but I looked around for a while and couldn't see anything else having to do with it.

PR #1091, opened 2019-07-26
Add StolenRecord#alert_image, AlertImageUploader

jmromer:
For bikes with a current stolen record, makes the alert image the opengraph page image
```rb
# app/helpers/header_tag_helper.rb L104-105 (98a75892)

if @bike.current_stolen_record.present?
  self.page_image = @bike.alert_image_url
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/98a75892/app/helpers/header_tag_helper.rb#L104-L105)]</sup>

`Bike#alert_image_url` generates alert image if it doesn't already exist

```rb
# app/models/bike.rb L516-520 (4b56ebff)

def alert_image_url
  return if current_stolen_record.blank? || public_images.none?
  current_stolen_record.generate_alert_image
  current_stolen_record.alert_image_url
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/4b56ebff/app/models/bike.rb#L516-L520)]</sup>

`StolenRecord#alert_image` processes in the background:
```rb
# app/models/stolen_record.rb L53-55 (4b56ebff)

mount_uploader :alert_image, AlertImageUploader
process_in_background :alert_image
after_commit :remove_outdated_alert_image
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/4b56ebff/app/models/stolen_record.rb#L53-L55)]</sup>

```rb
# app/uploaders/alert_image_uploader.rb L49-56 (8511c454)

def generate_alert_image
  AlertImageGenerator.generate_image(
    bike_image_path: bike_image.path,
    bike_url: bike_url,
    bike_location: bike_location,
    output_path: current_path,
  )
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/8511c454/app/uploaders/alert_image_uploader.rb#L49-L56)]</sup>

```rb
# app/services/alert_image_generator.rb L1-8 (a51dab5e)

module AlertImageGenerator
  BASE_TEMPLATE_PATH = Rails.root.join("app", "assets", "images", "premium_alert_template.png")
  HEADER_HEIGHT = 100
  FOOTER_HEIGHT = 50
  PADDING = 200

  def self.generate_image(bike_image_path:, bike_url:, bike_location:, output_path:)
    template = MiniMagick::Image.open(BASE_TEMPLATE_PATH)
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/a51dab5e/app/services/alert_image_generator.rb#L1-L8)]</sup>


app/models/bike.rb
```diff
@@ -497,6 +497,27 @@ def external_image_urls
...
+    generate_alert_image
+    current_stolen_record.alert_image_url
+  end
+
+  def generate_alert_image
+    return if current_stolen_record.blank?
+    return if current_stolen_record.alert_image.present?
+
+    # Remove any old alert images
+    stolen_records.each { |sr| sr.alert_image&.remove! }
```
sethherr: Wouldn't we want to trigger this when `current_stolen_record.blank?` too?

sethherr: Or maybe, I don't know that this fits will trigger all the times we want it to trigger. Would it make more sense on an after commit hook on stolen record?

config/initializers/carrierwave.rb
```diff
@@ -33,7 +33,11 @@ def strip
 
 # Additional carrierwave configurations
 CarrierWave.configure do |config|
-  config.cache_dir = "#{Rails.root}/tmp/uploads"
+  config.cache_dir = Rails.root.join("tmp", "uploads")
+
+  if Rails.env.test?
+    config.enable_processing = false
```
sethherr: Oooh! Does this make our tests faster in general? Nice addition.

spec/factories/stolen_records.rb
```diff
@@ -1,11 +1,17 @@
 FactoryBot.define do
   factory :stolen_record do
-    bike { FactoryBot.create(:bike) }
+    bike { FactoryBot.create(:bike, :with_image) }
```
sethherr: Does this mean that every bike factory now adds an image by default? If so, why do that rather than having it be a separate factory? Does this have a negative impact on test speed?

spec/factories/bikes.rb
```diff
@@ -14,6 +14,14 @@
     cycle_type { CycleType.slugs.first }
     propulsion_type { "foot-pedal" }
 
+    trait :with_image do
```
sethherr: Does this mean that every bike factory now adds an image by default? If so, why do that rather than having it be a separate factory? Does this have a negative impact on test speed?

app/uploaders/alert_image_uploader.rb
```diff
@@ -0,0 +1,57 @@
...
+    raise ArgumentError, "No bike image for Bike #{bike.id}" if public_image.blank?
+    public_image
+  end
+
+  def bike_url
+    "bikeindex.org/bikes/#{bike.id}"
+  end
+
+  def bike_location
+    city = bike.registration_address[:city]&.titleize
```
sethherr: I think this should grab the current stolen record theft location, and fallback to the registration address.

app/models/bike.rb
```diff
@@ -497,6 +497,27 @@ def external_image_urls
...
+
+    return if public_images.none?
+
+    # Generate alert image for current stolen record
+    current_stolen_record.alert_image = public_images.first.image
```
sethherr: Why is this defined here, but also in the uploader?

app/uploaders/alert_image_uploader.rb
```diff
@@ -0,0 +1,57 @@
...
+    elsif state
+      state
+    else
+      ""
+    end
+  end
+
+  def generate_alert_image
+    AlertImageGenerator.generate_image(
+      bike_image_path: bike_image.path,
```
sethherr: I _believe_ this is going to be problematic with images on s3. 

Ideally this with work correctly when locally (like this), but also work correctly for images that are stored on s3 (which still return a relative path, but the path `"uploads/Pu/229/IMG_3624.JPG"` doesn't correspond to a local file.

app/services/alert_image_generator.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+
+    alert_image = template.composite(bike) do |c|
+      c.gravity "Center"
+      c.compose "Over"
+      c.geometry "+0+#{HEADER_HEIGHT - FOOTER_HEIGHT}"
+    end
+
+    alert_image.combine_options do |i|
+      i.font "ArialI"
```
sethherr: is this a typo?

app/services/alert_image_generator.rb
```diff
@@ -0,0 +1,46 @@
+module AlertImageGenerator
+  BASE_TEMPLATE_PATH = Rails.root.join("app", "assets", "images", "premium_alert_template.png")
```
sethherr: Could we name this "promoted_alert_template" to go with our new name?

Also, does it matter that we're translating this to a jpg? Would it be faster/better to make this a jpg to begin with?

app/models/bike.rb
```diff
@@ -497,6 +497,27 @@ def external_image_urls
...
+  def alert_image_url
+    return if current_stolen_record.blank?
+
+    generate_alert_image
+    current_stolen_record.alert_image_url
+  end
+
+  def generate_alert_image
+    return if current_stolen_record.blank?
+    return if current_stolen_record.alert_image.present?
```
sethherr: How do we trigger regenerating if there is a new public image first?

spec/factories/bikes.rb
jmromer: > Does this mean that every bike factory now adds an image by default?

Not by default, only if we pass the trait to the factory -- so `Factory.create(:bike, :with_image)`  will create an associated image, but `Factory.create(:bike)` won't.

app/services/alert_image_generator.rb
jmromer: Nope, that's the italic version of the font:

```
% convert -list font | grep --before=5 'Arial Italic'
  Font: ArialI
    family: Arial
    style: Undefined
    stretch: Undefined
    weight: 0
    glyphs: /Library/Fonts//Arial Italic.ttf
```

spec/factories/bikes.rb
sethherr: Sweet addition. Cool new thing!

app/models/bike.rb
jmromer: Assigning the image to `alert_image` and saving here triggers the uploader to run.
Having some trouble parsing the question though, let me know if I'm missing something.

app/models/bike.rb
jmromer: > Wouldn't we want to also trigger this when current_stolen_record.blank?

I would think not? If the given bike isn't currently stolen, then it shouldn't have an alert image (the template says "stolen bike alert"). I might be imagining the invocation sequence differently, though.

> Would it make more sense on an after commit hook on stolen record?

Yeah, I think so... I had some concern about triggering the callback excessively in an after_commit hook, but it'll also reduce storage requirements.

With an `after_create`, a newly created stolen record will trigger the creation of an alert image (if there's a bike image available). When a bike is recovered, the alert image sticks around until a new one is needed.

With an `after_commit`, the outdated alert image is removed when the bike is recovered.

app/uploaders/alert_image_uploader.rb
jmromer: Hmm thanks — I should be able to test this locally with the prod carrierwave settings 👍

config/initializers/carrierwave.rb
jmromer: Yup! Not a huge difference in our case, but it's the approach suggested by the carrierwave docs.

app/models/bike.rb
sethherr: I don't know that the creation needs to be in an `after_` hook - I think the triggering based on the call to the alert image url makes sense and is slick.

I was just thinking specifically about the deletion. e.g. anytime the record is switched to be recovered, the alert image should deleted. Rather then delete method here, where it seems like it won't consistently delete the image.

app/models/bike.rb
sethherr: This is mostly a style nit, not a serious concern.

`def bike_image` in `AlertImageUploader` actually chooses what image is processed. But because this assigns to `stolen_record.alert_image`, I would assume this method in `bike.rb` is what chooses the image to process. 

This is a response to this imaginary scenario: I come back to this code and want to update which image is processed - so I change it here, and am confused about why it doesn't change which image ends up being processed.

This is a nit you can ignore, since the above isn't actually going to happen.

jmromer: Ah thanks for clarifying, that's a good call. `AlertImageUploader#bike_image` is vestigial, I'll get rid of it 👍

app/services/alert_image_generator.rb
jmromer: My hunch is that intermediate conversion might degrade the quality without any benefit. The final product is saved as a JPG to reduce the file save though.

app/models/bike.rb
jmromer: Updated -- moved image removal to an `after_commit` hook
```rb
# app/models/bike.rb L530-534 (55c09ed7)

def alert_image_url
  return if current_stolen_record.blank? || public_images.none?
  current_stolen_record.generate_alert_image
  current_stolen_record.alert_image_url
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/55c09ed7/app/models/bike.rb#L530-L534)]</sup>


```rb
# app/models/stolen_record.rb L53-55 (a0dc6a32)

mount_uploader :alert_image, AlertImageUploader
process_in_background :alert_image
after_commit :remove_outdated_alert_image
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/a0dc6a32/app/models/stolen_record.rb#L53-L55)]</sup>

```rb
# app/models/stolen_record.rb L283-286 (a0dc6a32)

# If the bike has been recovered, remove the alert_image
def remove_outdated_alert_image
  alert_image.remove! if date_recovered.present?
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/a0dc6a32/app/models/stolen_record.rb#L283-L286)]</sup>

app/uploaders/alert_image_uploader.rb
jmromer: Updated to prefer the stolen record's location:

```rb
# app/uploaders/alert_image_uploader.rb L33-39 (a0dc6a32)

def bike_location
  if stolen_record.location.present?
    stolen_record.location
  else
    bike.registration_location
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/a0dc6a32/app/uploaders/alert_image_uploader.rb#L33-L39)]</sup>

app/uploaders/alert_image_uploader.rb
jmromer: Worked around this by setting asset host in environment config (development.rb / test.rb / production.rb) and using that value for carrierwave:

```rb
# config/environments/development.rb L57-58 (c3eee6a6)

# Enable serving of images, stylesheets, and JavaScripts from an asset server.
config.action_controller.asset_host = ENV["BASE_URL"]
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/c3eee6a6/config/environments/development.rb#L57-L58)]</sup>


```rb
# config/environments/production.rb L67-68 (c3eee6a6)

# Enable serving of images, stylesheets, and JavaScripts from an asset server.
config.action_controller.asset_host = ENV["S3_ASSET_HOST"]
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/c3eee6a6/config/environments/production.rb#L67-L68)]</sup>


```rb
# config/initializers/carrierwave.rb L36 (1a242e11)

config.asset_host = ActionController::Base.asset_host
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/1a242e11/config/initializers/carrierwave.rb#L36-L36)]</sup>

```rb
bike.alert_image_url
# "http://localhost:3001/uploads/St/72666/flying_pigeon-alert.jpg"
```

app/models/bike.rb
jmromer: `generate_alert_image` now uses the most recently added bike image and re-generates the alert image only if the bike image has changed (using the filename equivalence as a heuristic)

```rb
# app/models/bike.rb L530-534 (55c09ed7)

def alert_image_url
  return if current_stolen_record.blank? || public_images.none?
  current_stolen_record.generate_alert_image
  current_stolen_record.alert_image_url
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/55c09ed7/app/models/bike.rb#L530-L534)]</sup>

```rb
# app/models/stolen_record.rb L262-281 (a0dc6a32)

# Generate the "premium alert image"
# (the most recently created bike image placed on a branded template)
#
# The URL is available immediately, processing is performed in the background.
# If processing fails, the unprocessed image will be loaded.
#
# TODO: Allow selection of which bike image to use for the alert image
def generate_alert_image
  new_image = bike.reload.public_images.order(:created_at).last&.image
  return if new_image.blank?

  if alert_image.present?
    new_file = File.basename(new_image.path, ".*")
    cur_file = File.basename(alert_image.path, ".*")
    return if "#{new_file}-alert" == cur_file
  end

  alert_image.remove!
  update(alert_image: new_image)
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/a0dc6a32/app/models/stolen_record.rb#L262-L281)]</sup>

app/models/bike.rb
```diff
@@ -505,6 +505,20 @@ def registration_address # Goes along with organization additional_registration_
     end
   end
 
+  def registration_location
+    address = registration_address.with_indifferent_access
+    city = address[:city]&.titleize
+    state = address[:state]&.upcase
+
+    if city && state
+      "#{city}, #{state}"
```
sethherr: Nit: you could just do `[city, state].reject(&:blank?).join(", ")`

app/models/stolen_record.rb
```diff
@@ -241,4 +245,43 @@ def find_or_create_recovery_link_token
     update_attributes(recovery_link_token: SecurityTokenizer.new_token)
     recovery_link_token
   end
+
+  def location
```
sethherr: There are already a number of address functions in stolen record, which are necessary for legacy reasons.

Is there a way to make this either use one of them, or consolidate them in some other way? Or unify it with the location method in bike? AKA I think you could do a little refactoring to make some of the address grossness nicer than anything I've done 🙂

app/models/bike.rb
jmromer: If city is blank that'll produce `", NY"`, e.g.

jmromer: Thanks! Updated in f6e18823

app/models/stolen_record.rb
jmromer: Renamed in 1bb09e6d to follow suit with the existing methods, but I'm not seeing a way to build on those without munging their output, which seems like needless coupling

sethherr: Fair. All I could ask 😉

PR #1097, opened 2019-07-29
Sync translations

jmromer:
Adds changes from `bin/rake translations:sync` on master

### DIFF
--- a/config/locales/.translation_io
+++ b/config/locales/.translation_io
@@ -1,2 +1,2 @@
 ---
-timestamp: 1564167198
+timestamp: 1564415844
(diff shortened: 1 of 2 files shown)


jmromer:
Brainstorming: Run the sync task on CI periodically, if it makes changes, open a PR?

sethherr:
I feel like it would be ideal to run it before deploying to production, every time - and if it creates changes, halt the deployment and open a PR with the changes. But that might be too much hassle?

jmromer:
Added [here](https://github.com/bikeindex/bike_index/pull/1100) -- elevated perms are required to add a read-write key and automate the PR creation, though.

PR #1102, opened 2019-07-31
Fix prod assets host config

jmromer:
#1091 Breaks asset loading -- the changes to the app asset_host were overkill. This patch reverts those changes and instead sets `config.asset_host = ENV["BASE_URL"]` for CarrierWave specifically.

### DIFF
--- a/app/services/stolen_record_updator.rb
+++ b/app/services/stolen_record_updator.rb
@@ -17,8 +17,7 @@ class StolenRecordUpdator
         s.save
       end
     end
-    @bike.reload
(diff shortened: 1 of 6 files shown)


jmromer:
> I can still see images on Bike Index

Good! I reverted immediately :) 

It looks like some assets load from `bikeindex.org`, others from `files.bikeindex.org`. The change turned out to be unnecessary in the first place because the CarrierWave image url methods prepend the asset host set in the CarrierWave config. The reason it wasn't showing up locally is that it wasn't set for development / test environments.

PR #1124, opened 2019-08-01
Fix premature image removal bug

jmromer:
Fixes a bug whereby alert images are removed prematurely because a bike's `current_stolen record` has `date_recovered` set.

```rb
irb(main):006:0> bike.current_stolen_record.date_recovered
=> Wed, 31 Jul 2019 20:00:00 CDT -05:00

irb(main):007:0> bike.current_stolen_record.current
=> true

irb(main):008:0> bike.stolen?
=> true

irb(main):009:0> bike.recovered?
=> false
```

Also fixes filename generation for alert images, which currently doesn't correctly handle bike images with hyphens in the name.

Resolves https://app.honeybadger.io/projects/35931/faults/52782138

### DIFF
--- a/app/models/stolen_record.rb
+++ b/app/models/stolen_record.rb
@@ -277,6 +277,6 @@ class StolenRecord < ActiveRecord::Base
 
   # If the bike has been recovered, remove the alert_image
   def remove_outdated_alert_image
-    alert_image.remove! if date_recovered.present?
+    alert_image.remove! unless bike.reload.stolen?
   end
 end
--- a/app/uploaders/alert_image_uploader.rb
+++ b/app/uploaders/alert_image_uploader.rb
@@ -12,7 +12,7 @@ class AlertImageUploader < ApplicationUploader
 
   def filename
     return if stolen_record.alert_image.blank?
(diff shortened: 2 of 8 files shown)


app/models/stolen_record.rb
```diff
@@ -277,6 +277,6 @@ def generate_alert_image
 
   # If the bike has been recovered, remove the alert_image
   def remove_outdated_alert_image
-    alert_image.remove! if date_recovered.present?
+    alert_image.remove! unless bike.reload.stolen?
```
sethherr: Why is `date_recovered` present?

jmromer: unclear. I'm looking through all the places in the code where it's set. Doesn't seem like a valid state to me.

sethherr: I agree, it doesn’t seem valid to me either. Weird that it isn’t causing other issues! Glad we’re catching it here I guess.

jmromer: There's only one, at least:
```rb
irb(main):012:0> StolenRecord.unscoped.where(current: true).where.not(date_recovered: nil).count
=> 1

```

PR #1129, opened 2019-08-02
Add landscape-oriented premium alert image

jmromer:
- Implements versions for alert image, with square and landscape
- Updates header tag helper to use landscape alert image version for stolen bikes
- Updates factories to allow setting image filenames

Screen Shot 2019-08-05 at 10 59 39 AM

Possible issues remaining:
- Cloudfront (?) and/or browser caching may be an issue when trying to regenerate images
- This [error](https://app.honeybadger.io/projects/35931/faults/52782138) seems to happen when processing begins but the base image isn't available yet, but I'm not sure about that yet — I want to see what impact the changes here have


sethherr:
_Hell_ yeah!

app/controllers/admin/stolen_bikes_controller.rb
```diff
@@ -21,8 +21,8 @@ def approve
   end
 
   def regenerate_alert_image
-    @bike.current_stolen_record.generate_alert_image!
-    redirect_to edit_admin_stolen_bike_url(@bike), notice: "Premium alert images regenerated."
+    @bike.current_stolen_record.generate_alert_image
+    redirect_to edit_admin_stolen_bike_url(@bike), notice: "Promoted alert images regenerated."
```
sethherr: Good catch

config/initializers/carrierwave.rb
```diff
@@ -34,7 +34,7 @@ def strip
 # Additional carrierwave configurations
 CarrierWave.configure do |config|
   config.cache_dir = Rails.root.join("tmp", "uploads")
-  config.storage :file
+  config.storage = :file
```
sethherr: Ha, what? And we don't need to put `storage :fog` in production?

spec/uploaders/alert_image_uploader_spec.rb
```diff
@@ -3,81 +3,4 @@
 
 RSpec.describe AlertImageUploader do
   include CarrierWave::Test::Matchers
```
sethherr: Would it make sense to delete this file? Or do you think you'll add to it in the future?

app/models/stolen_record.rb
```diff
@@ -259,36 +259,28 @@ def bike_main_image
...
   end
 
-  def generate_alert_image!(bike_image: bike_main_image)
-    return if bike_image.blank?
-
+  def remove_alert_image
     alert_image.remove!
-    self.alert_image = bike_image
-    save
+    self.alert_image = nil
```
sethherr: This doesn't happen automatically when you do `alert_image.remove!`?

app/uploaders/alert_image_uploader.rb
```diff
@@ -2,6 +2,8 @@ class AlertImageUploader < ApplicationUploader
   include CarrierWave::MiniMagick
   include ::CarrierWave::Backgrounder::Delay
 
+  alias stolen_record model
```
sethherr: I feel like `alias_method :stolen_record, :model` reads a little bit better. Open to different opinion though

app/uploaders/alert_image_uploader.rb
```diff
@@ -10,40 +12,47 @@ def base_store_dir
...
+    process resize_to_fill: [1200, 1200]
   end
 
-  def bike_location
-    if stolen_record.address_location.present?
-      stolen_record.address_location
-    else
-      bike.registration_location
+  def generate_landscape
+    Rails.logger.info "Processing landscape #{current_path}"
```
sethherr: I'm somewhat suspicious of logging statements, because it's difficult to know when it's ok to delete them.

I understand the utility of these, because this is background processed and otherwise it might be difficult to get visibility - but - would you create a separate PR once this is merged, that removes them? So we know it's something that should happen and it's a streamlined process

config/initializers/carrierwave.rb
jmromer: The prod value was being set twice:
https://github.com/bikeindex/bike_index/blob/cde3178651a00100803ebd97ed857f2e3e3fc097/config/initializers/carrierwave.rb#L35-L43

app/controllers/admin/stolen_bikes_controller.rb
sethherr: Actually - there are still a number of places that _"premium"_ is being used in this code. Are you willing to replace all of them? Could be a separate PR.

app/uploaders/alert_image_uploader.rb
jmromer: Agreed, but I think the semantic difference dominates for me — `alias` is lexically scoped, which is more restrictive / clearer in this case. (It was the style guide that switched my default on this... that said, it's a weakly set default):

> Prefer alias when aliasing methods in lexical class scope as the resolution of self in this context is also lexical, and it communicates clearly to the user that the indirection of your alias will not be altered at runtime or by any subclass unless made explicit.

> https://github.com/rubocop-hq/ruby-style-guide#alias-method-lexically

app/uploaders/alert_image_uploader.rb
jmromer: No prob -- I won't keep these around too long, just want to get some insight into that empty body exception.

app/controllers/admin/stolen_bikes_controller.rb
jmromer: Good catch thanks! Updated in b8b0fd45

spec/uploaders/alert_image_uploader_spec.rb
jmromer: Removed in d74292d4 -- I'll reintroduce if needed

app/models/stolen_record.rb
jmromer: Good call thanks -- updated in 010f4469

app/uploaders/alert_image_uploader.rb
sethherr: Thanks!

PR #1133, opened 2019-08-05
Add Ownership#owner_email format validation

jmromer:
OwnerEmail can be persisted with an invalid owner_email, leading to exceptions
downstream when triggering `OrganizedMailer.finished_registration`:

https://app.honeybadger.io/projects/35931/faults/52974853

```
 Ownership.find(id).owner_email
 # => "a.b.c"
```

```rb
 Ownership.where("owner_email NOT LIKE '%@%'").count
 # => 113
```

### DIFF
--- a/app/models/ownership.rb
+++ b/app/models/ownership.rb
@@ -4,6 +4,8 @@ class Ownership < ActiveRecord::Base
   validates_presence_of :owner_email
   validates_presence_of :creator_id
   validates_presence_of :bike_id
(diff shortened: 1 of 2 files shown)


jmromer:
> Does devise do this?

It has a default [email validator](https://github.com/plataformatec/devise/wiki/How-to:-Use-a-custom-email-validator-with-Devise) that's intentionally relaxed, from the looks of it.

PR #1134, opened 2019-08-05
Don't attempt to send ownership email unless owner_email is present

jmromer:
Resolves https://app.honeybadger.io/projects/35931/faults/53091091


jmromer:
We have format validation for `owner_email` now with https://github.com/bikeindex/bike_index/pull/1133, but those records without a valid email now have `owner_email` blanked out. 

```rb
Ownership.where(owner_email: nil).count
# 106
```

sethherr:
I feel like this erroring is a good thing then.

PR #1135, opened 2019-08-05
Kill logging in AlertImageUploader

jmromer:
Removes debug logging


sethherr:
What happened here?

jmromer:
> What happened here?

GitHub automatically closes PRs when its head commit is merged elsewhere.

sethherr:
It seems like the logger statements are still in master? So this should still be open?

jmromer:
> It seems like the logger statements are still in master? So this should still be open?

Yes -- didn't realize it remained closed after I re-pushed.

sethherr:
Force pushes scare me 😉

PR #1137, opened 2019-08-05
Show language selection to everyone

sethherr:
First, because the translator wants to see it. But also because why not!

- Setting language automatically leads to weird edge cases, and I don't think it adds much value. I've added a link to the edit account page, if there is a mismatch, which I think makes things better.
- When selecting a language, include the language's name in it's own language

### DIFF
--- a/spec/requests/locale_detection_request_spec.rb
+++ b/spec/requests/locale_detection_request_spec.rb
@@ -1,16 +1,9 @@
 require "rails_helper"
 
 RSpec.describe "Locale detection", type: :request do
-  before do
-    allow(Flipper).to receive(:enabled?).with(:localization, nil).and_return(true)
-  end
-
   describe "requesting the root path" do
     context "given a user preference" do
       include_context :request_spec_logged_in_as_user
-      before do
-        allow(Flipper).to receive(:enabled?).with(:localization, current_user).and_return(true)
-      end
 
       it "renders the homepage in the requested language" do
         get "/"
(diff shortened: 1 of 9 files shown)


jmromer:
Does this break locale consistency on signup?

sethherr:
@jmromer it did, good catch! Should be fixed by 6300a0f496d62a8d40df2b5afdd16fce06c2dc5b

spec/requests/locale_detection_request_spec.rb
```diff
@@ -57,26 +50,29 @@
 
     context "given multiple detected locales" do
       include_context :request_spec_logged_in_as_user
-      before do
-        allow(Flipper).to receive(:enabled?).with(:localization, current_user).and_return(true)
-      end
 
       it "gives highest precedence to query param" do
-        current_user.update(preferred_language: :es)
+        current_user.update_attribute :preferred_language, :es
```
jmromer: A bit early to worry about this, but just a [head's up](https://github.com/rails/rails/pull/31998):

> DEPRECATION WARNING: update_attributes! is deprecated and will be removed from Rails 6.1 (please, use update! instead)

jmromer: Also, `update_attribute` (sans -_s_) bypasses validations, not clear if / why that's intended here.

sethherr: That is intended here. Because `:es` isn't an official language

PR #1145, opened 2019-08-06
Don't raise translation error for unknown currencies

sethherr:


app/helpers/money_helper.rb
```diff
@@ -1,11 +1,17 @@
...
+  PERMITTED_CURRENCIES = %w[USD EUR CAD].freeze
+
   # TODO: Add a bank implementation that fetches conversion rate values
-  Money.default_bank.add_rate(:USD, :EUR, 0.88)
+  Money.default_bank.add_rate(:USD, :EUR, 0.89)
+  Money.default_bank.add_rate(:USD, :CAD, 1.32)
 
   def default_currency
-    t(I18n.locale, scope: [:money, :currencies])
+    currency = t(I18n.locale, scope: [:money, :currencies])
```
jmromer: Aren't we only setting `I18n.locale` to permitted locales?

sethherr: Should this be request locale? Or something?

I don’t understand why this is happening, frankly.

jmromer: > Aren't we only setting I18n.locale to permitted locales?

Oh we're actually not -- that'd be why

```rb
# app/controllers/application_controller.rb L66-72 

def requested_locale
  @requested_locale ||=
    locale_from_request_params ||
    current_user&.preferred_language.presence ||
    locale_from_request_header ||
    I18n.default_locale
end
```

jmromer: Hastily put together, but something like this might resolve the issue upstream?

```rb
# app/controllers/application_controller.rb L66-78 

def requested_locale
  return @requested_locale if defined?(@requested_locale)

  @requested_locale ||=
    locale_from_request_params ||
    current_user&.preferred_language.presence ||
    locale_from_request_header ||
    I18n.default_locale

  unless @requested_locale.in?(I18n.available_locales)
    @requested_locale = I18n.default_locale
  end
end
```

sethherr: I think we need to create a test that replicates the actual issue and then resolve it without rescuing random things. I'll get to that later this afternoon

jmromer: Reproduced and tested manually via browser settings -- we have some requests specs that send an invalid locale, I'm not sure why those aren't triggering the exception though.

sethherr: Me either!

Want to take this over @jmromer ? I don't think we need to investigate more deeply, if you can set headers in the locale request and trigger the error, that's good enough.

jmromer: No prob -- I think the see the problem in the specs

spec/requests/locale_detection_request_spec.rb
```diff
@@ -43,7 +43,9 @@
 
     context "given an unrecognized ACCEPT_LANGUAGE header" do
       it "renders the homepage in the default language" do
-        get "/", {}, { "HTTP_ACCEPT_LANGUAGE" => "es-ES;q=0.8,nl-NL;q=0.7" }
+        unrecognized_locale = :zh
+        expect(I18n.available_locales).to_not include(unrecognized_locale)
+        get "/", {}, { "HTTP_ACCEPT_LANGUAGE" => "#{unrecognized_locale};q=0.8" }
```
jmromer: @sethherr This is the test that should have been failing.

PR #1146, opened 2019-08-06
Remove unecessary layout code

Robertgervais:


Robertgervais:
Okay, I removed a lot of stuff. I'd love to remove more, because it feels good, but also I'm not sure if anything else *needs* to be deleted. Please advise or not.

app/assets/javascripts/application.js
```diff
@@ -32,11 +32,8 @@
 //= require external_scripts/jquery.fileupload.js
 //= require dataTables/jquery.dataTables
 //= require dataTables/bootstrap/2/jquery.dataTables.bootstrap
-//= require underscore
-//= require backbone
```
sethherr: So pumped we get to remove this finally!

Why do we need this file at all?

Robertgervais: Honestly I don't think we do! It just seemed like a big move to remove it. I'll dump it and see if anything breaks. Love to delete.

app/assets/javascripts/application.js
```diff
@@ -1,42 +0,0 @@
...
-//
-// It's not advisable to add code directly here, but if you do, it'll appear at the bottom of the
-// the compiled file.
-//
-// WARNING: THE FIRST BLANK LINE MARKS THE END OF WHAT'S TO BE PROCESSED, ANY BLANK LINE SHOULD
-// GO AFTER THE REQUIRES BELOW.
-//
-//= require jquery
-//= require jquery_ujs
-//= require external_scripts/bootstrap-transition.js
```
sethherr: Worth checking for each of these external scripts if they are still required.

Robertgervais: Mmm, I looked. Some of them are referenced in revised, and they have external script files in `external_scripts`, but otherwise none of them are referenced elsewhere. Should I delete the external script files?

Robertgervais:
#175

PR #1147, opened 2019-08-06
Add AlertImage model, update stolen_bike admin, kill logging

jmromer:
- Adds `AlertImage` model to back alert image versions
- Update stolen records admin to use the new model
- Updates specs so that image files are created on an opt-in basis with the inclusion of a trait: `:with_alert_image` `:with_image`, etc

Screen Shot 2019-08-06 at 6 29 42 PM

### DIFF
--- a/app/uploaders/alert_image_uploader.rb
+++ b/app/uploaders/alert_image_uploader.rb
@@ -2,14 +2,14 @@ class AlertImageUploader < ApplicationUploader
   include CarrierWave::MiniMagick
   include ::CarrierWave::Backgrounder::Delay
 
-  alias stolen_record model
+  delegate :stolen_record, to: :model
 
   def store_dir
-    "#{base_store_dir}/#{stolen_record.id}"
+    "#{base_store_dir}/#{model.id}"
   end
 
   def base_store_dir
-    "uploads/#{stolen_record.class.to_s[0, 2]}"
+    "uploads/#{model.class.to_s[0, 2]}"
   end
 
   def extension_white_list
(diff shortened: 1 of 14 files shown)


app/uploaders/alert_image_uploader.rb
```diff
@@ -2,14 +2,14 @@ class AlertImageUploader < ApplicationUploader
   include CarrierWave::MiniMagick
   include ::CarrierWave::Backgrounder::Delay
 
-  alias stolen_record model
+  delegate :stolen_record, to: :model
```
sethherr: Ooooh, nice. I think this reads better.

db/migrate/20190806155914_create_alert_images.rb
```diff
@@ -0,0 +1,11 @@
+class CreateAlertImages < ActiveRecord::Migration
+  def change
+    create_table :alert_images do |t|
+      t.belongs_to :stolen_record, null: false, index: true, foreign_key: true
+      t.boolean :current, null: false, default: true
```
sethherr: 1. I'd prefer not to have default true on booleans
2. Rather than `current` as a boolean, I feel like recording `retired_at` and then making a dynamic method for `current?` adds some value without adding much extra work

jmromer: Updated in 9ab375e

jmromer:
> I don't understand why `AlertImage` should stick around when it's no longer current?

Good point! Updated in debacc2

PR #1150, opened 2019-08-07
Add fallback font, use `caption_font`

jmromer:
Fixes to #1143


jmromer:
Simple bugfix, merging eagerly

PR #1154, opened 2019-08-07
Commit translation sync changes in steps

jmromer:
Looks like if there's only commit in a PR, GitHub's "squash and merge" feature just uses the commit message's subject line as the commit title, rather than the PR title.

As a result, the `[ci skip]` in the subject line prevents CI from running when the branch is merged into master, and a deployment isn't triggered.

This patch performs translation commits in two steps, retaining [ci skip]. GitHub's squash + merge should then use the PR's commit message, which won't have [ci skip].

### DIFF
--- a/bin/check_translations
+++ b/bin/check_translations
@@ -48,17 +48,27 @@ else
   git stash
   git checkout -b ${BRANCH}
   git reset --hard origin/master
-  git stash pop
+
+  # Commit translation changes
   git stash pop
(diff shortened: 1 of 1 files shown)


jmromer:
Thought about that, but I'm not sure I see much value in triggering the extra intermediate build for translation copy changes. If by some chance something does break the build, it'll fail on the merge.

Issue #1096, opened 2019-07-28
Rename recovered bikes to abandoned bikes


sethherr:
We have "recovered" stolen records - but we also have "recovered bikes"

Comment from the Bike model:

```ruby
# "Recovered" bikes are bikes that were found and are waiting to be claimed.
# This is confusing and should be fixed so it no longer is the same word as stolen recoveries
```

"Abandoned" is a much better word and we should switch to using it.

PR #1156, opened 2019-08-07
Rename recovered bikes to abandoned bikes

Robertgervais:


Robertgervais:
#1096

app/models/bike.rb
```diff
@@ -80,7 +80,7 @@ class Bike < ActiveRecord::Base
   scope :impounded, -> { includes(:impound_records).where(impound_records: { retrieved_at: nil }).where.not(impound_records: { id: nil }) }
   # "Recovered" bikes are bikes that were found and are waiting to be claimed. This is confusing and should be fixed
   # so that it no longer is the same word as stolen recoveries
-  scope :non_recovered, -> { where(recovered: false) }
+  scope :non_recovered, -> { where(abandoned: false) }
```
sethherr: I'd like this scope to be renamed as well.

app/views/bikes/_bike.html.haml
```diff
@@ -16,7 +16,7 @@
         %ul.attr-list
           %li
             %span.attr-title.text-danger
-              = bike.recovered ? t(".recovered") : t(".stolen")
+              = bike.abandoned ? t(".recovered") : t(".stolen")
```
sethherr: Could you change the translation key so it matches expectations?

app/views/bikes/show.html.haml
```diff
@@ -141,7 +141,7 @@
                 = @bike.attr_list_item(@stolen_record.address(skip_default_country: true), t(".location"))
                 = @bike.attr_list_item(@stolen_record.locking_description, t(".locking_description"))
                 = @bike.attr_list_item(@stolen_record.lock_defeat_description, t(".locking_circumvented"))
-                - stolen_type = @bike.recovered ? 'recovered' : 'stolen'
+                - stolen_type = @bike.abandoned ? 'recovered' : 'stolen'
```
sethherr: Update text here. Probably need to change some of the text as well, in the translation file

db/migrate/20190807175348_rename_recovered_to_abandonded.rb
```diff
@@ -0,0 +1,10 @@
+class RenameRecoveredToAbandonded < ActiveRecord::Migration
+
+  def up
+    rename_column :bikes, :recovered, :abandoned
```
sethherr: You can do a `change` migration, no need for up and down

app/models/bike.rb
sethherr: 🗣

PR #1157, opened 2019-08-07
Minor changes to bikes edit and feedbacks show for deleted bikes

Robertgervais:


Robertgervais:
Image 8-7-19 at 3 45 PM
Image 8-7-19 at 3 44 PM

app/views/admin/bikes/_bike.html.haml
```diff
@@ -111,6 +111,15 @@
                   - if bike.creation_state.bulk_import.ascend?
                     %small
                       Ascend
+          %tr
```
sethherr: The display should be much more prominent when the bike is deleted. And it doesn’t need to show up at all when it is t deleted.

app/views/admin/feedbacks/show.html.haml
```diff
@@ -44,7 +44,12 @@
               %pre
                 %code
                   = @feedback.feedback_hash.to_yaml.gsub("---\n", "")
-
+        - if @feedback.feedback_type == "bike_delete_request"
+          %tr
+            %td
+              View Bike
+            %td
+              = link_to "Deleted Bike", edit_admin_bike_path(id: @feedback.feedback_hash["bike_id"])
```
sethherr: can you show if the bike is actually deleted before you click this link?

app/controllers/admin/feedbacks_controller.rb
```diff
@@ -11,6 +11,9 @@ def index
 
   def show
     @feedback = Feedback.find(params[:id])
+    if @feedback.feedback_type == "bike_delete_request"
+      @bike = Bike.unscoped.find(@feedback.feedback_hash["bike_id"])
```
sethherr: `find` will throw an error if the bike is truly deleted.

app/views/admin/bikes/edit.html.haml
```diff
@@ -16,6 +16,11 @@
   .col-12
     %h1
       Edit Bike
+    - if @bike.deleted?.present?
+      %h2.text-danger
+        Deleted at:
+        %span.text-danger.convertTime
```
sethherr: No need for text danger on this too.

app/controllers/bikes_controller.rb
```diff
@@ -28,6 +28,9 @@ def index
 
   def show
     @components = @bike.components
+    if @bike.deleted?
```
sethherr: This should actually go in the `find_bike` method which already has something similar

PR #1161, opened 2019-08-08
Add AlertImageUploader specs, AlertImageGenerator branching

jmromer:
- Adds specs for `AlertImageUploader` documenting behavior when base image is invalid or empty
- Adds branching around absent bike_url, bike_location in `AlertImageGenerator`

### DIFF
--- a/app/services/alert_image_generator.rb
+++ b/app/services/alert_image_generator.rb
@@ -99,15 +99,19 @@ class AlertImageGenerator
       alert.font caption_font
 
       # Overlay bike url within lower border
-      alert.gravity "South"
-      alert.pointsize 50
-      alert.draw "text 0,25 '#{bike_url}'"
+      if bike_url.present?
+        alert.gravity "South"
+        alert.pointsize 50
+        alert.draw "text 0,25 '#{bike_url}'"
+      end
 
       # Overlay bike location on RHS of top border
-      alert.gravity "Northeast"
-      alert.pointsize 110
-      alert.size "x#{header_height}"
-      alert.draw "text 30,30 '#{bike_location}'"
+      if bike_location.present?
+        alert.gravity "Northeast"
+        alert.pointsize 110
+        alert.size "x#{header_height}"
+        alert.draw "text 30,30 '#{bike_location}'"
+      end
     end
 
     alert_image
@@ -129,6 +133,7 @@ class AlertImageGenerator
 
   # The bike url to be displayed on the promoted alert image
   def bike_url
+    return if bike&.id.blank?
     "bikeindex.org/bikes/#{bike.id}"
   end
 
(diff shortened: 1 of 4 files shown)


app/services/alert_image_generator.rb
```diff
@@ -129,6 +133,7 @@ def bike_location
 
   # The bike url to be displayed on the promoted alert image
   def bike_url
+    return if bike&.id.blank?
```
sethherr: This happened?

spec/uploaders/alert_image_uploader_spec.rb
```diff
@@ -0,0 +1,44 @@
...
+        expect(image_path).to match(%r{tmp})
+      end
+
+      expect { image_assignment.call }.to raise_error(CarrierWave::ProcessingError)
+    end
+  end
+
+  context "given a valid bike image" do
+    let(:bike_with_valid_image) do
+      FactoryBot.create(:stolen_bike).tap do |bike|
```
sethherr: Good solution, but definitely took me a second to parse why it was useful.

app/services/alert_image_generator.rb
jmromer: I'm not sure how, but there's no FK constraint on this relation, so it can. Probably a cascading delete / destroy is missing:

```rb
StolenRecord.where.not(bike_id: Bike.select(:id)).count
# (312.5ms)  SELECT COUNT(*) FROM "stolen_records" WHERE "stolen_records"."current" = $1 AND ("stolen_records"."bike_id" NOT IN (SELECT "bikes"."id" FROM "bikes" WHERE "bikes"."example" = $2 AND "bikes"."hidden" = $3 AND "bikes"."deleted_at" IS NULL  ORDER BY listing_order desc))  [["current", "t"], ["example", "f"], ["hidden", "f"]]
# => 170


StolenRecord.where.not(bike_id: Bike.select(:id)).map(&:bike).uniq
# [nil]
```

sethherr: Ahhh. Yeah, probably because of `default_scope` stuff. Might be resolved by `bike_id` instead? But doesn't really matter. You've made your point!

spec/uploaders/alert_image_uploader_spec.rb
jmromer: Which part? If you mean the file-in-let / close-in-after-each combination, I'm thinking of refactoring the other image-related tests toward something similar. Opening files in factories without closing them can lead to exceptions (so far they're not too common though.

```rb
# spec/uploaders/alert_image_uploader_spec.rb L27-29 

# define in a let in order to close file handler during teardown
let(:valid_image) { File.open(Rails.root.join("app/assets/images/pandabike.jpg"), "r") }
after(:each) { valid_image.close }
```

sethherr: The `tap`.

jmromer: Ah good nudge, thanks -- somewhat vestigial. Simplified in e06e4f6

PR #1162, opened 2019-08-08
Add user image selection to premium alert purchasing flow

jmromer:
- Provides a UI for a user to select one of their bike's public images to be used in the promoted alert image. 

- Provides a preview rendered using CSS

- Begins generating / processing the alert image when the first stage of the promoted alert form is submitted

demo

### DIFF
--- a/app/assets/javascripts/revised/pages/bikes/edit_alert.coffee
+++ b/app/assets/javascripts/revised/pages/bikes/edit_alert.coffee
@@ -2,11 +2,37 @@ class BikeIndex.BikesEditAlert extends BikeIndex
   constructor: ->
     super()
     @$planSelectionForm = $('#js-select-plan-form')
+    @$imageSelectionContainer = $('#js-select-image-container')
     @initializeEventListeners()
 
   initializeEventListeners: =>
     @$planSelectionForm.on "click", ".js-plan-select", (event) =>
       @highlightSelectedPlan(event)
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/revised/pages/bikes/edit_alert.coffee
```diff
@@ -2,11 +2,37 @@ class BikeIndex.BikesEditAlert extends BikeIndex
   constructor: ->
     super()
     @$planSelectionForm = $('#js-select-plan-form')
+    @$imageSelectionContainer = $('#js-select-image-container')
     @initializeEventListeners()
```
sethherr: Can you render with the first image by default on page load?

app/services/alert_image_generator.rb
```diff
@@ -118,17 +118,10 @@ def build_square
...
-      if stolen_record.address_location.present?
-        stolen_record.address_location
-      else
-        bike.registration_location
-      end
-
-    # escape single-quotes: location is passed to Imagemagick CLI inside
-    # single-quotes
-    location.gsub("'", "\\'")
+    stolen_record.bike_location&.gsub("'", "\\'")
```
sethherr: Nice refactor

app/assets/javascripts/revised/pages/bikes/edit_alert.coffee
jmromer: Good call! updated in fd492e51

PR #1166, opened 2019-08-08
Fix landing page recovery display caching issues

sethherr:
- ~Remove cache from individual recovery display templates (I'm not super concerned about the rendering speed of the recovery display page - I'm very concerned about the homepage rendering speed). I can see an argument for putting this back in, I just removed it because it was extra stuff.~ There was no reason to remove this, I put it back
- ~Remove the instance variable because I like being explicit in the cache block~
- Expire cache blocks based on locales - which is required because things are cached based on locales.

### DIFF
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -5,7 +5,9 @@ class WelcomeController < ApplicationController
   # Allow iframes on the index URL because safari is an asshole, and doesn't honor our iframe options
   skip_before_filter :set_x_frame_options_header, only: [:bike_creation_graph, :index]
 
-  def index; end
+  def index
+    @recovery_displays = RecoveryDisplay.limit(5)
+  end
 
   def bike_creation_graph
     @height = (params[:height] || 300).to_i
--- a/app/controllers/welcome_controller.rb
+++ b/app/controllers/welcome_controller.rb
@@ -5,9 +5,7 @@ class WelcomeController < ApplicationController
   # Allow iframes on the index URL because safari is an asshole, and doesn't honor our iframe options
   skip_before_filter :set_x_frame_options_header, only: [:bike_creation_graph, :index]
 
-  def index
-    @recovery_displays = RecoveryDisplay.limit(5)
-  end
+  def index; end
 
   def bike_creation_graph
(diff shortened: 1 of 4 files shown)


app/controllers/welcome_controller.rb
```diff
@@ -5,9 +5,7 @@ class WelcomeController < ApplicationController
   # Allow iframes on the index URL because safari is an asshole, and doesn't honor our iframe options
   skip_before_filter :set_x_frame_options_header, only: [:bike_creation_graph, :index]
 
-  def index
-    @recovery_displays = RecoveryDisplay.limit(5)
-  end
+  def index; end
```
jmromer: Why leave the empty method definition?

app/views/welcome/index.html.haml
```diff
@@ -16,7 +16,7 @@
     .root-landing-recovery-stories
       .container
         #recovery-stories-container.extras-hidden
-          = render @recovery_displays
+          = render RecoveryDisplay.limit(5)
```
jmromer: MVC nit: In general, having controller code in templates is an undesirable pattern to imitate — templates are happier the dumber they are.

As a smell of the same: this seems unidiomatic for Rails. See, e.g.:

```erb
<% @products.each do |product| %>
  <% cache product do %>
    <%= render product %>
  <% end %>
<% end %>
```
https://edgeguides.rubyonrails.org/caching_with_rails.html
https://blog.appsignal.com/2018/04/03/russian-doll-caching-in-rails.html
https://medium.com/@adrdilauro/master-html-russian-doll-caching-to-cut-server-response-time-e74cced05247

app/controllers/welcome_controller.rb
sethherr: Because I like having method definitions for all routes.

app/views/welcome/index.html.haml
sethherr: Fine

app/controllers/welcome_controller.rb
jmromer: Rails does it for us! :) But I guess I can see how it can serve as documentation to have the method signature there.

PR #1171, opened 2019-08-08
Fix alert image css preview

jmromer:
The current alert image preview CSS rules assume >100% font size. (which is what I have locally because because my eyes are trash). This updates the CSS rules to make them more robust to font size changes.

Before:

Screen Shot 2019-08-08 at 7 19 13 PM

After:

demo-branch


jmromer:
Hotfixing since this is broken on prod

PR #1173, opened 2019-08-09
Specify which paid features are shared with child organizations

sethherr:
Example:

Edmonton has paid for QR stickers. All the shops in Edmonton need access to the QR sticker interface.

The shops in Edmonton _should not_ be forced to use passwordless log in.

### DIFF
--- a/app/models/invoice.rb
+++ b/app/models/invoice.rb
@@ -76,7 +76,9 @@ class Invoice < ActiveRecord::Base
     end
   end
 
-  def assign_child_paid_feature_slugs=(val)
+  def child_paid_feature_slugs_string; child_paid_feature_slugs.join(", ") end
+
+  def child_paid_feature_slugs_string=(val)
     return true unless val.present?
     unless val.is_a?(Array)
       val = val.strip.split(",").map(&:strip)
@@ -152,6 +154,7 @@ class Invoice < ActiveRecord::Base
       self.subscription_end_at ||= subscription_start_at + subscription_duration
     end
     self.is_active = !expired? && (force_active || paid_in_full?)
+    self.child_paid_feature_slugs ||= []
     true # TODO: Rails 5 update
   end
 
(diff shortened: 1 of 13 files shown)


app/models/invoice.rb
```diff
@@ -76,6 +76,17 @@ def paid_feature_ids=(val) # This isn't super efficient, but whateves
     end
   end
 
+  def child_paid_feature_slugs_string; (child_paid_feature_slugs || []).join(", ") end
+
+  def child_paid_feature_slugs_string=(val)
+    return true unless val.present?
```
jmromer: nit: `if ... blank?` is available here and doesn't suffer the [drawbacks of `unless`](http://www.railstips.org/blog/archives/2008/12/01/unless-the-abused-ruby-conditional/)

sethherr: Good call. Fixed in 5767fbc

Do you see drawbacks the `unless` on the following line?

jmromer: _attention conservation notice: bike-shedding_

In the context of this method body, `unless ... Array` signals that `Array` is the typical input case, which is useful (assuming that's an accurate reading), but now that you mention it, the only type that responds to `.strip` here would be `String`, so the following also feels more truthful, in some (I'm sure) annoyingly hand-wavy sense of the term:

```rb
    if val.is_a?(String)
      val = val.strip.split(",").map(&:strip)
    end
```

sethherr: Fair. That seems like a reasonable response. But ultimately, I think either one would be ok, so I'm leaving it.

My preference is unless because I feel like the `unless val.is_a?(Array)` documents that we're trying to convert this into an array. And if we get something that isn't a string and call `.strip` on it, the error will make sense to me. YMMV

jmromer: > unless val.is_a?(Array) documents that we're trying to convert this into an array.

Makes sense, although I actually think `if !...Array` does a better job of that, but let me not start a holy war :) 

`unless` === `if !` is one of the Great Lies of Ruby imo

PR #1187, opened 2019-08-12
BikeHub sign in page

Robertgervais:


Robertgervais:
Image 8-14-19 at 4 48 PM

app/assets/stylesheets/revised/sections/partner_sign_in_up.scss
```diff
@@ -125,3 +162,49 @@ nav.bikehub-nav {
     }
   }
 }
+
+
+.bikehub-signup-landing-page.landing-page {
+  @include media-breakpoint-down(lg) {
```
sethherr: One of the biggest reasons to use `@include` is to make it so that the media queries can be in the same place the nested selectors are, rather than a separate group.

PR #1189, opened 2019-08-13
[i18n] Translation update: Build 6465

jmromer:
Merge to unblock CI job 6465: https://circleci.com/gh/bikeindex/bike_index/6465

### DIFF
--- a/config/locales/.translation_io
+++ b/config/locales/.translation_io
@@ -1,2 +1,2 @@
 ---
-timestamp: 1565644558
+timestamp: 1565710359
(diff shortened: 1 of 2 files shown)


sethherr:
@jmromer this clobbered the fix. How can we actually fix this?

jmromer:
@sethherr Does `sync_and_purge` persist the interpolation key change (to translation.io)? If not, we may need to update the interpolation key on translation.io

sethherr:
It does not appear to fix interpolation. This PR is being replaced by #1191, and the incorrect interpolation needs to be fixed on translation.io

PR #1190, opened 2019-08-13
Add StopHelingClient

jmromer:
- Adds an API client for the StopHeling web service
- Translates query params and response object keys


app/controllers/api/v3/search.rb
```diff
@@ -98,6 +98,30 @@ def forwarded_ip_address
...
+               notes: <<-NOTE,
+                 This endpoint accepts a serial number and searches external registries for it.
+
+                 It returns matches found.
+               NOTE
+             }
+        paginate
+        params do
+          requires :serial, type: String, desc: "Serial, homoglyph matched"
+          optional :per_page, type: Integer, default: 25, desc: "Bikes per page (max 100)"
```
sethherr: I think we should accept all the search parameters, and we can improve how accurately they match as time goes on.

I think we should launch with the proximity parameters.

app/models/external_registries/stop_heling_client.rb
```diff
@@ -0,0 +1,115 @@
...
+      case response_body
+      when Array
+        response_body
+          .map { |result| translate_keys(result) }
+          .map { |attrs| StopHelingResult.new(**attrs) }
+          .select(&:bike?)
+          .map(&:to_external_bike)
+      else
+        if Rails.env.production?
+          Honeybadger.notify(request: req_params, response: response_body)
```
sethherr: Hmmmm, what? Why?

app/serializers/external_bike_v3_serializer.rb
```diff
@@ -0,0 +1,81 @@
+class ExternalBikeV3Serializer < ActiveModel::Serializer
+  attributes \
```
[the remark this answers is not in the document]
sethherr: Why not inherit from the v3 bike serializer?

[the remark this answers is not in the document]
sethherr: You shouldn't need to put delegate here - it automatically delegates

app/serializers/bike_v2_serializer.rb
```diff
@@ -1,17 +1,41 @@
 class BikeV2Serializer < ActiveModel::Serializer
```
sethherr: Wait, is this modifying our live API V2?

sethherr: Ahh. It's just adding new things.

PR #1194, opened 2019-08-14
Fix issue with ambassador membership assignment

sethherr:
Fix an exception that occurs when inviting an unregistered user to an ambassador organization:

Resolves https://app.honeybadger.io/projects/35931/faults/53420005

Continuation of #1183


sethherr:
We test `expect(user.ambassador_tasks).to eq([ambassador_task])` which I think covers what you're talking about?

`user.perform_create_jobs` triggers the job that assigns the memberships, by calling `ProcessMembershipWorker` again.

(Once we make the Rails 5 upgrade, we won't need to call that manually in the specs because it will trigger after commit during specs.)

PR #1199, opened 2019-08-19
Externalize organized skeleton templates

### LINKED ISSUE #1185: Add translations to skeletons
There are some translations that make otherwise translated pages not be completely translated.

Priority templates:

- `_organized_skeleton.html.haml`
- `_edit_bike_skeleton.html.haml`

We can skip the content skeleton for now.

jmromer:
- Refactors to fix exceptions from `Organization#child_ids` being nil.
- Externalizes strings from `_organized_skeleton.html.haml` and `_edit_bike_skeleton.html.haml`

Resolves #1185

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -209,7 +209,7 @@ class Organization < ActiveRecord::Base
       self.slug = new_slug
     end
     self.access_token ||= SecurityTokenizer.new_token
-    self.child_ids = calculated_children.pluck(:id)
+    self.child_ids = calculated_children.pluck(:id) || []
     set_auto_user
     set_ambassador_organization_defaults if ambassador?
     locations.each { |l| l.save unless l.shown == allowed_show }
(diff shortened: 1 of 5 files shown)


app/models/organization.rb
```diff
@@ -227,7 +227,7 @@ def current_invoices; invoices.active end
   def current_parent_invoices; Invoice.where(organization_id: parent_organization_id).active end
 
   def incomplete_b_params
-    BParam.where(organization_id: child_ids + [id]).partial_registrations.without_bike
+    BParam.where(organization_id: [*child_ids, id]).partial_registrations.without_bike
```
sethherr: Is this better than using flatten? I feel like the splat is surprising, but maybe that's just me

jmromer: I'm used to it from seeing the pattern in JS, but it might be on the terse side for Ruby — updated in ce440a08

PR #1214, opened 2019-08-30
Port stolen bike twitter alerts to bike_index

jmromer:
- Ports `TwitterTweeterIntegration` from `stolen_bike_alerter` to `bike_index` and invokes it directly
- Adds `twitter_accounts`
- Associates `tweets` with `stolen records` and `twitter_accounts`
- Adds admin area for twitter accounts

Examples:
https://twitter.com/BikeIndexTest


.env
```diff
@@ -39,3 +37,13 @@ STOP_HELING_BASE_URL: something
 STOP_HELING_API_KEY: something
 STOP_HELING_APP_ID: something
 VERLOREN_OF_GEVONDEN_BASE_URL: something
+STOLEN_ALERT_TWITTER_1_USERNAME: test
```
sethherr: Is there a better way to store these? This seems sort of gross.

jmromer: Can you elaborate? I'm not sure what gross is calling out here, so I don't know what would constitutes "better" conditional on that metric. 

If it's something to do with the naming pattern, that's an attempted improvement over the following original naming:

```
TEST_SCREEN_NAME
CONSUMER_KEY
CONSUMER_SECRET
ACCESS_TOKEN
ACCESS_TOKEN_SECRET
OMNIAUTH_CONSUMER_KEY
OMNIAUTH_CONSUMER_SECRET
SECOND_TEST_SCREEN_NAME
SECOND_ACCESS_TOKEN
SECOND_ACCESS_TOKEN_SECRET
```

app/models/tweet.rb
```diff
@@ -59,10 +77,33 @@ def tweetor_name
...
-    "https://twitter.com/#{tweetor}/status/#{twitter_id}"
+    if twitter_account.present?
+      [twitter_account.twitter_account_url, "status", twitter_id].join("/")
+    else
+      "https://twitter.com/#{tweetor}/status/#{twitter_id}"
+    end
+  end
+
+  def details_hash
+    @details_hash ||= begin
```
sethherr: Is begin idiomatic here? I feel like I've only seen begin for rescuing, and it seems surprising.

app/models/twitter_account.rb
```diff
@@ -0,0 +1,78 @@
+class TwitterAccount < ActiveRecord::Base
+  scope :active, -> { where(active: true) }
+  scope :national, -> { active.where(national: true) }
+
+  has_many :tweets, dependent: :destroy
+
+  serialize :twitter_account_info
```
sethherr: Any reason why you do this rather than using a json (or jsonb) column? I believe json is more performant, and you can query the db for matching attributes rather than needing to instantiate...

app/controllers/admin/twitter_accounts_controller.rb
```diff
@@ -0,0 +1,74 @@
...
+
+  private
+
+  def permitted_parameters
+    params.require(:twitter_account).permit(
+      :active,
+      :address,
+      :append_block,
+      :city,
+      :consumer_key,
```
sethherr: Are we using multiple/different keys currently?

app/workers/approve_stolen_listing_worker.rb
```diff
@@ -2,6 +2,47 @@ class ApproveStolenListingWorker < ApplicationWorker
   sidekiq_options queue: "notify"
 
   def perform(bike_id)
-    StolenTwitterbotIntegration.new.send_tweet(bike_id)
+    bike = Bike.find(bike_id)
+    new_tweet = TwitterTweeterIntegration.new(bike).create_tweet
```
sethherr: I think it might make sense to put this into a background job, and have that call the email (or put them both into the same job).

But that can be done later.

app/models/twitter_account.rb
sethherr: Ahhh. Because it's a direct port from the existing service. I still think it would be worth switching to jsonb here.

db/migrate/20190904161424_add_columns_to_tweets.rb
```diff
@@ -0,0 +1,9 @@
+class AddColumnsToTweets < ActiveRecord::Migration
+  def change
+    change_table :tweets do |t|
+      t.string :twitter_account_id
+      t.string :stolen_record_id
```
sethherr: Why strings rather than references? I think they were strings originally because I suggested making any external ids strings, but since this is internal, I think it might make more sense to make them standard references with index.

spec/factories/tweets.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :tweet do
-    twitter_id { "874644243737751553" }
+    twitter_id { 18.times.map { (0..9).entries.sample }.join }
```
sethherr: Ha! I like it 🙂

.env
sethherr: If feels like there are a bunch of duplicates, and if there was a way to pass them more directly in, that would be cool. But I understand why that would be much more difficult.

I don't have a good answer to what to do instead. You can ignore this.

app/controllers/admin/twitter_accounts_controller.rb
sethherr: Does this mean that when you sign up a new account, you have to go to a twitter developer panel to collect the API keys, etc? If so, that's fine, we just need to document it.

app/models/tweet.rb
jmromer: > Is begin idiomatic here?


'tis

https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/MultilineMemoization
https://medium.com/helpful-tips/memoization-patterns-in-ruby-9efacbcee130
https://www.justinweiss.com/articles/4-simple-memoization-patterns-in-ruby-and-one-gem/
http://www.biggillystyle.com/2014/11/28/ruby-memoization-techniques.html
http://vaidehijoshi.github.io/blog/2015/11/10/methods-to-remember-things-by-ruby-memoization/

app/controllers/admin/twitter_accounts_controller.rb
jmromer: > Are we using multiple/different keys currently?

We are — each account has its own set of key, token, and secrets

> Does this mean that when you sign up a new account, you have to go to a twitter developer panel to collect the API keys, etc?

Yup! Added validation to those fields and a note to the admin dashboard — 67699a84

Screen Shot 2019-09-10 at 5 10 04 AM

app/workers/approve_stolen_listing_worker.rb
jmromer: > I think it might make sense to put this into a background job

Why's that? 🤔 (`TwitterTweeterIntegration` is invoked in a bg job)

db/migrate/20190904161424_add_columns_to_tweets.rb
jmromer: Good call thanks — Updated in 55a014c8

.env
jmromer: These env vars are only needed in `.env.test`—they're only in `.env` here for documentation. I'll add a note to the README. Happy to remove them from `.env` if you think their inclusion creates confusion.

app/models/twitter_account.rb
jmromer: Good call thanks — the text column is vestigial from the original implementation. Updated in 6ae1d74

app/controllers/admin/twitter_accounts_controller.rb
sethherr: Great. Thanks!!

.env
sethherr: Nah. It’s fine. Skip it

app/workers/approve_stolen_listing_worker.rb
sethherr: Omg I have no idea how I got lost here. Thought this was a controller. Ignore

app/models/tweet.rb
sethherr: Cool!

PR #1221, opened 2019-09-03
Fix placeholder image in bike serializer

jmromer:
Resolves https://app.honeybadger.io/projects/35931/faults/54294084

### DIFF
--- a/app/serializers/bike_v2_serializer.rb
+++ b/app/serializers/bike_v2_serializer.rb
@@ -58,13 +58,11 @@ class BikeV2Serializer < ActiveModel::Serializer
   end
 
   def placeholder_image
-    svg_path =
-      Rails
-        .application
-        .assets["revised/bike_photo_placeholder.svg"]
-        .digest_path
+    assets = Rails.application.assets
+    return if assets.blank?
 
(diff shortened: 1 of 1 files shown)


jmromer:
Hotfixing

app/serializers/bike_v2_serializer.rb
```diff
@@ -58,13 +58,11 @@ def date_stolen_string
...
-    svg_path =
-      Rails
-        .application
-        .assets["revised/bike_photo_placeholder.svg"]
-        .digest_path
+    assets = Rails.application.assets
+    return if assets.blank?
 
-    "#{ENV["BASE_URL"]}/assets/#{svg_path}"
+    svg_path = assets["revised/bike_photo_placeholder.svg"]&.digest_path
```
sethherr: Should we create some sort of helper for this? Or some sort of caching? Does this create any performance problems?

sethherr: Nevermind, this is resolved by #1223

jmromer: This was an expedient fix — https://github.com/bikeindex/bike_index/pull/1223 takes a better approach.

PR #1222, opened 2019-09-03
Don't search external registries if no serial given

jmromer:
Prevent External Registry Search component from mounting if no serial number has been entered.


jmromer:
Included in https://github.com/bikeindex/bike_index/pull/1223

PR #1223, opened 2019-09-03
Fix external registry search placeholder images

jmromer:
- Assets aren't available from the API
- Sets bike_placeholder_image_path on the window object to fall back to it in the React code

### Before

Screen Shot 2019-09-03 at 1 54 22 PM

### After

Screen Shot 2019-09-03 at 1 53 39 PM


jmromer:
Hotfixing to fix this on prod

PR #1225, opened 2019-09-05
Externalize remaining templates

jmromer:
Externalizes strings from the last batch of templates.

Removes the following partials, which I can't find a reference to anywhere:

- app/views/stolen/_stickers.html.haml
- app/views/stolen_notifications/_form.html.haml


app/views/memberships/edit.html.haml
```diff
@@ -1,53 +1,45 @@
...
           %td
             - if bike.created_at.today?
-              %span.hidden
-                = bike.created_at.strftime("%-m.%-d.%Y %-H.%M")
-              Today at #{bike.created_at.strftime("%-I %p")}
-            - elsif Date.yesterday().beginning_of_day() <= bike.created_at
-              %span.hidden
-                = bike.created_at.strftime("%-m.%-d.%Y %-H.%M")
-              Yesterday at #{bike.created_at.strftime("%-I %p")}
+              %span.hidden= bike.created_at.strftime("%-m.%-d.%Y %-H.%M")
```
sethherr: 1. We have the `convertTime` js stuff now.
2. The `span.hidden` is a remanant of using datatables, so that things were sorted correctly.

So this entire chunk should be removed and replaced with the `convertTime` class and format

sethherr: Also, it might be worth checking to see if this entire table could be wholesale replaced by the organized bikes table or something similar.

jmromer: This template appears to be vestigial -- it's not referenced anywhere and the only memberships controller I'm seeing is admin-namespaced. Removed it in 58a89e9.

PR #1227, opened 2019-09-05
Add link to sell a bike on bikefair

sethherr:


config/locales/en.yml
```diff
@@ -4785,7 +4785,7 @@ en:
       add_a_bike: Add a bike
       add_a_lock: Add a lock
       add_a_photo: Add a photo
-      add_some_info: Add some info
+      add_some_info: Next steps
```
sethherr: Will the translation of this update when this PR is merged? Stay tuned to find out!

jmromer: Sadly no :(
https://github.com/bikeindex/bike_index/pull/1230

Even changing the key, running `sync_and_purge`, and then preparing translations again (to try to get a new automated translation) doesn't do it... 🤔

sethherr: Thanks for trying!

This definitely belongs in your blog about how translation went, and we should send a support request to translation.io

PR #1236, opened 2019-09-06
Reverse geocode stolen records

jmromer:
- Adds reverse geocoding to stolen records
- Adds StolenRecord#neighborhood

Extracted from #1214 

Re. tweets for international stolen records:

Looks like it might be a data issue -- when finding nearby twitter accounts, we use lat/lon and fall back to the default account for a given country...

```rb
# app/models/stolen_record.rb L64-69 (034a762f)

def twitter_accounts_in_proximity
  [
    TwitterAccount.default_account_for_country(country),
    TwitterAccount.active.near(self, 50),
  ].flatten.compact.uniq
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/034a762f/app/models/stolen_record.rb#L64-L69)]</sup>

```rb
# app/models/twitter_account.rb L44-46 (26066344)

def self.default_account_for_country(country)
  national.where(country: country).first || default_account
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/26066344/app/models/twitter_account.rb#L44-L46)]</sup>

A number of accounts are missing lat/long and country fields, and some national accounts are missing the national attribute. I'll correct this in the data migration.

```
screen_name	lat	lon	natl	country	consumer_key	address
BikeIndexMN	NULL	NULL	False	NULL	3iu.......3	730 E 22nd St, Minneapolis, MN 55404
bikeindexuk	NULL	NULL	False	NULL	3iu.......3	Whitechapel Rd, Whitechapel, London E1 1FR, UK
stolenbikesnl	NULL	NULL	False	NULL	3iu.......3	The Netherlands
stolenbikeseu	NULL	NULL	False	NULL	3iu.......3	Europe
stolenbikesco	NULL	NULL	False	NULL	3iu.......3	Denver CO
bikeindexinglew	NULL	NULL	False	NULL	3iu.......3	Inglewood, CA
BikeIndex	NULL	NULL	False	NULL	3iu.......3
``` 

Some records also have identical consumer keys, which seems odd to me — I'd expect them to differ by account, but it's a significant proportion of the accounts that are set up this way, so I'm guessing it's not an issue.


app/models/stolen_record.rb
```diff
@@ -51,6 +51,16 @@ def self.old_attr_accessible
   geocoded_by :address_override_show_address
   after_validation :geocode, if: lambda { (self.city.present? || self.zipcode.present?) && self.country.present? }
 
+  reverse_geocoded_by :latitude, :longitude do |stolen_record, results|
+    if (geo = results.first)
```
sethherr: Can you do `next` or `return` here rather than conditional assignment?

PR #1242, opened 2019-09-10
Add terminal notifier, growlnotify

jmromer:
#### Add terminal-notifier to Gemfile

Resolves the following error:

```
~/.asdf/installs/ruby/2.5.5/lib/ruby/gems/2.5.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:462: 
in `block in replace_bin_path': can't find executable terminal-notifier for gem terminal-notifier. 
terminal-notifier is not currently included in the bundle, perhaps you meant to add it to your Gemfile? (Gem::Exception)
```

#### Add growlnotify to bin/setup

Resolves the following error when starting sidekiq:

```
which: no growlnotify in (....)
Traceback (most recent call last):
        4: from ~/.asdf/installs/ruby/2.5.5/bin/ruby_executable_hooks:24:in `<main>'
        3: from ~/.asdf/installs/ruby/2.5.5/bin/ruby_executable_hooks:24:in `eval'
        2: from ~/.asdf/installs/ruby/2.5.5/bin/terminal-notifier:23:in `<main>'
        1: from ~/.asdf/installs/ruby/2.5.5/lib/ruby/gems/2.5.0/gems/bundler-1.17.3/lib/bundler/rubygems_integration.rb:482:in `block in replace_bin_path'
```

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -90,6 +90,7 @@ gem "secure_headers", "~> 2.5.0"
 # Frontend
 gem "bootstrap", "~> 4.0.0.alpha3" # Bootstrap 4 - used for revised stylesheets
 gem "chartkick" # Display charts
+gem "coderay" # Pretty print code
 gem "coffee-rails"
 gem "groupdate" # Required for charts
 gem "jquery-datatables-rails", "~>3.4.0"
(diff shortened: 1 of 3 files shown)


jmromer:
> Could we just remove growlnotify? I don't think it's necessary?

👍 I don't use it -- thought it might be a Guard dependency. I'll circle back to remove it though

sethherr:
I think it used to be used by guard, but is no longer required.

PR #1246, opened 2019-09-11
Collapse columns, add link to twitter account

jmromer:
Adds some UI polish to the twitter accounts admin dashboard

Screen Shot 2019-09-11 at 10 25 43 AM


sethherr:
Personal usability preference here:

- I don't think there is a reason to show the keys on the index page
- I don't think there is a reason to show the lat & long on the index page
- I like having a column with a single value, and having that be a checkmark for important things, so that it's easier to match (active, default, national)

jmromer:
Updated in d8c58fb

PR #1266, opened 2019-09-17
Add possibly-found-bike searching and notifications

jmromer:
- Adds `EmailBikePossiblyFoundNotificationWorker`, which runs daily
- Finds stolen bikes that have a counterpart record with a matching normalized serial
- Emails each stolen bike's `owner_email` with information about their bike (the target stolen bike) and the matching bike
- Does not trigger an email if we've already contacted that specific owner about that bike/match pair or if the owner has disabled notifications for the associated stolen record

```rb
# app/workers/email_bike_possibly_found_notification_worker.rb L12-23 (e56ccbd5)

def notify_of_bike_index_held
  Bike.possibly_found_with_match.each do |bike, match|
    next if CustomerContact.possibly_found_notification_sent?(bike, match)

    contact = CustomerContact.build_bike_possibly_found_notification(bike, match)
    next unless contact.receives_stolen_bike_notifications?

    email = CustomerMailer.bike_possibly_found_email(contact)
    contact.email = email
    email.deliver_now if contact.save
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/e56ccbd5/app/workers/email_bike_possibly_found_notification_worker.rb#L12-L23)]</sup>

```rb
# app/models/customer_contact.rb L25-34 (38aa84ce)

# Given a Bike `bike` and a corresponding matching record `match` (a Bike or
# ExternalBike), determine if an email has been sent alerting the current
# `bike` owner that the given `match` may be their found bike.
def self.possibly_found_notification_sent?(bike, match)
  where(kind: kinds["bike_possibly_found"], bike: bike, user_email: bike.owner_email)
    .where("info_hash->>'match_id' = ?", match.id.to_s)
    .where("info_hash->>'match_type' = ?", match.class.to_s)
    .where("info_hash->>'stolen_record_id' = ?", bike&.current_stolen_record&.id.to_s)
    .exists?
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/38aa84ce/app/models/customer_contact.rb#L25-L34)]</sup>

Generated SQL:

```sql
SELECT  1 AS one 
FROM "customer_contacts" 
WHERE "customer_contacts"."kind" = $1 
AND "customer_contacts"."user_email" = $2 
AND "customer_contacts"."bike_id" = 658626 
AND (info_hash->>'match_id' = '658233') 
AND (info_hash->>'match_type' = 'Bike') 
AND (info_hash->>'stolen_record_id' = '76915') 
LIMIT 1  
```

```rb
# app/models/customer_contact.rb L36-45 (38aa84ce)

def self.build_bike_possibly_found_notification(bike, match)
  new(bike: bike,
      kind: kinds["bike_possibly_found"],
      info_hash: {
        stolen_record_id: bike&.current_stolen_record&.id.to_s,
        match_type: match.class.to_s,
        match_id: match.id.to_s,
        match: match.as_json,
      })
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/38aa84ce/app/models/customer_contact.rb#L36-L45)]</sup>

### Matching bikes

```rb
Bike
  .possibly_found_with_match
  .map { |a, b| { serial: a.serial_normalized, bike: [a.id, a.owner_email, a.frame_colors], match: [b.id, b.owner_email, b.frame_colors] } }
```

```json
[
  {
    "serial": "CD515347017",
    "bike": [
      634269,
      "...inh@gmail.com",
      ["Silver, gray or bare metal", "White"]
    ],
    "match": [
      642875,
      "...gan@gmail.com",
      ["Silver, gray or bare metal", "White"]
    ]
  },
  {
    "serial": "F3P8J0806",
    "bike": [
      62532,
      "...@yahoo.com",
      ["Silver, gray or bare metal"]
    ],
    "match": [
      59949,
      "...@gmail.com",
      ["Silver, gray or bare metal"]
    ]
  }
]
```

Screen Shot 2019-09-18 at 2 01 46 PM

Extracted external-registry searching to #1272 

Depends on:
https://github.com/bikeindex/bike_index/pull/1268
https://github.com/bikeindex/bike_index/pull/1269


sethherr:
Two comments on naming, which will have an impact on this email.

1. We have a state that we're adding here, and I think it makes sense to call it *held*.
2. Rather than "possible-recovery" I think it makes more sense to describe this as "possibly-found" - because there is another step in between found and recovery. I also think it will make more sense to the user to say "we may have found your stolen bike"

I think these two work together nicely: "We may have found your bike, it is being held by George"

Additional information on the new **held** state: we have *abandoned* which is the state of a bike that you find and register, *held* is after that - you found a bike and now you hold it.

app/models/bike.rb
```diff
@@ -75,8 +75,10 @@ class Bike < ActiveRecord::Base
   }
   scope :stolen, -> { where(stolen: true) }
   scope :non_stolen, -> { where(stolen: false) }
+  scope :abandoned, -> { where(abandoned: true) }
   scope :organized, -> { where.not(creation_organization_id: nil) }
   scope :with_known_serial, -> { where.not(serial_number: "unknown") }
+  scope :with_known_normalized_serial, -> { where.not(serial_normalized: %w[absent N0NE]) }
```
sethherr: This scope shouldn't be added, it only seems necessary because of incompletely migrated data (from #988). Rather than adding this scope, we should just fix the data:

- `absent` is the legacy version of `made_without_serial` - the serial_normalized fields that contain absent should be blanked out.
- Bikes with `none` should be switched to the current pattern - which is `serial_number: "made_without_serial",  made_without_serial: true` (it looks like there haven't been any new bikes added with "none" since the #988 shipped, so I think that update fixed that problem).

app/workers/email_held_bike_notification_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+
+  def perform
+    record_scheduler_started
+    notify_of_bike_index_held
+    record_scheduler_finished
+  end
+
+  def notify_of_bike_index_held
+    Bike.held_with_match.each do |bike, match|
+      next if CustomerContact.held_bike_notification.exists?(bike_id: bike.id, user_email: bike.owner_email)
```
sethherr: I believe there should be additional data stored (maybe in customer contact's info_hash):

- I think it's reasonable to use the `held_bike_notification`s to track the held matches, we need to record enough things to differentiate different matches. e.g. a bike is recovered then stolen again, a match is found but turns out to be incorrect and another match is found. So we at least need to store the `stolen_record_id`, the `match_id` and the `match_source` (whether it's Bike Index or somewhere else)
- `CustomerContact` record should be sufficient to send the email (and should be the thing that is passed in to the mailer)

app/workers/scheduled_worker_runner.rb
```diff
@@ -33,9 +33,17 @@ def self.write_worker_history(worker_string, record, value = Time.current)
...
+      UpdateExpiredInvoiceWorker,
+      UpdateCountsWorker,
+      UpdateOrganizationPosKindWorker,
+      FileCacheMaintenanceWorker,
+      UnusedOwnershipRemovalWorker,
+      TsvCreatorWorker,
+      GetManufacturerLogoWorker,
+      DeactivateExpiredTheftAlertWorker,
+      EmailHeldBikeNotificationWorker,
+    ] + [self]
```
sethherr: Nit: can we just include `self` in the regular array? (I know I did this, just seeing it now and wondering)

app/models/bike.rb
```diff
@@ -158,6 +160,38 @@ def organization(org_or_org_id)
       organization = org_or_org_id.is_a?(Organization) ? org_or_org_id : Organization.friendly_find(org_or_org_id)
       includes(:bike_organizations).where(bike_organizations: { organization_id: organization.id })
     end
+
+    # Bikes in a `held` state are stolen bikes that have a counterpart record
```
sethherr: I think `held` is actually just the state of an abandoned bike after it is registered (you register an abandoned bike, and then you hold that abandoned bike). So I don't think this definition works correctly, and I think it will be confusing down the line when we're dealing with external searching.

app/workers/email_held_bike_notification_worker.rb
```diff
@@ -0,0 +1,27 @@
...
+
+      email = CustomerMailer.held_bike_email(bike, match)
+      contact = CustomerContact.build_held_bike_notification(
+        bike: bike,
+        subject: email.subject,
+        body: email.text_part.to_s,
+        sender: email.from.first,
+      )
+
+      email.deliver_now if contact.save
```
sethherr: This needs to implement the check whether the user wants to receive notifications about the bike or not - https://github.com/bikeindex/bike_index/blob/master/app/workers/email_stolen_bike_alert_worker.rb#L7

I think it makes sense to put that logic into a method and check that method in both places. Initially, I was feeling like all the records should pass through the `EmailStolenBikeAlertWorker`, but I don't think that makes sense.

spec/mailers/previews/customer_mailer_preview.rb
```diff
@@ -55,6 +55,12 @@ def recovered_from_link
     CustomerMailer.recovered_from_link(recovered_record)
   end
 
+  def held_bike_email
+    target = Bike.stolen.limit(1).first
+    match = Bike.held.limit(1).first
+    CustomerMailer.held_bike_email(target, match)
+  end
```
sethherr: If all the data is in the CustomerMailer, you don't need to build something here.

This can just be `CustomerContact.held_bike_notification.last`. The other existing email previews should be similarly scoped, since there are no longer just one type of CustomerContact.

sethherr: Corollary of this (and other comments on this review) - I'm not certain `externally_held_bike_notification` needs to be a separate kind of `CustomerContact`. I think having scopes for both types makes sense, so it might make sense to have different kinds, but I think having them route through the same mailer method, and just render different HTML would be nice

app/models/bike.rb
jmromer: Thanks for clarifying!

Just to check my understanding: 

If a bike record's 

1. `serial_number` is any of `absent` `none` or `N0NE` (but not if it's `"unknown"`), or its
2. `serial_normalized` is any of `absent` `none` or `N0NE`

then the record needs to be migrated to have the following:

```rb
{
  made_without_serial: true
  serial_number: "made_without_serial"
  serial_normalized: nil
}
```

Is that correct?



Just want to make sure I understand correctly, because this will migrate about ~12% of the bikes table:

```rb
Bike.where(serial_number: ["absent", "none", "N0NE"]).count
# => 27
Bike.where(serial_normalized: ["absent", "none", "N0NE"]).count
# => 32032
```

Do we also expect a `serial_number` of `"unknown"` to have a blank `serial_normalized`?

```rb
Bike.where(serial_number: "unknown").where.not(serial_normalized: nil).count
# => 32189

Bike.where(serial_number: %w[unknown made_without_serial]).pluck(:serial_normalized).compact.uniq.sort

# => ["1 D0N T KN0W", "1 D0N T KN0W 1T", "1 D0NT KN0W", "1 D0NT KN0W 1T", "1 D0NT KN0W REMEM8ER", "180968", "1DK", "1N2N8143234", "48TC3329", "5AD1Y 1 D0N T KN0W", "5Q1862028", "C070143604", "CU5T0M", "CU5T0M 81KE", "CU5T0M 81KE N0 5ER1A1 HA5 A UN1QUE FRAME DE51GN", "CU5T0M 8U11T", "CU5T0M F100R 81KE", "CU5T0M FRAME", "CU5T0M FRAME N0 5ER1A1", "CU5T0M MADE", "D0 N0T KN0W", "D0N T KN0W", "D0N T KN0W 1T", "D0N T KN0W 5ER1A1", "D0N T KN0W THE 5ER1A1 NUM8ER", "D0NT KN0W", "D0NT KN0W 1T", "D0NT KN0W 1T 8UT R0CKY M0UNTA1N 5P0RT5 MA1NE 5T1CKER", "D0NT KN0W 1T YET WA1T1NG 0N 01D 0WNER", "F1 56 767", "F1T 111", "M0000000000N A", "M1551NG", "M1551NG 5ER1A1", "M1551NG 5ER1A1 NUM8ER", "N A", "N A CU5T0M 8U11T N0 5ER1A1", "N A DAK0TA CARVED 0N 80TT0M 8RACKET", "N AXXX", "N0 5ER1A1", "N0 5ER1A1 5PEC1A1 MADE", "N0 5ER1A1 HAND 8U11T 81KE", "N0 5ER1A1 NUM8ER", "N0 5ER1A1 NUM8ER CU5T0M 8U11D", "N0 5ER1A1 NUM8ER CU5T0M HEAD8ADGE", "N0NE CU5T0M", "N0NE CU5T0M HANDMADE 81KE", "PR08A81Y HA5 0NE D0N T KN0W 1T", "UNKN0WN", "YF006", "absent"]

```

I can add a callback to enforce all this — it looks like there's some other dirty / inconsistent fields:

```rb
Bike.where(serial_number: "made_without_serial").where.not(made_without_serial: true).count
=> 4

Bike.where(serial_number: "unknown").where.not(serial_normalized: nil).count
# => 32189
```

app/workers/scheduled_worker_runner.rb
jmromer: I left it as-is out of an abundance of caution, I'm not sure what the intent was but I'll update 👍

app/models/bike.rb
jmromer: Ah thanks. Agreed. I think I misunderstood your previous comment about the `held`  and `possibly_found` states. Does `Bike.possibly_found` seem more intuitive?

spec/mailers/previews/customer_mailer_preview.rb
jmromer: > I think having scopes for both types makes sense, so it might make sense to have different kinds, but I think having them route through the same mailer method, and just render different HTML would be nice

That's the plan at the moment — they have different `kind` enums but use the same mailer. But I can collapse the `kinds` if you think it's overly granular.

jmromer: > If all the data is in the CustomerMailer, you don't need to build something here.

> This can just be CustomerContact.held_bike_notification.last. The other existing email previews should be similarly scoped, since there are no longer just one type of CustomerContact.

Can you internalize your pronoun-antecedents here?

app/workers/scheduled_worker_runner.rb
sethherr: Sweet. I have no idea why I left it out originally 🤷‍♂️ - I assume it was just foolishness on my part.

app/models/bike.rb
sethherr: Yes! 👍💯

jmromer:
@sethherr Thanks! Updated the description with changes — the main ones are in 79c0c6f

spec/factories/bikes.rb
```diff
@@ -100,15 +100,15 @@
     end
 
     trait :blue_trek_930 do
+      frame_model { "930" }
```
sethherr: Any particular reason for this change?

spec/factories/users.rb
```diff
@@ -1,6 +1,6 @@
 FactoryBot.define do
   factory :user do
-    sequence(:name) { |n| "User #{n.to_s.rjust(3, "0")}" }
+    sequence(:name) { |n| "User #{n.to_s.rjust(4, "0")}" }
```
sethherr: We keep growing 🤗

spec/factories/bikes.rb
jmromer: Mainly vestigial (added a before(:save), turned out not to need it) but the attributes are sorted now.

spec/mailers/previews/customer_mailer_preview.rb
```diff
@@ -32,12 +32,26 @@ def invoice_email
   end
 
   def stolen_bike_alert_email
-    customer_contact = CustomerContact.last
+    customer_contact =
+      CustomerContact
+        .includes(:bike)
+        .limit(50)
+        .where.not(info_hash: {})
```
sethherr: Is the need to do this based on not having migrated data locally? If you pull a current version of the database is it still necessary?

jmromer: Nope — just pulled a fresh prod clone. We don't enforce referential integrity in the db, so some customer_contact records reference bikes that don't exist

spec/mailers/previews/customer_mailer_preview.rb
```diff
@@ -55,6 +69,13 @@ def recovered_from_link
     CustomerMailer.recovered_from_link(recovered_record)
   end
 
+  def bike_possibly_found_email
+    target = Bike.stolen.last
+    match = Bike.abandoned.last
+    contact = CustomerContact.build_bike_possibly_found_notification(target, match)
```
sethherr: Why do this rather than inspecting the most recent matching customer mailer?

Reason I like the most recent match - it's convenient if the mailer previews match the most recent because then you can go and generate another one of the things, and then view that instance in the mailer preview.

spec/factories/bikes.rb
sethherr: Cool. Just was curious.

spec/mailers/previews/customer_mailer_preview.rb
jmromer: Good point thanks! Updated in 23a9d1be

spec/mailers/previews/customer_mailer_preview.rb
jmromer: Simplified the query a bit in 23a9d1be

PR #1268, opened 2019-09-18
Migrate customer_contacts.contact_type to kind enum

jmromer:
- Adds a `kind` enum to `CustomerContact`
- Adds a data migration to populate `kind` with an appropriate mapping from `contact_type`
- Adds logic to keep `contact_type` and `kind` in sync until the former is dropped

### Post-deploy
```
bin/rake data:migrate_customer_contact_types:up
```

### DIFF
--- a/app/models/customer_contact.rb
+++ b/app/models/customer_contact.rb
@@ -3,10 +3,20 @@ class CustomerContact < ActiveRecord::Base
   belongs_to :user
   belongs_to :creator, class_name: "User"
 
+  KIND_ENUM = {
+    stolen_contact: 0,
+    stolen_twitter_alerter: 1,
+    held_bike_notification: 2,
+    externally_held_bike_notification: 3,
+  }.freeze
+
+  enum kind: KIND_ENUM
+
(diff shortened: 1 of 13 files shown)


app/models/customer_contact.rb
```diff
@@ -27,4 +37,11 @@ def normalize_emails_and_find_users
 
     true
   end
+
+  # TODO: Remove after `contact_type` is migrated to `kind` enum
+  before_save :sync_contact_type_and_kind
+
+  def sync_contact_type_and_kind
+    self[:kind] = self.class.kinds[self[:contact_type]]
```
sethherr: Why isn't this `self.kind`? Do we have to do this because the rails version? Or is this just a different way of doing it?

jmromer: In this case they're equivalent. I think `self[:kind]` was my default here because in the preceding data migration I had to get a value in the attribute hash bypassing the custom getter, and `self[...]` ensures you're getting the raw value. But it's not actually needed here.

PR #1276, opened 2019-09-19
Add language to blog posts

jmromer:
- Adds a `language` attribute to `Blog`, defaulting to English
- `blogs.language` enum includes only current locale languages
- Adds a drop-down to the public-facing blog index to select posts in other languages
- Current blog language defaults to the user's current locale
- Changing blog language doesn't change the user's current locale

### UI
Screen Shot 2019-09-19 at 1 35 58 PM
Screen Shot 2019-09-19 at 1 36 04 PM
Screen Shot 2019-09-19 at 1 36 11 PM

### Admin UI

Screen Shot 2019-09-19 at 1 38 14 PM

### DIFF
--- a/app/controllers/admin/news_controller.rb
+++ b/app/controllers/admin/news_controller.rb
@@ -1,7 +1,6 @@
 class Admin::NewsController < Admin::BaseController
   before_filter :find_blog, only: [:show, :edit, :update, :destroy]
   before_filter :set_dignified_name
-  before_filter :normalize_params
 
   def index
     @blogs = Blog.order("created_at asc")
@@ -83,11 +82,6 @@ class Admin::NewsController < Admin::BaseController
       :user_email,
       :user_id,
     )
-  end
-
-  def normalize_params
-    language = params.dig(:blog, :language)
-    params[:blog][:language] = language.to_i if language.present?
   end
 
   def set_dignified_name
--- a/app/controllers/admin/news_controller.rb
+++ b/app/controllers/admin/news_controller.rb
@@ -63,8 +63,26 @@ class Admin::NewsController < Admin::BaseController
   protected
 
   def permitted_parameters
-    params.require(:blog).permit(*%w(title body user_id published_at post_date post_now tags published old_title_slug
-                                     timezone description_abbr update_title is_listicle listicles_attributes user_email index_image_id index_image).map(&:to_sym).freeze)
+    params.require(:blog).permit(
+      :body,
+      :description_abbr,
+      :index_image,
+      :index_image_id,
+      :is_listicle,
+      :language,
+      :listicles_attributes,
+      :old_title_slug,
+      :post_date,
+      :post_now,
+      :published,
+      :published_at,
+      :tags,
+      :timezone,
+      :title,
+      :update_title,
+      :user_email,
+      :user_id,
+    )
   end
 
   def set_dignified_name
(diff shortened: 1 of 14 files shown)


app/controllers/admin/news_controller.rb
```diff
@@ -23,14 +24,13 @@ def edit
   end
 
   def update
-    body = "blog"
-    title = params[:blog][:title]
-    body = params[:blog][:body]
-    if @blog.update_attributes(permitted_parameters)
+    if @blog.update(permitted_parameters)
```
sethherr: 👍 💯 ✨

app/helpers/application_helper.rb
```diff
@@ -231,6 +231,11 @@ def language_choices
     end.sort { |a, b| a[0].downcase <=> b[0].downcase }
   end
 
+  def blog_languages
+    @blog_languages ||=
+      Blog.languages.map { |lang, num| [t(lang, scope: [:locales]), num] }
```
sethherr: Why make this a number?

sethherr: Ahhhh. for enums.

app/controllers/admin/news_controller.rb
```diff
@@ -63,8 +63,31 @@ def destroy
...
+      :timezone,
+      :title,
+      :update_title,
+      :user_email,
+      :user_id,
+    )
+  end
+
+  def normalize_params
+    language = params.dig(:blog, :language)
```
sethherr: When would there not be a language parameter in the update? Why not just include this in the parameters?

app/views/news/index.html.haml
```diff
@@ -1,9 +1,34 @@
...
+      View posts in
+      = form_tag nil, method: :get,
+        class: "blog-language-selection", id: "js-blog-language-selection" do
+        = select_tag :language,
+          options_for_select(blog_languages, selected: params[:language].presence),
+          onchange: "this.form.submit()"
   %hr
+
+:javascript
+  const setBlogLanguageSelection = () => {
```
sethherr: I'd also like the parameters to use "en"/"nl" rather than 0 and 1.

I don't think the language sticking when using the back button is a large enough concern to add this JS (concerns I have with this JS are that it will break older browsers, and that using "en"/"nl" will make it more complicated).

app/controllers/admin/news_controller.rb
jmromer: > When would there not be a language parameter in the update? 

I can't think of a situation where that would be the case.

> Why not just include this in the parameters?

It's ambiguous what "this" refers to here — orphaned pronouns are like missing external dependencies. 


If you mean why are we casting the language param to an integer here, it's because ActiveRecord raises an exception on `"1"` when it needs `1` 🤷‍♂ 

```
Blog.where(language: "1").first.update(language: "1")

Blog Load (0.6ms)  SELECT  "blogs".* FROM "blogs" WHERE "blogs"."language" = $1  ORDER BY published_at desc LIMIT 1  [["language", 1]]
 (0.1ms)  BEGIN
 (0.1ms)  ROLLBACK
ArgumentError: '1' is not a valid language
```

app/views/news/index.html.haml
jmromer: It looks pretty rough without it. Happy to convert to ES5 if that'd be better...

app/controllers/admin/news_controller.rb
sethherr: Awwwww bummers. I think this might be fixed in Rails 5.

Ok

app/views/news/index.html.haml
sethherr: Why does it look rough? If it's just the back browser state thing, I don't think that's much of a concern. When I navigate backward, I expect the state of select & text boxes to be sometimes funky.

jmromer: > the state of select & text boxes to be sometimes funky.

That's what I'm describing as rough

PR #1282, opened 2019-09-21
Add bike no-serial sanitizing

### LINKED ISSUE #1277: Catch and coerce people entering silly things for serial number
e.g. `none`, `idk`, etc.

jmromer:
- Sanitizes non-standardized forms of "made_without_serial"
- Adds a `before_save` hook to ensure the three components of "unknown" / "made_without_serial" / neither state are kept consistent with each other.

Resolves #1277

### DIFF
--- a/app/models/bike.rb
+++ b/app/models/bike.rb
@@ -454,15 +454,31 @@ class Bike < ActiveRecord::Base
     true
   end
 
-  def normalize_attributes
-    self.serial_number = "made_without_serial" if made_without_serial?
-    self.serial_number = SerialNormalizer.unknown_and_absent_corrected(serial_number)
-    self.serial_normalized = SerialNormalizer.new(serial: serial_number).normalized
+  def normalize_emails
     if User.fuzzy_email_find(owner_email)
       self.owner_email = User.fuzzy_email_find(owner_email).email
     else
       self.owner_email = EmailNormalizer.normalize(owner_email)
     end
+    true
+  end
+
+  def normalize_serial_number
+    if made_without_serial?
+      self.serial_number = "made_without_serial"
+      self.serial_normalized = nil
+      return true
+    end
+
+    self.serial_number = SerialNormalizer.unknown_and_absent_corrected(serial_number)
+    if serial_number == "made_without_serial"
+      self.serial_normalized = nil
+      self.made_without_serial = true
+    else
+      self.serial_normalized = SerialNormalizer.new(serial: serial_number).normalized
+      self.made_without_serial = false
+    end
+
     true
   end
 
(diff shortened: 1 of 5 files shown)


spec/models/bike_spec.rb
```diff
@@ -22,37 +23,86 @@
...
+          bike.normalize_serial_number
+          expect(bike.serial_number).to eq("unknown")
+          expect(bike.made_without_serial).to eq(false)
+          expect(bike.serial_normalized).to eq(nil)
+        end
       end
     end
   end
 
+  context "actual tests for ascend and lightspeed" do
```
sethherr: 💚

app/models/bike.rb
```diff
@@ -466,6 +463,30 @@ def normalize_attributes
...
+    if made_without_serial?
+      self.serial_number = "made_without_serial"
+      self.serial_normalized = nil
+      return true
+    end
+
+    self.serial_number = SerialNormalizer.unknown_and_absent_corrected(serial_number)
+
+    case serial_number
+    when "made_without_serial"
```
sethherr: nit: might be cleaner to add `made_without_serial?` to this `when` and then remove the above if statement. But doesn't really matter.

app/services/serial_normalizer.rb
```diff
@@ -3,11 +3,18 @@ def self.unknown_and_absent_corrected(str = nil)
...
     return "unknown" if str.blank? || str.gsub(/\s|\?/, "").blank? || str.downcase == "absent"
-    return "made_without_serial" if str == "made_without_serial"
+    return "made_without_serial" if str == "made_without_serial" || looks_like_made_without?(str.downcase)
     return "unknown" if looks_like_unknown?(str.downcase)
     str
   end
 
+  def self.looks_like_made_without?(str_downcase)
+    case str_downcase
+    when /no serial/i then true
```
sethherr: I feel like the string "no serial" should look like unknown, rather than made_without_serial.

The bikes that are made without serial are few and far between, and generally custom. If the person is unable to correctly check the "made without serial" bike flow, it should be marked as unknown, rather than "made without"

Caveat: we should add the made without checkbox flow and functionality to the embed forms.

app/models/bike.rb
jmromer: That would evaluate differently, and the semantics of the case statement make it harder to reason about:

```rb
def case_statement(serial_number, made_without_serial:)
  case serial_number
  when "made_without_serial", made_without_serial
    :made_without_serial
  else
    :made_with_serial
  end
end

case_statement("made_without_serial", made_without_serial: false)
# :made_without_serial

case_statement("serial_number", made_without_serial: true)
# :made_with_serial
```

app/services/serial_normalizer.rb
jmromer: Updated in 4488285c

jmromer:
> we should add the "made without serial" flow to embed forms

Makes sense! Created https://github.com/bikeindex/bike_index/issues/1284

Unless there's a dependency between #1284 and this PR that I'm missing, I'll ship this one and pick up the other asap

PR #1283, opened 2019-09-22
Twitter account remove error on update

sethherr:
I'm not sure what to make of this line in [twitter_account.rb](https://github.com/bikeindex/bike_index/blob/master/app/models/twitter_account.rb#L83):

```ruby
clear_error if twitter_client.verify_credentials.present?
```

I would like to add a checkbox to the admin page to `check_credentials`, but I don't want to mess with this line because I'm not sure why it's there. I don't understand why we wouldn't always clear the error before running `check_credentials`

It doesn't seem like anything in the app calls `check_credentials` anyway though?

### DIFF
--- a/app/controllers/admin/twitter_accounts_controller.rb
+++ b/app/controllers/admin/twitter_accounts_controller.rb
@@ -28,6 +28,10 @@ class Admin::TwitterAccountsController < Admin::BaseController
 
   def update
     if @twitter_account.update_attributes(permitted_parameters)
+      if ParamsNormalizer.boolean(params[:check_credentials])
+        @twitter_account.reload
+        @twitter_account.check_credentials
+      end
       flash[:notice] = "Twitter account saved!"
       redirect_to admin_twitter_account_url(@twitter_account)
     else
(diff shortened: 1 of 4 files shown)


jmromer:
> I would like to add a checkbox to the admin page to check_credentials

Sweet! 

> It doesn't seem like anything in the app calls check_credentials anyway though?

Yeah, I wouldn't take it as stable or anything — `check_credentials` was defined because it was useful for debugging and left in in anticipation of an admin dashboard button.

> I don't understand why we wouldn't always clear the error before running check_credentials

If `verify_credentials` raises an exception not caught by the rescue, we'll lose the last saved error--just being cautious I guess, but I agree that eagerly clearing the last error would be more legible from a readability standpoint 👍

sethherr:
So does `twitter_client.verify_credentials` actually verify the account?

jmromer:
> So does twitter_client.verify_credentials actually verify the account?

I'm not sure I follow, but I have tasted the fruits of its labor. What it says on the tin:

- https://www.rubydoc.info/gems/twitter/Twitter%2FREST%2FUsers:verify_credentials
- https://developer.twitter.com/en/docs/accounts-and-users/manage-account-settings/api-reference/get-account-verify_credentials

sethherr:
I have no idea why I asked that. Thanks for a kind RTFM 😀

PR #1286, opened 2019-09-24
Add Bike.currently_stolen_in

jmromer:
Introduces `Bike.currently_stolen_in` to query for bikes with currently stolen records in the given location (passed as city / state / country, matched conjointly).

Will be used to find currently stolen in NL to check against the VOG registry.

Extracted from #1280

### DIFF
--- a/app/models/bike.rb
+++ b/app/models/bike.rb
@@ -192,6 +192,22 @@ class Bike < ActiveRecord::Base
         pairs << [bike, match]
       end
     end
+
+    # Search for currently stolen bikes reported stolen in the given city, state
+    # and/or country. `city`, `state` and `country` are accepted as strings /
+    # symbols of the name or abbreviation, and are matched conjointly.
+    def currently_stolen_in(city: nil, state: nil, country: nil)
+      location = { city: city, state: state, country: country }.select { |_, v| v.present? }
+      location[:state] &&= State.find_by("name = ? OR abbreviation = ?", state, state)
+      location[:country] &&= Country.find_by("name = ? OR iso = ?", country, country)
+      return none if location.values.any?(&:blank?)
+
+      unscoped
+        .stolen
+        .current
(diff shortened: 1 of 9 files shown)


jmromer:
@sethherr Any idea what this test is testing? The description is `"does not get out of integer errors"` but I can't get it to produce any, it just fails because `expect(bike.listing_order).to be > 10000`...

```rb
# spec/models/bike_spec.rb L1308-1325 

  context "problem date" do
    let(:problem_date) do
      digits = (Time.current.year - 1).to_s[2, 3] # last two digits of last year
      Date.strptime("#{Time.current.month}-22-00#{digits}", "%m-%d-%Y")
    end
    let(:bike) { FactoryBot.create(:stolen_bike) }
    it "does not get out of integer errors" do
      expect(bike.listing_order).to be < 10000
      # We protect against this on stolen record now, so manually set this (still doesn't work :/)
      bike.current_stolen_record.update_attribute :date_stolen, problem_date
      # TODO: Rails 5 update - enable this, rspec doesn't correctly manage after_commit right now -
      # but stolen records don't actually have an after_commit hook to update bikes (they probably should though)
      # This is just checking this is called correctly on save
      bike.update_attributes(updated_at: Time.current)
      expect(bike.listing_order).to be > (Time.current - 13.months).to_i
    end
  end
end
```

```
  1) Bike calculated_listing_order problem date does not get out of integer errors
     Failure/Error: expect(bike.listing_order).to be < 10000
     
       expected: < 10000
            got:   1569331519
     # ./spec/models/bike_spec.rb:1315:in `block (4 levels) in <top (required)>'

Finished in 6.56 seconds (files took 6.77 seconds to load)
189 examples, 1 failure

Failed examples:

rspec ./spec/models/bike_spec.rb:1314 # Bike calculated_listing_order problem date does not get out of integer errors
```

PR #1287, opened 2019-09-24
Refactor CustomerContact.possibly_found_notification_sent?

jmromer:
Adds lower-level tests for `CustomerContact.possibly_found_notification_sent?`, adds a guard clause. Used in `EmailBikePossiblyFoundNotificationWorker` to prevent re-sending notices:

```rb
# app/workers/email_bike_possibly_found_notification_worker.rb L13-14 (ad0842fd)

Bike.possibly_found_with_match.each do |bike, match|
  next if CustomerContact.possibly_found_notification_sent?(bike, match)
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/ad0842fd/app/workers/email_bike_possibly_found_notification_worker.rb#L13-L14)]</sup>

```rb
# spec/workers/email_bike_possibly_found_notification_worker_spec.rb L36-51 (ad0842fd)

it "skips the bike if a notification for it has already been sent" do
  bike = FactoryBot.create(:stolen_bike, abandoned: false, serial_number: "HELL0")
  match = FactoryBot.create(:abandoned_bike, stolen: false, serial_number: "HE11O")

  # an already-sent notification
  contact = CustomerContact.build_bike_possibly_found_notification(bike, match)
  contact.email = CustomerMailer.bike_possibly_found_email(contact)
  contact.save

  allow(CustomerMailer).to receive(:bike_possibly_found_email).and_call_original

  described_class.new.perform

  expect(CustomerMailer).to_not have_received(:bike_possibly_found_email)
  expect(ActionMailer::Base.deliveries.length).to eq(0)
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/ad0842fd/spec/workers/email_bike_possibly_found_notification_worker_spec.rb#L36-L51)]</sup>

Extracted from #1280

### DIFF
--- a/spec/factories.rb
+++ b/spec/factories.rb
@@ -130,9 +130,6 @@ FactoryBot.define do
     factory :customer_contact_potentially_found_bike do
       creator { FactoryBot.create(:user) }
       bike { FactoryBot.create(:stolen_bike) }
-      title { "We may have found your bike" }
-      body { "Your bike may have been found" }
-      creator_email { "sender@example.com" }
       kind { :bike_possibly_found }
 
       transient do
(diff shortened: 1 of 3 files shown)


spec/factories.rb
```diff
@@ -126,5 +126,27 @@
     trait :stolen_bike do
       bike { FactoryBot.create(:stolen_bike) }
     end
+
+    factory :customer_contact_potentially_found_bike do
+      creator { FactoryBot.create(:user) }
+      bike { FactoryBot.create(:stolen_bike) }
+      title { "We may have found your bike" }
```
sethherr: Nit: Is it possible to generate the title, body and creator_email from the inputs (the bike and match)? I feel like it would be better if they were consistently generated in the same way.

jmromer: good call thanks — updated in a54f38b2

PR #1288, opened 2019-09-24
Add ExternalRegistry, ExternalRegistryBike models

jmromer:
- Adds `ExternalRegistry` and associated `ExternalRegistryBike` models
- Replaces `ExternalBike` data class and associated serializer with `ExternalRegistryBike`

Moves the all-external-registries query method to `ExternalRegistry`, which will allow it to be scoped to a subset of registries if needed (e.g. `ExternalRegistry.netherlands.search_for_bikes_with`)

```rb
# app/models/external_registry.rb L11-19 (d142b2db)

def self.search_for_bikes_with(serial_number:)
  results = all
    .map { |registry| Thread.new { registry.search_registry(serial_number: serial_number) } }
    .map(&:value)
    .flatten
    .compact

  ExternalRegistryBike.where(id: results.map(&:id))
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/d142b2db/app/models/external_registry.rb#L11-L19)]</sup>

### Post-deploy

This PR adds a rake task to persist registries that should be run post-deploy

```
% bundle exec rake data:seed_external_registries
Ensuring 2 external registries are persisted...
2/2 : 100.0%
Done!
```

Extracted from #1280


app/models/country.rb
```diff
@@ -25,6 +25,10 @@ def self.united_states
     where(name: "United States", iso: "US").first_or_create
   end
 
+  def self.netherlands
```
sethherr: There is also a place where this netherlands is referenced in `app/views/bikes/_stolen_checklist.html.haml`, and should be replaced with this.

jmromer: Updated! thanks - 939a2ea6

app/models/bike.rb
```diff
@@ -69,7 +69,7 @@ class Bike < ActiveRecord::Base
   enum propulsion_type: PropulsionType::SLUGS
 
   default_scope {
-    includes(:tertiary_frame_color, :secondary_frame_color, :primary_frame_color, :current_stolen_record)
+    includes(:tertiary_frame_color, :secondary_frame_color, :primary_frame_color, current_stolen_record: [:state, :country])
```
sethherr: I realize we've already committed to this default scope thing, but adding more to this still scares me.

Probably don't do anything about this. I guess we can just make the default scope everything we want, and just call unscoped on `Bike` wherever that makes sense. I just worry about it.

jmromer: I'm not crazy about it either, actually — I'll kill it. There's an N+1 query somewhere making Bullet mad, but adding to the default scope is overkill. Shouldn't be hard to track down.

sethherr: I think it would be great to add a named scope and put the includes in there. And then use that named scope all over.

I think the default scope hiding bikes that need to be hidden makes sense (even though I'm not in love with it).

But even moving the listing order into the named searching scope probably makes sense (or making it its own named scope and including it in the named searching scope, etc).

spec/rails_helper.rb
```diff
@@ -189,3 +189,14 @@ def initialize(meta)
   config.cache_dir = Rails.root.join("tmp", "cache", "carrierwave#{ENV["TEST_ENV_NUMBER"]}")
   config.enable_processing = false
 end
+
+# Create external registries
+RSpec.configure do |config|
+  config.before(:all) do
+    ExternalRegistry.create_all
```
sethherr: Do we really need to run this before every test?

app/models/external_registries/stop_heling_result.rb
```diff
@@ -45,33 +55,33 @@ def serial_number
...
+
+      bike.url = url
+      bike.cycle_type = "bike"
+      bike.status = "stolen"
+      bike.mnfg_name = brand
+      bike.frame_model = brand_type
+      bike.location_found = location
+      bike.frame_colors = colors
+      bike.date_stolen = date_found
+      bike.info_hash = json
```
sethherr: If we're persisting this in the DB, can we skip including the original JSON?

db/migrate/20190923181351_create_external_registries.rb
```diff
@@ -0,0 +1,12 @@
+class CreateExternalRegistries < ActiveRecord::Migration
+  def change
+    create_table :external_registries do |t|
```
sethherr: I don't understand why these are in the DB, rather than just ruby classes not backed by the database

sethherr: Something like the models that use enums for associations, e.g. [HandlebarType](https://github.com/bikeindex/bike_index/blob/master/app/models/handlebar_type.rb)

db/migrate/20190925162318_add_serial_normalized_to_external_registry_bike.rb
```diff
@@ -0,0 +1,9 @@
+class AddSerialNormalizedToExternalRegistryBike < ActiveRecord::Migration
+  def change
```
sethherr: please merge this into the create migration, since they're occurring in the same PR

app/serializers/external_registry_bike_v3_serializer.rb
```diff
@@ -1,8 +1,8 @@
-class ExternalBikeV3Serializer < BikeV2Serializer
+class ExternalRegistryBikeV3Serializer < BikeV2Serializer
   delegate \
```
sethherr: I don't know that you need to include this `delegate` at all?

db/migrate/20190923181351_create_external_registries.rb
sethherr: I think it might make more sense to follow the single table inheritance pattern (naming and file structure) for this, though making it a plain ruby object rather than an active record database backed thing makes more sense to me.

So the class would look like `ExternalRegistry::StopHeling`

This is sort of a nit, but I think it will fundamentally change how things look and work and make it less surprising.

app/serializers/external_registry_bike_v3_serializer.rb
sethherr: Is it because of the `allow_nil`?

If so, if we're using inheritance, we could specify the attributes on the `ExternalRegistryBike` base model.

sethherr: Also, I don't know why `date_stolen_string` was added in #1190 - it shouldn't have been. Sorry for not catching this earlier! It shouldn't be in there.

As per our [documentation](https://bikeindex.org/documentation/api_v3), time is represented by unix timestamps.

app/models/bike.rb
jmromer: Updated in fe316716

db/migrate/20190923181351_create_external_registries.rb
jmromer: > I don't understand why these are in the DB, rather than just ruby classes not backed by the database

For the sake of data normalization — the alternative is to duplicate the registry name and url (and any other potentially associated attributes) across the `external_registry_bikes` table.

jmromer:
> If we're putting external bikes in the database, I think it might make more sense to use single table inheritance for ExternalRegistryBike

What's the case for that? None of the `Bike` associations would apply for an `ExternalRegistryBike`, and similarly a `Bike` has no associated `external_registry` or, e.g., a `url` to store in the db — this suggested the lack of a true "is-a" relationship, so it's not obvious to me that inheritance makes sense here (despite the desirability of their interfaces mirroring each other), but I might be neglecting to consider something.

db/migrate/20190923181351_create_external_registries.rb
sethherr: Ahhh. But I think if we use the enums and classes we can achieve the same thing and don't need to use the database - also I think it better represents our data/architecture, since we will have a class for each client.

sethherr:
A big case for it is you could dry up `stop_heling_result.rb` and `verloren_of_gevonden_result.rb`

spec/rails_helper.rb
jmromer: This is only invoked once before / after an entire test run, but I can move it into the test suites that depend on it—it's a little more explicit. Updated in 8361b50b

app/models/external_registries/stop_heling_result.rb
jmromer: no prob! dropped in ccc4fa28

db/migrate/20190925162318_add_serial_normalized_to_external_registry_bike.rb
jmromer: Squashed in 3fce2c5c

app/serializers/external_registry_bike_v3_serializer.rb
jmromer: > I don't know that you need to include this delegate at all?

Why's that?

Here's what I'm seeing:

With:

with


Without:

without

jmromer: > Also, I don't know why date_stolen_string

Ah thanks -- that was to fix the date rendering in the search listing but `convertTime` should be handling that 👍 Updated in 8b5414b8

app/models/external_registries/stop_heling_result.rb
```diff
@@ -12,21 +12,15 @@ class StopHelingResult
       :location,
       :object,
       :registry_id,
-      :registry_name,
-      :registry_url,
       :source_name,
-      :source_unique_id,
-      :url
+      :source_unique_id
```
sethherr: There is no need for this attribute.

This should be `external_id`. The attribute that is being put into `external_id` in the PR right now should actually be stored in `additional_registration`

db/migrate/20190923181351_create_external_registries.rb
jmromer: >  But I think if we use the enums and classes we can achieve the same thing and don't need to use the database

Sure, but that would be denormalized data (still stored in the database), with no clear rationale for that denormalization. It also elides the associated data (should we have enums for external registry URLs and external registry countries too?)

You're confusing the `external_registries` table with the coupling of that table with client api classes. I see no reason to do away with the former. If we don't see a need for location-based selection of client classes, we can do away with the latter (which draws some inspiration from the way twitter accounts are currently handled—I'm surprised it's as illegible as it appears to be).

app/models/external_registries/stop_heling_result.rb
jmromer: Please elaborate

sethherr: The original intention of the attributes was that they would do for the purposes specified - where `external_id` is the unique source id of the registration.

The additional registration number, if we do need to store that, would be the displayed id if they have it.

jmromer: > external_id is the unique source id of the registration.

If I understand correctly, the `external_id` column should store an `external_registry_bike` record's uniquely identifying id in the associated bike registry (in this case stopheling.nl) — is that correct? 

That's what's implemented currently. (Hence the uniqueness validation for `external_id` scoped to `registry`)

`:source_unique_id` and `:source_name` are specific to stopheling.nl and refer to the source from the perspective of stopheling.nl (there are multiple sources for records within stopheling). I would think that one or both of those would be stored as `additional_registration` if we want to store them at all. Please advise if that's mistaken.

```
        "BronNaam" => "POLITIE BRABANT-NOORD",
        "BronUniekID" => "458171",
```

```
        "BronNaam" => "POLITIE KENNEMERLAND",
        "BronUniekID" => "83760",
```

sethherr: Ok. Sorry! I miss-understood.

PR #1292, opened 2019-09-27
Add STI for external registry bike

sethherr:


sethherr:
Currently this might not be enough to go on, I can continue work on it if it still doesn't seem convincing. Basic path:

- There are subclasses of `ExternalRegistryBike`, each one serves as a transformer of the unique attributes of the external registry into our registry attributes (ideally via functions rather than mutation)
- Integrations for each external registry create the external registry bikes

This isn't that different from what is occurring right now. The big difference is that you can leverage subclassing to reduce the amount of code that is shared between most of the inherited types (for instance, `initialize`). It also will make what is going on easier for someone who has dealt with STI in rails to figure out what is going on.

Let me know if you need more convincing @jmromer !

PR #1293, opened 2019-10-01
Add background search of Dutch registries for Dutch stolen bikes

jmromer:
- [x] Add daily checking for normalized serials of stolen dutch bikes in dutch registries (exact matches only)
- [x] Add `ExternalRegistryBike` to persist any found matches  STI for registry-specific sub-types
- [x] Add AR model subclasses `.build_from_api_response` method to handle doing what it says on the tin

```rb
Bike.currently_stolen_in(country: "NL").count
# Country Load (0.7ms)  SELECT  "countries".* FROM "countries" WHERE (name = 'NL' OR iso = 'NL') LIMIT 1
# (16.1ms)  SELECT COUNT(DISTINCT "bikes"."id") FROM "bikes" LEFT OUTER JOIN "stolen_records" ON "stolen_records"."id" = "bikes"."current_stolen_record_id" AND "stolen_records"."current" = $1 WHERE "bikes"."stolen" = $2 AND "bikes"."example" = $3 AND "bikes"."hidden" = $4 AND "bikes"."deleted_at" IS NULL AND ("bikes"."serial_number" != $5) AND "stolen_records"."country_id" = 154  [["current", "t"], ["stolen", "t"], ["example", "f"], ["hidden", "f"], ["serial_number", "unknown"]]
=> 85

EnqueueSearchesForExternalRegistryBikesWorker.new.perform
# Country Load (1.5ms)  SELECT  "countries".* FROM "countries" WHERE (name = 'NL' OR iso = 'NL') LIMIT 1
# (45.9ms)  SELECT "bikes"."serial_normalized" FROM "bikes" LEFT OUTER JOIN "stolen_records" ON "stolen_records"."id" = "bikes"."current_stolen_record_id" AND "stolen_records"."current" = $1 WHERE "bikes"."stolen" = $2 AND "bikes"."example" = $3 AND "bikes"."hidden" = $4 AND "bikes"."deleted_at" IS NULL AND ("bikes"."serial_number" != $5) AND "stolen_records"."country_id" = 154  [["current", "t"], ["stolen", "t"], ["example", "f"], ["hidden", "f"], ["serial_number", "unknown"]]

# Enqueued jobs complete in ~2.5 mins (~1.75s per search request)
```

Related:

https://github.com/bikeindex/bike_index/pull/1286
https://github.com/bikeindex/bike_index/pull/1287
https://github.com/bikeindex/bike_index/pull/1288
https://github.com/bikeindex/bike_index/pull/1292


jmromer:
@sethherr This is in good shape for some eyes. Total run time for the existing NL bikes is around ~3 minutes currently

app/assets/javascripts/init.coffee
```diff
@@ -114,6 +114,16 @@ class BikeIndex.Init extends BikeIndex
       return unless time.isValid
       $this.text(displayLocalDate(time, $this.hasClass("preciseTime")))
 
+    # Write local time (format: 1569955709)
+    $(".convertUnixTime").each ->
```
sethherr: Rather than doing this, the `convertTime` function in this file should be updated to match the function in [time_parser.js](https://github.com/bikeindex/bike_index/blob/master/app/javascript/packs/utils/time_parser.js) - which converts both timestamps and `ISO_8601`.

Sorry, should've pointed to this earlier

app/mailers/customer_mailer.rb
```diff
@@ -118,10 +118,13 @@ def bike_possibly_found_email(contact)
     @bike = contact.bike
     @user = User.fuzzy_email_find(@bike.owner_email)
     @match =
-      if contact.info_hash["match_type"] == "Bike"
+      case contact.info_hash["match_type"]
```
sethherr: 👍 switch to case here.

app/models/bike.rb
```diff
@@ -161,36 +161,69 @@ def organization(org_or_org_id)
...
-          .where.not(stolen: true)
-          .select(:serial_normalized)
-
       unscoped
         .current
         .stolen
-        .where.not(abandoned: true)
-        .where(serial_normalized: abandoned_serials)
+        .non_abandoned
+        .where(serial_normalized: abandoned.non_stolen.select(:serial_normalized))
```
sethherr: I think it would be better to `pluck` here

app/assets/javascripts/init.coffee
sethherr: Can you move this fix into a separate PR?

app/models/external_registries/external_registries.rb
```diff
@@ -0,0 +1,21 @@
+module ExternalRegistries
+  module ExternalRegistries
```
sethherr: why the double module?

app/workers/search_for_external_registry_bikes_worker.rb
```diff
@@ -0,0 +1,28 @@
+class SearchForExternalRegistryBikesWorker < ScheduledWorker
+  prepend ScheduledWorkerRecorder
```
sethherr: This is clever and I support doing it this way.

However, please do this in a separate PR, include it in `ScheduledWorker` itself, and remove all the calls to `record_scheduler_started` and `_finished` in the existing scheduled workers

db/migrate/20190923181352_create_external_registry_bikes.rb
```diff
@@ -0,0 +1,32 @@
+class CreateExternalRegistryBikes < ActiveRecord::Migration
+  def change
+    create_table :external_registry_bikes do |t|
+      t.string :external_registry_name, null: false, index: true
+      t.string :external_registry_url, null: false, index: true
+      t.integer :external_registry_country_id, null: false, index: true
```
sethherr: Not all external registries are country specific.

app/workers/search_for_external_registry_bikes_worker.rb
```diff
@@ -0,0 +1,28 @@
...
+  def self.frequency
+    24.hours
+  end
+
+  def perform
+    search_dutch_impounded_registries
+  end
+
+  # Search external registries. Results are persisted as ExternalRegistryBike records
+  def search_dutch_impounded_registries
```
sethherr: Please create separate sidekiq jobs for each serial. Sidekiq should manage the parallelism.

One pattern being used is to run a single search when the job is passed an argument, and enqueue the running of the jobs if no argument is passed.

lib/tasks/data_migrations.rake
```diff
@@ -1,70 +1,4 @@
 namespace :data do
-  namespace :migrate_info_hash do
```
sethherr: Please move to a separate PR

spec/models/external_registries/stop_heling_bike_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+
+module ExternalRegistries
+  RSpec.describe ExternalRegistries::StopHelingBike, type: :model do
+    let(:api_response) do
+      {}.tap do |h|
```
sethherr: Why tap here rather than just defining a hash?

spec/models/external_registries/verloren_of_gevonden_bike_spec.rb
```diff
@@ -0,0 +1,111 @@
...
+        it "returns the location with country name if one is available" do
+          expect(bike.location_found).to(eq <<~STR.chomp)
+            Pieter Steynstraat, Rest van de stad Zwolle - NL
+          STR
+        end
+      end
+    end
+
+    let(:non_bike_json) do
+      JSON.parse(<<~JSON.gsub(/\s/, " "))
```
sethherr: I don't believe you need to gsub here

app/workers/email_bike_possibly_found_notification_worker.rb
```diff
@@ -1,24 +1,25 @@
...
+
   def self.frequency
     24.hours
   end
 
   def perform
-    record_scheduler_started
-    notify_of_bike_index_held
-    record_scheduler_finished
+    Bike
```
sethherr: Similarly to the other worker, please separate this out into two things - one which finds the matches and one which sends messages to the matches.

Reason being - it's easier to test and it makes it more fault tolerant if a single message fails to send.

app/assets/javascripts/init.coffee
jmromer: Thanks, updated -- https://github.com/bikeindex/bike_index/pull/1294

app/models/bike.rb
jmromer: What would Benny Wong say? 🤔 
(I'm forced to guess at / interpolate context and reasoning here.)

Contra `pluck`: Because `select` evaluates to an AR relation, using it here triggers a single query overall. `pluck` triggers two, and is slower:

```rb
Bike.possibly_found
# (198.0ms)  SELECT COUNT(*) FROM "bikes" WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."deleted_at" IS NULL AND "bikes"."stolen" = $3 AND "bikes"."abandoned" = $4 AND "bikes"."serial_normalized" IN (SELECT "bikes"."serial_normalized" FROM "bikes" WHERE "bikes"."example" = $5 AND "bikes"."hidden" = $6 AND "bikes"."deleted_at" IS NULL AND "bikes"."abandoned" = $7 AND "bikes"."stolen" = $8  ORDER BY listing_order desc)  [["example", "f"], ["hidden", "f"], ["stolen", "t"], ["abandoned", "f"], ["example", "f"], ["hidden", "f"], ["abandoned", "t"], ["stolen", "f"]]
# Bike Load (130.2ms)  SELECT "bikes".* FROM "bikes" WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."deleted_at" IS NULL AND "bikes"."stolen" = $3 AND "bikes"."abandoned" = $4 AND "bikes"."serial_normalized" IN (SELECT "bikes"."serial_normalized" FROM "bikes" WHERE "bikes"."example" = $5 AND "bikes"."hidden" = $6 AND "bikes"."deleted_at" IS NULL AND "bikes"."abandoned" = $7 AND "bikes"."stolen" = $8  ORDER BY listing_order desc)  [["example", "f"], ["hidden", "f"], ["stolen", "t"], ["abandoned", "f"], ["example", "f"], ["hidden", "f"], ["abandoned", "t"], ["stolen", "f"]]

Bike.possibly_found_with_pluck
# (52.1ms)  SELECT "bikes"."serial_normalized" FROM "bikes" LEFT OUTER JOIN "colors" ON "colors"."id" = "bikes"."tertiary_frame_color_id" LEFT OUTER JOIN "colors" "secondary_frame_colors_bikes" ON "secondary_frame_colors_bikes"."id" = "bikes"."secondary_frame_color_id" LEFT OUTER JOIN "colors" "primary_frame_colors_bikes" ON "primary_frame_colors_bikes"."id" = "bikes"."primary_frame_color_id" LEFT OUTER JOIN "stolen_records" ON "stolen_records"."id" = "bikes"."current_stolen_record_id" AND "stolen_records"."current" = $1 WHERE "bikes"."example" = $2 AND "bikes"."hidden" = $3 AND "bikes"."deleted_at" IS NULL AND "bikes"."abandoned" = $4 AND "bikes"."stolen" = $5  ORDER BY listing_order desc  [["current", "t"], ["example", "f"], ["hidden", "f"], ["abandoned", "t"], ["stolen", "f"]]
# (57.2ms)  SELECT COUNT(*) FROM "bikes" WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."deleted_at" IS NULL AND "bikes"."stolen" = $3 AND "bikes"."abandoned" = $4 AND ("bikes"."serial_normalized" IN ('5NF5D12HG1615', 'F N GD6D255', '5520518', 'CM150520146', 'CM13G515132', '2595T0087D', 'P3DC26547', '1M33513', '6009619', '14H25430', 'U90421804', 'AJ30018231', '8243035130', 'WTU0181105', '5NF 10FP1387', 'W8K255322T', '1080600585', 'M15061152', 'F031102774F25', 'H819DD1158 FR80140', 'F96J98812', 'P2JA01627', '5NF3D16A86926', 'AAA18F00844', '81 12189576', 'W13435191', 'E3727910', '16A25772', 'CD515347017', 'W58C601025554K', '80C52222361U0419', '1CF6107761', 'AC16F11135', 'F002K20486', '1CC011F00315', 'X22997', 'P9EA611855', '5NX9514JA01264', 'F3P8J0806', 'HA8083061595C5N18') OR "bikes"."serial_normalized" IS NULL)  [["example", "f"], ["hidden", "f"], ["stolen", "t"], ["abandoned", "f"]]
# Bike Load (145.3ms)  SELECT "bikes".* FROM "bikes" WHERE "bikes"."example" = $1 AND "bikes"."hidden" = $2 AND "bikes"."deleted_at" IS NULL AND "bikes"."stolen" = $3 AND "bikes"."abandoned" = $4 AND ("bikes"."serial_normalized" IN ('5NF5D12HG1615', 'F N GD6D255', '5520518', 'CM150520146', 'CM13G515132', '2595T0087D', 'P3DC26547', '1M33513', '6009619', '14H25430', 'U90421804', 'AJ30018231', '8243035130', 'WTU0181105', '5NF 10FP1387', 'W8K255322T', '1080600585', 'M15061152', 'F031102774F25', 'H819DD1158 FR80140', 'F96J98812', 'P2JA01627', '5NF3D16A86926', 'AAA18F00844', '81 12189576', 'W13435191', 'E3727910', '16A25772', 'CD515347017', 'W58C601025554K', '80C52222361U0419', '1CF6107761', 'AC16F11135', 'F002K20486', '1CC011F00315', 'X22997', 'P9EA611855', '5NX9514JA01264', 'F3P8J0806', 'HA8083061595C5N18') OR "bikes"."serial_normalized" IS NULL)  [["example", "f"], ["hidden", "f"], ["stolen", "t"], ["abandoned", "f"]]
```

```rb
timeit { Bike.possibly_found }
# => 0.0008

timeit { Bike.possibly_found_with_pluck }
# => 0.0744

timeit { Bike.possibly_found.to_a }
# => 0.1422

timeit { Bike.possibly_found_with_pluck.to_a }
# => 0.7154
```

app/workers/search_for_external_registry_bikes_worker.rb
jmromer: > include it in ScheduledWorker itself

NB: That's not how `prepend` works — it has to be added to leaf / child classes to be inserted into the inheritance chain correctly.

```rb
UpdateExpiredInvoiceWorker.new.perform

#    6: def perform
#    7:   binding.pry
#    8:   # Invoice.should_expire.each { |invoice| invoice.update_attributes(updated_at: Time.current) # }

self.class.ancestors
# => [
#  [ 0] UpdateExpiredInvoiceWorker < ScheduledWorker,
#  [ 1] ScheduledWorkerRecorder,
#  [ 2] ScheduledWorker < ApplicationWorker,
#  [ 3] ApplicationWorker < Object,
```

lib/tasks/data_migrations.rake
jmromer: Updated in https://github.com/bikeindex/bike_index/pull/1296

spec/models/external_registries/stop_heling_bike_spec.rb
jmromer: `tap` is a sensible default because it's often useful to have each insertion happen in its own expression, in case there's logic surrounding the insertion. With a hash literal, it gets very cluttered and you end up cramming ternaries into the literal, for example. Updated in afe4daea

spec/models/external_registries/verloren_of_gevonden_bike_spec.rb
jmromer: Updated in 898bc23c

db/migrate/20190923181352_create_external_registry_bikes.rb
jmromer: Renamed to `country` in 48ece8a1
Should we anticipate bike records without a country? 🤔

app/models/external_registries/external_registries.rb
jmromer: Following naming conventions subject to the constraint of `ExternalRegistries` not living in `/app/models`. Constraint relaxed and namespaces flattened in af5a42df

app/models/bike.rb
sethherr: OH SNAP. Cool, I had no idea it evaluated to a single query, that's neat.

db/migrate/20190923181352_create_external_registry_bikes.rb
sethherr: Probably not? At least not for stolen records

app/models/external_registries/stop_heling_client.rb
```diff
@@ -4,11 +4,13 @@ module ExternalRegistries
...
     APP_ID = ENV["STOP_HELING_APP_ID"]
     API_KEY = ENV["STOP_HELING_API_KEY"]
+    TTL_HOURS = ENV.fetch("STOP_HELING_API_TTL_HOURS", 24).to_i.hours
 
-    attr_accessor :conn, :base_url
+    attr_accessor :conn, :base_url, :request_gap_seconds
 
     def initialize(base_url: nil)
       self.base_url = base_url || ENV["STOP_HELING_BASE_URL"]
+      self.request_gap_seconds = 1
```
sethherr: I am really concerned about this, and it being in the `v3` search path.

Do you share this concern? I suppose since they're unicorn processes they will get killed eventually...

sethherr: My personal preference would be to store the api rate limited status globally (perhaps in redis?), and making it so that failures automatically retry (hopefully leveraging sidekiq's retry).

So if we're running queries in parallel they don't keep hammering the API and cause a state where things are always sleeping and never succeeding at fetching new pages.

This might be a concern that is unnecessary at this stage, it's just something that has bit me before.

app/models/external_registries/verloren_of_gevonden_bike.rb
```diff
@@ -0,0 +1,89 @@
+module ExternalRegistries
+  class VerlorenOfGevondenBike < ExternalRegistryBike
```
sethherr: The issue with this not being STI is that there isn't a way to get back to this class from an instantiated `ExternalRegistryBike`.

If instead, this was STI, when you instantiated the `ExternalRegistryBike`, you would get back a `VerlorenOfGevondenBike`.

As of now, this class only has the `build_from_api_response`, which might be enough.

But what if we want to add a method to `VerlorenOfGevondenBike`s, not just another database attribute? It would free us up to have `external_registry_url` be a method rather than in the database.

spec/models/external_registries/stop_heling_bike_spec.rb
sethherr: Ok, I'm fine defaulting to tap

app/workers/email_bike_possibly_found_notification_worker.rb
jmromer: Updated in f2fca2fa

app/models/external_registries/verloren_of_gevonden_bike.rb
jmromer: Gotcha, thanks for elaborating. STI would not have occurred to me for this use case, so the intuition isn't quite there — the detail is helpful.

> As of now, this class only has the `build_from_api_response`

I'm not sure I see a way around a builder method since the ActiveRecord constructor can't accept arbitrary keys (and overriding it does not seem desirable) and we need to accommodate varying schema from different api response objects, but I'm open to suggestions.

app/models/external_registries/stop_heling_client.rb
jmromer: > I am really concerned about this, and it being in the v3 search path.

That sleep was added for the sake of pagination, but I don't think it needs to be there now that we're not paginating. I'll kill it.

app/models/external_registries/verloren_of_gevonden_bike.rb
jmromer: Updated in 817062aeac9

jmromer:
I'm a little worried about these individually-enqueued search jobs taking long and piling up as the workload increases. Doing some manual testing.

jmromer:
Re. 

> I'm a little worried about these individually-enqueued search jobs taking long and piling up as the workload increases. Doing some manual testing.

The [timeout](https://github.com/bikeindex/bike_index/pull/1293/commits/43b81437383aac78759211be87a012ad71b3a233) should take care of ☝️. Looks like my IP address might be blocked — I have to get on the VPN to get any results returned from VOG.

app/serializers/external_registry_bike_v3_serializer.rb
```diff
@@ -0,0 +1,63 @@
...
+  def frame_model
+    object.frame_model&.titleize
+  end
+
+  def frame_colors
+    object.frame_colors&.split(/\s*,\s*/)&.map(&:titleize)
+  end
+
+  def registry_name
+    object.external_registry_name
```
sethherr: Naming here makes me think - why not have the same attribute in the database that we use here?

Would it make more sense to just call the attribute `registry_name`?

app/serializers/external_registry_bike_v3_serializer.rb
```diff
@@ -0,0 +1,63 @@
...
+    :url,
+    to: :object,
+    allow_nil: true
+
+  def stolen
+    object.stolen?
+  end
+
+  def status
+    object.status&.titleize
```
sethherr: I don't think you need to declare this, if the only purpose is titleizing. I think you can skip this method. Ditto for `mnfg_name` and `frame_model`

app/serializers/external_registry_bike_v3_serializer.rb
sethherr: Ditto for `registry_url`

app/workers/enqueue_bike_possibly_found_notifications_worker.rb
```diff
@@ -0,0 +1,20 @@
...
+
+  def perform
+    bikes_with_matches.each do |bike, match|
+      EmailBikePossiblyFoundNotificationWorker
+        .perform_async(bike.id, match.class.to_s, match.id)
+    end
+  end
+
+  def bikes_with_matches
+    internal_matches = Bike.possibly_found_with_match
```
sethherr: Would it make sense to pluck here?

app/workers/enqueue_searches_for_external_registry_bikes_worker.rb
```diff
@@ -0,0 +1,15 @@
+class EnqueueSearchesForExternalRegistryBikesWorker < ScheduledWorker
```
sethherr: I think it's good to switch to these separate workers, rather than stuffing them both into the same one.

Can you make the class names the same, pluralization be damned - except for `Enqueue`? It will just make matching things easier for me. so `EnqueueSearchForExternalRegistryBikesWorker` and `EnqueueEmailBikePossiblyFoundNotificationsWorker`

app/models/external_registry/verloren_of_gevonden_bike.rb
```diff
@@ -0,0 +1,91 @@
...
+    class << self
+      def build_from_api_response(attrs = {})
+        is_bike = attrs["Category"] == "fiets"
+        return unless is_bike
+
+        description = attrs["Description"].presence
+
+        bike = find_or_initialize_by(
+          external_id: attrs["ObjectNumber"].presence,
+          type: "ExternalRegistry::VerlorenOfGevondenBike",
```
sethherr: I don't like this naming pattern because it doesn't seem accurate. I'd prefer it if it were `ExternalRegistryBike:: VerlorenOfGevondenBike` - since that what it's actually inheriting from. I'm happy with two different folders.

db/migrate/20190923181352_create_external_registry_bikes.rb
```diff
@@ -0,0 +1,26 @@
+class CreateExternalRegistryBikes < ActiveRecord::Migration
+  def change
+    create_table :external_registry_bikes do |t|
+      t.string :type, null: false, index: true
+      t.belongs_to :country, null: false, index: true
+
+      t.string :serial_number, null: false, index: true
```
sethherr: Do we actually need to index `serial_number`? I don't think we actually search on that column normally, do we?

app/serializers/external_registry_bike_v3_serializer.rb
jmromer: Makes sense. Updated in 5933eeb3

app/serializers/external_registry_bike_v3_serializer.rb
jmromer: > I don't think you need to declare this

In the sense of:
(a) the strings in question are already titleized, or 
(b) we shouldn't bother titleizing them for the UI?  🤔

app/workers/enqueue_searches_for_external_registry_bikes_worker.rb
jmromer: I dig it. Updated in 6e643789 and f43621fb

app/models/external_registry/verloren_of_gevonden_bike.rb
jmromer: Updated in 3526cb39 -- went with the pluralized form since the singular conflicts with the ActiveRecord class.

db/migrate/20190923181352_create_external_registry_bikes.rb
jmromer: We do when searching for exact matches:

```rb
# app/models/external_registry_bike.rb L15-17 (5933eeb3)

def find_or_search_registry_for(serial_number:)
  matches = ExternalRegistryBike.where(serial_number: serial_number)
  return matches if matches.any?
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/5933eeb3/app/models/external_registry_bike.rb#L15-L17)]</sup>

app/workers/enqueue_bike_possibly_found_notifications_worker.rb
jmromer: Strictly speaking we can't, because `Bike.possibly_found_with_match` returns an array of tuples of models, but we could hypothetically `pluck` the underlying attributes instead of instantiating models. I'm on the bubble about this. 

On the one hand, doing so might be 

(1) unintuitive since AR class methods typically return models, and
(2) an over-optimization at the expense of legibility, since we access a fair number of them: aside from `id`: `owner_email`, `serial_normalized`, and the class name,

but
 
(1a) unintuitiveness might ameliorated by renaming the methods (`possibly_found_ids_with_matches` e.g.)
(1b) the method returns an Array rather than an AR relation, so it's already departing from expectations
(2a) 4 attributes is arguably not a large number
(2b) We're feeding the results to workers, so the workers can query for the full AR models based on the class name and `id`

I'm comfortable refactoring if you think it makes sense to.

sethherr: Nah. Tuples are good enough. Sorry for trying to find plucks.

db/migrate/20190923181352_create_external_registry_bikes.rb
sethherr: We shouldn’t actually be doing this. Every search should be for the normalized serial. That’s one of our promises.

app/serializers/external_registry_bike_v3_serializer.rb
sethherr: Touché.

Are things capitalized automatically for the other serializer though? If not, it feels weird

db/migrate/20190923181352_create_external_registry_bikes.rb
sethherr: Sorry I didn’t catch this before!

app/serializers/external_registry_bike_v3_serializer.rb
jmromer: > Are things capitalized automatically for the other serializer though? If not, it feels weird

mm true. I think that suggests (c) — I'll move the titleizing to the React component.

db/migrate/20190923181352_create_external_registry_bikes.rb
jmromer: Good catch thanks -- updated in 7b012932

app/serializers/external_registry_bike_v3_serializer.rb
jmromer: Updated in bc2ab9ac

PR #1295, opened 2019-10-02
Add ScheduledWorkerRecorder

jmromer:
- Introduces `ScheduledWorkerRecorder`, which can be prepended to record start/stop of the worker's `perform` method

- `record_scheduler_started`/`record_scheduler_finished` can still be used where greater customization is needed

### DIFF
--- a/app/workers/scheduled_worker_recorder.rb
+++ b/app/workers/scheduled_worker_recorder.rb
@@ -1,0 +1,7 @@
+module ScheduledWorkerRecorder
+  def perform
+    record_scheduler_started
+    super
+    record_scheduler_finished
+  end
+end
--- a/app/workers/deactivate_expired_theft_alert_worker.rb
+++ b/app/workers/deactivate_expired_theft_alert_worker.rb
@@ -1,15 +1,13 @@
 class DeactivateExpiredTheftAlertWorker < ScheduledWorker
+  prepend ScheduledWorkerRecorder
+
   def self.frequency
     24.hours
   end
 
   def perform
-    record_scheduler_started
-
     TheftAlert
       .should_expire
       .update_all(status: TheftAlert.statuses[:inactive])
-
-    record_scheduler_finished
   end
 end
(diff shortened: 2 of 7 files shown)


app/workers/scheduled_worker_recorder.rb
```diff
@@ -0,0 +1,7 @@
+module ScheduledWorkerRecorder
```
sethherr: The reason I think it makes sense to include this by default in the `ScheduledWorker` class and only record changes if `arg`, `kwargs` or `block` are passed is because the scheduled runner isn't set up to pass arguments to scheduled jobs.

And I think this is reasonable - scheduled jobs shouldn't receive arguments when they're run regularly, and the record functionality should only happen when they are run from the scheduler

jmromer: Following up on the offline discussion: 

We could use the following to conditionally record / not record based on the presence / absence of arguments:

```rb
[*args, *kwargs.values].all?(&:blank?)
```

But this won't do, because scheduled workers that do double-duty as enqueuers and enqueued need recording to happen when arguments are passed, but "single-duty" scheduled jobs need to record when no arguments are passed. 

Checking the arity of the `perform` also doesn't help — it's variable.

I'll ship as-is, prepending `ScheduledWorkerRecorder` only where it applies.

sethherr: > But this won't do, because scheduled workers that do double-duty as enqueuers and enqueued need recording to happen when arguments are passed, but "single-duty" scheduled jobs need to record when no arguments are passed

I don't believe this is accurate.

"Double duty" workers only should record when no arguments are passed. "Single duty" workers should only record when no arguments are passed.

jmromer: > "Double duty" workers only should record when no arguments are passed.

D'oh. Thanks—I think I got turned around somewhere down the rabbit hole. 

Updated in https://github.com/bikeindex/bike_index/pull/1299

PR #1300, opened 2019-10-03
Update VOG image url

jmromer:
Updates `image_url` for VOG bikes (VOG has changed the endpoint from which it renders bike images).

Before:

Screen Shot 2019-10-03 at 10 53 59 AM

After:

Screen Shot 2019-10-03 at 10 52 18 AM

Hotfixing since this is broken on prod

### DIFF
--- a/app/javascript/packs/external_registry_search/components/ExternalRegistrySearchResult.js
+++ b/app/javascript/packs/external_registry_search/components/ExternalRegistrySearchResult.js
@@ -66,7 +66,6 @@ const ResultImage = ({ bike }) => {
 
   return (<a href={bike.url}
              style={backgroundStyles(bike)}
-             data-img-src={bike.thumb}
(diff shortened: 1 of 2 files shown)


app/javascript/packs/external_registry_search/components/ExternalRegistrySearchResult.js
```diff
@@ -66,7 +66,6 @@ const ResultImage = ({ bike }) => {
 
   return (<a href={bike.url}
              style={backgroundStyles(bike)}
-             data-img-src={bike.thumb}
```
jmromer: This should not have been deleted 👎

jmromer: https://github.com/bikeindex/bike_index/pull/1301

PR #1303, opened 2019-10-06
Render close serials asynchronously 

jmromer:
Refactors close-serials search results to render asynchronously on page load

Screen Shot 2019-10-06 at 5 27 19 PM

demo

### DIFF
--- a/app/models/stolen_record.rb
+++ b/app/models/stolen_record.rb
@@ -105,18 +105,20 @@ class StolenRecord < ActiveRecord::Base
     ].reject(&:blank?).join(", ")
   end
 
-  # Doesn't include street
-  def address_short
-    [city, state&.abbreviation, zipcode].reject(&:blank?).join(",")
-  end
+  # The stolen bike's general location (city and state / city and country if non-US)
+  # Include all available components (city, state, country) unconditionally if
+  # `include_all` is passed.
+  def address_location(include_all: false)
+    city_and_state =
+      [city&.titleize, state&.abbreviation&.upcase].reject(&:blank?).join(", ")
 
-  def address_location
-    return "" if state&.abbreviation.blank?
-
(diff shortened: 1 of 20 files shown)


app/serializers/bike_v2_serializer.rb
```diff
@@ -42,32 +41,43 @@ def title
...
-    end
+  end
+
+  def url
+    "#{ENV["BASE_URL"]}/bikes/#{object.id}"
   end
 
   def is_stock_img
-    object.public_images.present? ? false : object.stock_photo_url.present?
+    object.public_images.blank? && object.stock_photo_url.present?
```
sethherr: 👍

app/serializers/bike_v2_serializer.rb
```diff
@@ -42,32 +41,43 @@ def title
...
   end
 
   def stolen_location
-    object.current_stolen_record && object.current_stolen_record.address_short
+    stolen_record = object.current_stolen_record
```
sethherr: Why this change? What is wrong with the `address_short` method? If you're just adding the country here, why not add it to the address_short method?

jmromer: acfadad2

app/models/stolen_record.rb
```diff
@@ -105,18 +105,20 @@ def address(skip_default_country: false, override_show_address: false)
     ].reject(&:blank?).join(", ")
   end
 
-  # Doesn't include street
-  def address_short
-    [city, state&.abbreviation, zipcode].reject(&:blank?).join(",")
```
jmromer: The only call sites for `StolenRecord#address_short` are in the v2 stolen record and bike serializers. This commit refactors to use `address_location` instead, which better accounts for international locations.

PR #1305, opened 2019-10-07
Externalize strings for BikeController

jmromer:
Externalize strings for BikeController

### DIFF
--- a/app/controllers/concerns/controller_helpers.rb
+++ b/app/controllers/concerns/controller_helpers.rb
@@ -104,9 +104,9 @@ module ControllerHelpers
     end
   end
 
-  def translation(key, scope: nil)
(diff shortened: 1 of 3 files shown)


app/controllers/concerns/controller_helpers.rb
```diff
@@ -104,9 +104,9 @@ def return_to_if_present
     end
   end
 
-  def translation(key, scope: nil)
+  def translation(key, scope: nil, **kwargs)
```
sethherr: Can you still pass in the `scope` keyword argument here?

jmromer: yup — any other kwargs are captured by `**kwargs`

PR #1315, opened 2019-10-08
Consolidate Country.fuzzy find and Country.fuzzy_iso_find

jmromer:
- Optimizes `fuzzy_find` by performing the search in a single query
- Increases module depth
- There are only two call sites for `fuzzy_iso_find`

### DIFF
--- a/app/models/country.rb
+++ b/app/models/country.rb
@@ -13,8 +13,8 @@ class Country < ActiveRecord::Base
 
   def self.fuzzy_find(name_or_iso)
     name_or_iso = name_or_iso.to_s.strip.downcase
-    name_or_iso = "us" if name_or_iso == "usa"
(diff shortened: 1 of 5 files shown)


app/models/country.rb
```diff
@@ -11,14 +11,12 @@ def self.select_options
     end
   end
 
-  def self.fuzzy_find(str)
-    return nil unless str.present?
-    fuzzy_iso_find(str) || where("lower(name) = ?", str.downcase.strip).first
-  end
+  def self.fuzzy_find(name_or_iso)
+    name_or_iso = name_or_iso.to_s.strip.downcase
+    name_or_iso = "us" if name_or_iso == "usa"
```
sethherr: Might make sense to just return `united_states` in this situation.

PR #1319, opened 2019-10-10
Externalize controller flash strings

jmromer:
- Externalizes flash strings from ~half our controllers

- Deepens `ControllerHelpers#translation` so it infers the method invoking it rather than the controller action in which it's invoked. This is useful because a private helper method or before action can be invoked in the context of multiple controller actions but we need a 1:1 mapping from call site to translation file key. 

```rb
# app/controllers/concerns/controller_helpers.rb L108-120 (c60a89ea)

def translation(key, scope: nil, controller_method: nil, **kwargs)
  if scope.blank? && controller_method.blank?
    controller_method =
      caller_locations
        .slice(0, 2)
        .map(&:label)
        .reject { |label| label =~ /rescue in/ }
        .first
  end

  scope ||= [:controllers, controller_namespace, controller_name, controller_method.to_sym]
  I18n.t(key, kwargs, scope: scope.compact)
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/c60a89ea/app/controllers/concerns/controller_helpers.rb#L108-L120)]</sup>

This allows the method to be invoked more often without passing the scope:

```rb
# app/controllers/bike_codes_controller.rb L22-30 (5b97a9e6)

protected

# . . .

def find_bike_code
  unless current_user.present?
    flash[:error] = translation(:must_be_signed_in)
# . . .
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/5b97a9e6/app/controllers/bike_codes_controller.rb#L22-L30)]</sup>

Instead of 
```rb
def find_bike_code
  unless current_user.present?
    flash[:error] = translation(:must_be_signed_in, scope: %i[controllers bike_codes find_bike_code])
```

Tracking: #963

### DIFF
--- a/app/controllers/bikes_controller.rb
+++ b/app/controllers/bikes_controller.rb
@@ -16,7 +16,7 @@ class BikesController < ApplicationController
     @interpreted_params = Bike.searchable_interpreted_params(permitted_search_params, ip: forwarded_ip_address)
     @stolenness = @interpreted_params[:stolenness]
     if params[:stolenness] == "proximity" && @stolenness != "proximity"
-      flash[:info] = t(:we_dont_know_location, location: params[:location])
+      flash[:info] = translation(:we_dont_know_location, location: params[:location])
     end
     @bikes = Bike.search(@interpreted_params).page(params[:page] || 1).per(params[:per_page] || 10).decorate
     @selected_query_items_options = Bike.selected_query_items_options(@interpreted_params)
@@ -62,7 +62,7 @@ class BikesController < ApplicationController
   def scanned
     @bike_code = BikeCode.lookup_with_fallback(scanned_id, organization_id: params[:organization_id], user: current_user)
     if @bike_code.blank?
-      flash[:error] = t(:unable_to_find_sticker, scanned_id: params[:scanned_id])
+      flash[:error] = translation(:unable_to_find_sticker, scanned_id: params[:scanned_id])
       redirect_to user_root_url
     elsif @bike_code.bike.present?
       redirect_to bike_url(@bike_code.bike_id, scanned_id: params[:scanned_id], organization_id: params[:organization_id]) and return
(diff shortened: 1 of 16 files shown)


app/controllers/bikes_controller.rb
```diff
@@ -266,14 +266,14 @@ def theft_templates
...
     {}.with_indifferent_access.tap do |h|
-      h[:bike_details] = translation(:bike_details, scope: %i[controllers bikes edit])
-      h[:photos] = translation(:photos, scope: %i[controllers bikes edit])
-      h[:drivetrain] = translation(:drivetrain, scope: %i[controllers bikes edit])
-      h[:accessories] = translation(:accessories, scope: %i[controllers bikes edit])
-      h[:ownership] = translation(:ownership, scope: %i[controllers bikes edit])
-      h[:groups] = translation(:groups, scope: %i[controllers bikes edit])
-      h[:remove] = translation(:remove, scope: %i[controllers bikes edit])
-      h[:report_stolen] = translation(:report_stolen, scope: %i[controllers bikes edit]) unless @bike.stolen?
+      h[:bike_details] = translation(:bike_details, controller_method: :edit)
```
sethherr: Does this really change? This is the template name, is it not? Shouldn't this be translated somewhere else?

jmromer: > Does this really change?

taking `this` to mean the change in keyword arg used from `scope:` to `controller_method:`:

It doesn't necessarily need to change, but I think it's preferable to prefer the `controller_method` kwarg where possible — doing so is DRYer and has smaller scope for error.

> This is the template name, is it not?

taking `this` to mean `h[:bike_details] = <value>`:

The hash key is the template name and the value is a title string rendered in the template. The hash overall provides the navigation menu.

> Shouldn't this be translated somewhere else?

taking `this` to mean all the translations in `bike_templates` and `theft_templates`:

These hashes couple filenames to query params to rendered strings to the navigation menu, so I'd rather they weren't defined at all. But, conditional on this abstraction, translating these strings at their origin point seems simpler than doing so everywhere they're used in templates. I'm open to refactoring it though — concretely, we could map from template names to translation keys instead of mapping to strings.

sethherr: I was being stupid. It all makes sense now!

PR #1323, opened 2019-10-11
Add Project529 integration

jmromer:
- Adds Project 529 client and bikes classes
- Adds STI external registry credential class to manage rotating credentials
- Moves stopheling creds to database
- Adds a background job to fetch and save any 529 bikes updated over the past day

Screen Shot 2019-10-11 at 5 45 25 PM


app/models/external_registry_client/project529_client.rb
```diff
@@ -0,0 +1,82 @@
+class ExternalRegistryClient::Project529Client
+  BASE_URL = ENV.fetch("PROJECT_529_BASE_URL", "https://project529.com/garage")
+  APP_ID = ENV["PROJECT_529_APP_ID"]
```
sethherr: Does it make sense to specify this in the credential model instead? It might make viewing history more useful.

app/models/external_registry_credential.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  # Set an error and return false if access_token is not yet expired.
+  def access_token_can_be_reset?
+    return true unless access_token_expires_at.present? && access_token_valid?
+    errors.add(:access_token, "not expired")
+    false
+  end
+
+  def api
```
sethherr: I'm not confident that every single external registry client **must** have this method, so I don't know if this is necessary.

app/models/external_registry_client/project529_client.rb
jmromer: Ah thanks -- should have removed that

PR #1348, opened 2019-10-17
Fix asset compilation on CircleCI 

jmromer:
The builds on `master` are failing because assets built pre-#1339 are cached and the cache key is unaffected by changes to yarn.lock. 

This patch:

- Adds an env-settable `CACHE_VERSION` value to cache keys

- Resolves a warning about implicit core-js version in babelrc 42636e7

```
% bundle exec rake assets:precompile
# . . .

Compiling…
Compilation failed:

WARNING: We noticed you're using the `useBuiltIns` option without declaring a
core-js version. Currently, we assume version 2.x when no version is passed.
Since this default version will likely change in future versions of Babel, we
recommend explicitly setting the core-js version you are using via the `corejs`
option.

You should also be sure that the version you pass to the `corejs` option matches
the version specified in your `package.json`'s `dependencies` section. If it
doesn't, you need to run one of the following commands:

  npm install --save core-js@2    npm install --save core-js@3
  yarn add core-js@2              yarn add core-js@3


Hash: eb5d2737ca9a5c4c7970
Version: webpack 4.41.1
Time: 83071ms
Built at: 10/16/2019 9:14:23 PM

# . . .
```

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -8,7 +8,8 @@
           "browsers": "> 1%",
           "node": "current"
         },
-        "useBuiltIns": "usage"
+        "useBuiltIns": "usage",
+        "corejs": "2"
       }
     ],
     "@babel/react"
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -37,7 +37,7 @@ jobs:
             DOCKERIZE_VERSION: v0.6.1
 
       - restore_cache:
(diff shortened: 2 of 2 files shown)


jmromer:
Merging to fix https://circleci.com/gh/bikeindex/bike_index/8198#tests/containers/1

PR #1352, opened 2019-10-20
Localize stripe payment templates, TheftAlertPlan models

jmromer:
-  Localizes promoted alert purchase template
- Adds `TheftAlertPlan#language` attribute
- Consolidates language-selection helpers in `language_choices`
- Adds `in_language` scopes to `TheftAlertPlan` and `Blog`
- Adds `language` to promoted alert plan admin form and table
- Introduces `Localizable` concern

### Before

Screen Shot 2019-10-18 at 9 50 39 AM

Screen Shot 2019-10-18 at 9 50 52 AM

### After

Screen Shot 2019-10-18 at 10 53 25 AM
Screen Shot 2019-10-18 at 10 52 49 AM

Screen Shot 2019-10-21 at 10 48 55 AM


app/models/theft_alert_plan.rb
```diff
@@ -1,6 +1,12 @@
 class TheftAlertPlan < ActiveRecord::Base
   include Amountable
 
+  LANGUAGE_ENUM = {
```
sethherr: Do you think it makes sense to share this with `Blog`? Even if it's as simple as doing `enum language: Blog::LANGUAGE_ENUM`, I think that would be better than having multiple hashes that are the same thing and have the opportunity to become out of sync. But maybe that doesn't make sense!

jmromer: Good call thanks -- updated in 32879040

jmromer:
@sethherr Just a head's up: c4d6b48 updates to make payments in the localized currency. Let me know if we don't want to do that yet. If we keep it, we'll probably want to add `currency` to the remaining `Amountable` models.

PR #1355, opened 2019-10-21
Translate welcome index recovery stories

jmromer:
- Adds translation of the recovery stories displayed on welcome_index and welcome_recovery_stories.

### Before

Screen Shot 2019-10-21 at 11 01 33 AM

### After

Screen Shot 2019-10-21 at 3 17 13 PM

Screen Shot 2019-10-21 at 3 19 09 PM


jmromer:
This was 90% an experiment that panned out better than expected (from a technical pov not necessarily wrt the translations), so I'm cool with not shipping it if the translations will look less polished than the alternative.

That said, I did a sanity check with Spanish and the translations look decent. I'm not competent to judge the Dutch ones though.

sethherr:
I think if you put a little "translated from #{language name}" or "original comment in #{language name}" somewhere (when the story has been translated), I would be very happy with this.

Two potential places: 

- underneath _bike recovered xxx_
- in the lower right of the quote box

sethherr:
Damn, current version is great! Good work.

PR #1362, opened 2019-10-22
Add currency column to Amountable models

jmromer:
- Adds a `currency` column to `Invoice`,  `PaidFeature`, `TheftAlertPlan`
- Drops `Amountable#currency` method


app/models/concerns/localizable.rb
```diff
@@ -0,0 +1,13 @@
+# Add to a model that can be localized independent of the locale but with the
+# same set of languages (e.g. News, TheftAlertPlan models).
+#
+# Expects a `language` column defined on the corresponding table.
+module Localizable
+  extend ActiveSupport::Concern
+
+  included do
+    enum language: I18n.available_locales.sort.map.with_index.to_h.freeze
```
sethherr: Beautiful! I love this concern.

Except - it makes accessing the actual value of the enum a little difficult - which might need to happen, especially in this version of Rails.

I _think_ putting this in a class method resolves it?

If you think this is a non-issue, or you have a different solution, let me know!

app/views/bikes/edit_alert_purchase.html.haml
```diff
@@ -11,12 +11,18 @@
               -# highlight the same nav menu item
               = @edit_templates.select { |k, v| @edit_template.match(/#{k}/) }.values.first
 
+          - currency = t(I18n.locale, scope: %i[money currencies])
```
sethherr: I think it might make sense to put this into a helper, a la `theft_alert_plan_title`.

app/helpers/application_helper.rb
```diff
@@ -225,18 +225,18 @@ def pretty_print_json(data)
...
+  # Language choices available for request locale and localized models (Blog,
+  # TheftAlertPlan).
+  #
+  # Return an array of tuples, each with the following strings:
+  #
   # [<localized language name>, <language key (from Blog::LANGUAGE_ENUM)>]
-  def blog_languages
-    @blog_languages ||=
-      Blog.languages.map { |lang, _| [t(lang, scope: [:locales]), lang] }
+  def language_choices
```
sethherr: Should we create a localization helper? `TheftAlertPlanHelper` seems lonely

app/models/concerns/localizable.rb
jmromer: > it makes accessing the actual value of the enum a little difficult

Do you mean the enum hash? The Rails inflector helps out with that, but let me know if you had something else in mind.

```ruby
TheftAlertPlan.languages
=> {
  "en" => 0,
  "nl" => 1
}
```

app/views/bikes/edit_alert_purchase.html.haml
jmromer: Good call thanks -- updated in 1d39916e

app/models/concerns/localizable.rb
sethherr: OH SNAP. Yes, that, it's a non-issue 😊

PR #1369, opened 2019-10-24
Automate TwitterAccount creation

jmromer:
- Adds "check credentials" button to twitter account show page
- Adds omniauth-twitter gem
- Replaces manual `TwitterAccount` creation with automated flow using omniauth login


app/views/admin/twitter_accounts/index.html.haml
```diff
@@ -2,7 +2,7 @@
   .col-sm-9.col-md-10
     %h1 Twitter Accounts
   .col-sm-3.col-md-2
-    = link_to "New Account", new_admin_twitter_account_path, class: "btn btn-success float-right"
+    = link_to "New Account", "/auth/twitter", class: "btn btn-success float-right"
```
sethherr: What makes this route?

config/routes.rb
```diff
@@ -279,6 +281,7 @@
   get "manufacturers_tsv", to: "manufacturers#tsv"
 
   resource :integrations, only: [:create]
+  get "/auth/twitter/callback", to: "admin/twitter_accounts#create"
```
sethherr: I like moving the integration to admin rather than handling it in the integrations controller, good call.

app/views/admin/twitter_accounts/index.html.haml
jmromer: The `OmniAuth::Builder` block in  `config/initializers/omniauth.rb` -- they don't show up in the routes listing, but a `/auth/:provider` route gets defined for each provider.

PR #1372, opened 2019-10-25
Fix undefined variable currency

jmromer:
Fixes https://app.honeybadger.io/projects/35931/faults/56314792


jmromer:
Hotfixing -- this is blocking promoted alert purchases

PR #1378, opened 2019-10-28
Add canonical_url to blog posts

jmromer:
- Adds `canonical_url` to blog model
- Renders canonical url in the header for blog posts
- Adds field to blog edit form

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -228,4 +228,11 @@ module ApplicationHelper
   def bike_placeholder_image_path
     image_path("revised/bike_photo_placeholder.svg")
   end
+
+  # Return the canonical url for the given blog post.
+  # If none available, default to the bike index url, removing any query params.
+  def canonical_url(blog)
+    url = blog.canonical_url.presence || news_url(blog)
+    url.split("?").first
+  end
 end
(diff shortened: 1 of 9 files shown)


app/helpers/application_helper.rb
```diff
@@ -228,4 +228,11 @@ def pretty_print_json(data)
   def bike_placeholder_image_path
     image_path("revised/bike_photo_placeholder.svg")
   end
+
+  # Return the canonical url for the given blog post.
+  # If none available, default to the bike index url, removing any query params.
+  def canonical_url(blog)
+    url = blog.canonical_url.presence || news_url(blog)
+    url.split("?").first
```
sethherr: Why are there query parameters?

app/views/layouts/application.html.haml
```diff
@@ -14,6 +14,7 @@
       window.BikeIndex.translator = (keyspace) => {
         return (key, args={}) => I18n.t(`javascript.${keyspace}.${key}`, args);
       }
+    = yield :header
```
sethherr: Could you add a comment in here about why this is here? (e.g. `see news/show for example of content_for :header`)

I think this is the correct way to solve this problem, but I worry that coming back to this in a few months, it will be unclear why this yield is here.

app/helpers/application_helper.rb
jmromer: That mainly accounts for the locale param — it might be overkill, since I'm not sure if a crawler could trigger that.

app/views/layouts/application.html.haml
jmromer: No prob!

PR #1382, opened 2019-10-28
Don't discard non-normalized serial number when querying by serial

jmromer:
Currently we replace a user's serial number query with its normalized form, but this results in counterintuitive behavior when searching external registries (a search for a non-normalized serial will always return no results even if results matching the non-normalized serial are present).

~This patch sets `window.interpreted_params.serial` to the originally submitted serial number, and sets the normalized form as `serial_normalized`.~

This patch sets `window.interpreted_params.raw_serial` to the originally submitted serial number, and retains the normalized form as `serial`. (see 40e6ce934)


sethherr:
After thinking about this a little - I think the optimal solution to this would be searching external registries for both for the exact serial entered and also for the normalized serial. I don't know how possible this is.

jmromer:
Makes sense — I'm testing d31eb2e to see how it perfoms

app/models/concerns/bike_searchable.rb
```diff
@@ -14,8 +14,14 @@ module ClassMethods
     #   distance: distance in miles for matches. Only for stolenness == 'proximity'
     #   bounding_box: bounding box generated by geocoder. Only for stolenness == 'proximity'
     def searchable_interpreted_params(query_params, ip:)
-      serial = SerialNormalizer.new(serial: query_params[:serial]) if query_params[:serial].present?
-      (serial ? { serial: serial.normalized } : {}) # serial if present
+      params = {}
```
sethherr: Why is this here?

sethherr: aka, why not add this stuff to `query_params`?

jmromer: Mainly following suit with the original code, which builds a new hash. 
I see merit in that approach though, since mutating a function parameter feel like an act of hostility.

PR #1386, opened 2019-10-29
Automate twitter accounts credential refresh

jmromer:
- Adds a "log in to refresh credentials" button to the admin/twitter_accounts#show template
- Updates the credentials on any pre-existing record and re-verifies credentials to clear errors

### DIFF
--- a/app/views/admin/twitter_accounts/show.html.haml
+++ b/app/views/admin/twitter_accounts/show.html.haml
@@ -3,7 +3,6 @@
     .mr-4
       = link_to "Edit", edit_admin_twitter_account_url(@twitter_account), class: "btn btn-success"
       = link_to "Check Credentials", check_credentials_admin_twitter_account_url(@twitter_account), class: "btn btn-success"
-      = link_to "Log in to refresh credentials", "/auth/twitter", class: "btn btn-success"
 
 - if @twitter_account.errored?
   .row.mt-4.mb-4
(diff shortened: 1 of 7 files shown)


jmromer:
@sethherr All set 👍

app/views/admin/twitter_accounts/show.html.haml
```diff
@@ -1,8 +1,9 @@
 .admin-subnav
   .col-6
     .mr-4
-      = link_to "Edit", edit_admin_twitter_account_url(@twitter_account), class: "btn btn-lg btn-success"
-      = link_to "Check Credentials", check_credentials_admin_twitter_account_url(@twitter_account), class: "btn btn-lg btn-success"
+      = link_to "Edit", edit_admin_twitter_account_url(@twitter_account), class: "btn btn-success"
+      = link_to "Check Credentials", check_credentials_admin_twitter_account_url(@twitter_account), class: "btn btn-success"
+      = link_to "Log in to refresh credentials", "/auth/twitter", class: "btn btn-success"
```
sethherr: Rather than putting this on the show page, could you put it on the edit page? I think it might make sense to put in the account credentials box.

Maybe it makes sense to have in both places, but it definitely makes sense to have on the page next to the credentials.

jmromer: No prob -- updated in eb0518ee

Screen Shot 2019-10-29 at 4 13 03 PM
Screen Shot 2019-10-29 at 4 13 12 PM
Screen Shot 2019-10-29 at 4 13 19 PM

PR #1401, opened 2019-11-07
Notify Admin if stolen bike with promoted alert is recovered

jmromer:
When marking a bike as recovered, send an admin notification if the bike has an active promoted alert.

Screen Shot 2019-11-07 at 12 46 23 PM

Small UI fix (render the CTA as a button rather than a link):

Screen Shot 2019-11-07 at 12 40 28 PM


app/controllers/bikes/recovery_controller.rb
```diff
@@ -16,7 +16,8 @@ def update
         flash[:success] = translation(:bike_recovered)
         redirect_to bike_path(@bike)
       else
-        render :edit, bike_id: @bike.id, token: params[:token]
+        session[:recovery_link_token] = params[:token]
```
sethherr: Hot take: SICK! Good addition, makes life better for everyone.

app/mailers/admin_mailer.rb
```diff
@@ -41,11 +41,20 @@ def lightspeed_notification_email(organization, api_key)
...
 
-  def theft_alert_purchased(theft_alert)
+  def theft_alert_purchased(theft_alert, notify_of_recovery: false)
     @theft_alert = theft_alert
     @theft_alert_plan = theft_alert.theft_alert_plan
     @creator = theft_alert.creator
     @bike = theft_alert.bike
-    mail(to: "stolenbikealerts@bikeindex.org", subject: "Theft Alert purchased: #{@theft_alert.id}")
+
+    if notify_of_recovery
```
sethherr: This confused me at first, that both emails where in the same mailer method

sethherr: I would like it to be made clearer, somehow?

I think being in the same method makes sense. Idea for how to resolve:  mailer method was renamed `theft_alert_notification`, second arg was `notification_type: "purchased"`

My

app/controllers/bikes/recovery_controller.rb
sethherr: Oh. This was something that should've been there already. Oops, my bad. Any way you can add a test?

app/mailers/admin_mailer.rb
jmromer: Good call thanks — updated in 1998c27d

PR #1403, opened 2019-11-11
Add regional organization associations with geocoding, bike lat/long

jmromer:
- Adds regional organization associations to `Organization`, geocodes by the first location's coordinates
- ~Adds admin UI to set regional field values on `Organization` records~
- Adds lat/long and city columns to `Bikes`, sets bike location info on registration

### Post-deploy

```
bin/rake data:set_bike_location_info 
```

### Add admin regional organization affiliations UI

Screen Shot 2019-11-07 at 11 01 00 AM
Screen Shot 2019-11-07 at 11 01 06 AM

Screen Shot 2019-11-06 at 1 55 07 PM


app/models/organization.rb
```diff
@@ -310,4 +342,11 @@ def set_ambassador_organization_defaults
     self.ascend_name = nil
     self.parent_organization_id = nil
   end
+
+  def not_both_regional_organization_and_suborganization
+    if regional? && regional_organization_id.present?
+      errors.add(:regional, :cannot_be_regional_and_belong_to_one)
+      errors.add(:regional_organization, :cannot_be_regional_and_belong_to_one)
```
sethherr: I don't think you need to add this error to both? I think adding to both will make the error message more confusing.

I think it makes more sense to just add it to `regional_organization`.

app/views/admin/organizations/_form.html.haml
```diff
@@ -52,8 +52,26 @@
...
+      = f.text_field :city, class: "form-control"
+    .form-group
+      = f.label :state
+      = f.collection_select(:state_id, State.all, :id, :name,
+        { include_blank: true, prompt: "Select a state" },
+        { class: "form-control" })
+    .form-group
+      = f.label :zipcode
+      = f.text_field :zipcode, class: "form-control"
+    .form-group
```
sethherr: Could you add `fancy-select` class to this `form-group`?

app/views/admin/organizations/_form.html.haml
```diff
@@ -73,7 +91,19 @@
     .form-group
       = f.check_box :approved
       = f.label :approved, 'Validated & approved'
+
+    .form-group
```
sethherr: I know this is additional complexity, but I think I would prefer if `regional` was actually a `paid_feature_slug`, and the regional organizations were just the organizations that fall within the organization search radius.

So you would add the search_radius, but otherwise, you wouldn't need to select the organizations and update them. This seems like a less labor intensive process, and less potentially error prone. It also means we'll focus on adding locations to organizations, rather than adding regional organizations (which is what I'd prefer was focused on).

I would either add another column to the db like `child_ids`, or rename `child_ids`, and switch it to a structure like `suborganizations: { child_ids: [], regional_ids: [] }`. Or something similar!

sethherr: Potentially related to https://github.com/bikeindex/bike_index/pull/1404#discussion_r345238664 - if we're associating based on geocoding and locations, it might make sense to skip adding the address inputs to the organization - and instead just geocode based on the first location for the organization (with some note on the organization form around "search radius" that makes it explicit what is happening).

Just an idea, not a requirement.

jmromer: Makes sense — updated in 2656c690
What's the intent with adding a `regional_ids` column? (as opposed to querying for those orgs, that is)

sethherr: Just a guess that the geospatial query to select the regional organizations was an expensive query. Probably doesn't matter, caching can be resolved later.

app/models/organization.rb
```diff
@@ -181,6 +188,20 @@ def bike_shop_display_integration_alert?
     bike_shop? && %w[no_pos broken_pos].include?(pos_kind)
   end
 
+  def bikes_in_region_counts
+    return unless regional?
+    return @bikes_in_region_counts if defined?(@bikes_in_region_counts)
+
+    bikes_in_orgs = regional_suborganizations.includes(:bikes).flat_map(&:bikes).map(&:id)
```
sethherr: I would like `bikes_in_orgs` and `bikes_in_region` as separately queryable active record collections, because the eventual goal here is to add graphs, time series, etc, which will use scopes to display counts for specific conditions.

app/models/organization.rb
```diff
@@ -61,10 +64,14 @@ class Organization < ActiveRecord::Base
   scope :approved, -> { where(is_suspended: false, approved: true) }
   # Eventually there will be other actions beside organization_messages, but for now it's just messages
   scope :bike_actions, -> { where("paid_feature_slugs ?| array[:keys]", keys: %w[messages unstolen_notifications impound_bikes]) }
+  scope :regional, -> { where(regional: true) }
```
sethherr: Should this be removed?

app/models/location.rb
```diff
@@ -40,14 +41,24 @@ def set_phone
     self.phone = Phonifyer.phonify(self.phone) if self.phone
   end
 
+  def set_organization_coordinates
```
sethherr: Things I'm worried about - Location updating, organization doesn't update coordinates
- Organization has multiple locations, it's unclear which one sets the coordinates for the organization.

Does putting this method on `organization` instead of location make more sense? If we're already bumping the `updated_at`, seems like this would fit well in the `set_calculated_attributes` method on Organization. And also would make sense to utilize the `search_location` method.

app/models/organization.rb
```diff
@@ -293,6 +314,27 @@ def update_associations
...
+    location_latitude.present? && location_longitude.present?
+  end
+
+  def regional_suborganizations
+    return Organization.none unless regional? && coordinates_set?
+
+    address = search_location&.address
+    return Organization.none unless address.present?
+
+    Organization.near(address, search_radius).where.not(id: id)
```
sethherr: Why use address here and not `[latitude, longitude]`? I believe Geocoder will just geocode that location and query based on the lat, long anyway...

db/migrate/20191106210313_add_location_fields_to_organization.rb
```diff
@@ -0,0 +1,12 @@
+class AddLocationFieldsToOrganization < ActiveRecord::Migration
+  def change
+    change_table :organizations do |t|
```
sethherr: Is this syntax roll-backable?

spec/support/geocoder_default_location.rb
```diff
@@ -38,7 +39,7 @@
   let(:bounding_box) { [39.989124784445764, -74.96065051723293, 41.43644261555424, -73.05123208276707] }
 
   before do
-    Geocoder.configure(lookup: :test)
+    Geocoder.configure(lookup: :test, ip_lookup: :test)
```
sethherr: Oooh. Nifty, I didn't know about this. This might fix some broken/untested things.

app/views/bikes/edit_bike_details.html.haml
```diff
@@ -92,6 +92,13 @@
             .form-well-input
               = f.text_field :zipcode, class: 'form-control', required: true, placeholder: t(".zipcode")
 
+          .form-group.row.unnested-field.no-divider-row
```
sethherr: Should city be added here too, or is it going to be set behind the scenes?

sethherr: Also, could you add the `fancy-select` class?

app/models/organization.rb
jmromer: Makes sense — I think that'll save an API call

db/migrate/20191106210313_add_location_fields_to_organization.rb
jmromer: It is — can't find it in the docs but simple enough to verify

sethherr: Cool, I learn new things every day!

jmromer:
@sethherr Query methods added in 3fab3ee8

Let me know if any of this UI copy is vague or confusing — 

Screen Shot 2019-11-15 at 3 34 15 PM

app/models/bike.rb
```diff
@@ -433,7 +433,9 @@ def visible_by(passed_user = nil)
   end
 
   def find_current_stolen_record
-    stolen_records.last if stolen_records.any?
+    return unless stolen_records.any?
+    self.current_stolen_record = stolen_records.last
+    current_stolen_record
```
sethherr: Nit: is `current_stolen_record` necessary? Won't the previous line return that anyway?

app/models/bike.rb
```diff
@@ -625,6 +627,55 @@ def registration_location
...
+  end
+
+  # Set the bike's location (city, postal code, and country)
+  # based in the following order of precedence:
+  # 1. Set explicitly on the bike
+  # 2. From the current stolen record, if one is present
+  # 3. From the creation organization, if one is present
+  # 4. From the bike owner's address
+  # 5. From the request's IP address
+  def set_location_info(request_location: nil)
```
sethherr: I'm not sure this is currently the case, and if it is I think it's something we should solve in a future PR - but I want to note it now.

Something I just realized we ought to do is make it so if a user removes the location from their bike, we respect that and don't set the location via other methods.

app/models/location.rb
```diff
@@ -45,9 +42,10 @@ def org_location_id
   end
 
   def update_organization
-    # Because we need to update the organization and make sure it is shown on the map correctly
-    # Manually update to ensure that it runs the before save stuff
-    organization && organization.update_attributes(updated_at: Time.current)
+    # Because we need to update the organization and make sure it is shown on
+    # the map correctly, manually update to ensure that it runs the before save
+    # stuff
```
sethherr: nit: probably ought to just delete this hanging stuff now, it looks silly.

app/models/twitter_account.rb
```diff
@@ -1,7 +1,9 @@
 class TwitterAccount < ActiveRecord::Base
+  include Geocodeable
+
   has_many :tweets, dependent: :destroy
 
-  attr_accessor :no_geocode
+  attr_accessor :skip_geocoding
```
sethherr: Can remove?

app/models/organization.rb
```diff
@@ -181,6 +186,36 @@ def bike_shop_display_integration_alert?
...
+    return Bike.none unless regional? && search_coordinates_set?
+    Bike.near(search_coordinates, search_radius).reorder(id: :asc)
+  end
+
+  # Bikes nearby not associated with any nearby organizations.
+  def bikes_nearby_unaffiliated
+    bikes_nearby.where.not(id: bikes_in_nearby_organizations.pluck(:id))
+  end
+
+  def bikes_in_region_counts
```
sethherr: Nit: Why put this belongs in the model rather than controller?

sethherr:
The UI looks great. We'll definitely iterate on it, it's currently readable, 🚢 🎉

PR #1404, opened 2019-11-11
Add bikes geocoding

jmromer:
- Adds geocoding for bikes
- Provides `skip_geocoding` attribute to set lat/long directly in tests
- Gives precedence to current stolen geolocation data, falls back on bike / org / owner / request data

```rb
# app/models/bike.rb L49-50 (deb6c69c)

geocoded_by :geo_data
after_validation :geocode, if: :should_be_geocoded?
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/deb6c69c/app/models/bike.rb#L49-L50)]</sup>

```rb
# app/models/bike.rb L800-802 (deb6c69c)

def should_be_geocoded?
  !skip_geocoding? && geo_data.present?
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/deb6c69c/app/models/bike.rb#L800-L802)]</sup>

```rb
# app/models/bike.rb L780-794 (deb6c69c)

# Geolocate based on the full current stolen record address, if available.
# Otherwise, use the data set by set_location_info.
def geo_data
  return @geo_data if defined?(@geo_data)
  set_location_info

  @geo_data =
    current_stolen_record
      &.address(override_show_address: true)
      .presence ||
    [city, zipcode, country&.name]
      .select(&:present?)
      .join(" ")
      .presence
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/deb6c69c/app/models/bike.rb#L780-L794)]</sup>

### DIFF
--- a/app/models/bike.rb
+++ b/app/models/bike.rb
@@ -46,8 +46,8 @@ class Bike < ActiveRecord::Base
   accepts_nested_attributes_for :stolen_records
   accepts_nested_attributes_for :components, allow_destroy: true
 
-  geocoded_by nil, latitude: :stolen_lat, longitude: :stolen_long
-  after_validation :geocode, if: lambda { false } # Never geocode, it's from stolen_record
+  geocoded_by :geo_data
+  after_validation :geocode, unless: :skip_geocoding
 
   validates_presence_of :serial_number
   validates_presence_of :propulsion_type
@@ -60,7 +60,8 @@ class Bike < ActiveRecord::Base
   attr_accessor :other_listing_urls, :date_stolen, :receive_notifications, :has_no_serial, # has_no_serial included because legacy b_params, delete 2019-12
                 :image, :b_param_id, :embeded, :embeded_extended, :paint_name,
                 :bike_image_cache, :send_email, :marked_user_hidden, :marked_user_unhidden,
-                :b_param_id_token, :address, :address_city, :address_state, :address_zipcode
+                :b_param_id_token, :address, :address_city, :address_state, :address_zipcode,
+                :skip_geocoding
 
   attr_writer :phone, :user_name, :organization_affiliation, :external_image_urls # reading is managed by a method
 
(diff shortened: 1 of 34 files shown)


app/models/organization.rb
```diff
@@ -75,7 +75,7 @@ class Organization < ActiveRecord::Base
   after_commit :update_associations
 
   geocoded_by :search_location
-  after_validation :geocode
+  after_validation :geocode, if: ->(org) { org.search_location.present? }
```
sethherr: We still need them geocoded for the map (`/where`)

app/models/bike.rb
```diff
@@ -752,4 +756,12 @@ def cycle_type_name
   def propulsion_type_name
     PropulsionType.new(propulsion_type).name
   end
+
+  def geo_data
+    stolen_record_location = find_current_stolen_record&.address(override_show_address: true)
```
sethherr: I would like the current stolen record location to override the location info, if it's set and the location info isn't

app/models/organization.rb
sethherr: Maybe not, because that's locations? But I still think we want it.

app/models/twitter_account.rb
```diff
@@ -1,7 +1,7 @@
 class TwitterAccount < ActiveRecord::Base
   has_many :tweets, dependent: :destroy
 
-  attr_accessor :no_geocode
+  attr_accessor :skip_geocoding
```
sethherr: Good. Nice consistency.

app/models/organization.rb
jmromer: The `if` here is meant to skip the geocoding API call if `search_location` isn't set, since we won't be able to geolocate without it: https://github.com/alexreisner/geocoder#avoiding-unnecessary-api-requests

I added some documentation and a check for the component location fields having been changed in 35588ce3 -- let me know if I'm missing something though.

app/models/bike.rb
jmromer: Updated in deb6c69c

PR #1406, opened 2019-11-12
Extract Geocodeable mixin

jmromer:
- Adds a `Geocodable` concern to reduce some duplication and consistify across geocoded models.  
- Expects `geocode_data` to be defined on the including model

### DIFF
--- a/app/models/concerns/geocodeable.rb
+++ b/app/models/concerns/geocodeable.rb
@@ -1,0 +1,34 @@
+# Add to a model that can be geolocated
+# Expects `latitude` and `longitude` columns to be defined.
+module Geocodeable
+  extend ActiveSupport::Concern
+
+  included do
+    geocoded_by :geocode_data
+    after_validation :geocode, if: :should_be_geocoded?
+
+    attr_accessor :skip_geocoding
+
+    def skip_geocoding?
+      !!skip_geocoding
+    end
+
+    # Override in Geocodeable model.
+    def geocode_data; end
+
+    # An array of symbols of the db columns upon which `geocode_data`
+    # depends.These are checked for changes before geocoding using
+    # `should_be_geocoded?`. By default, empty, so we don't check for changes.
+    def geocode_columns; []; end
+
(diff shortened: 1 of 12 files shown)


spec/factories/locations.rb
```diff
@@ -9,5 +9,6 @@
     street { "foo address" }
     latitude { 41.9282162 }
     longitude { -87.6327552 }
+    skip_geocoding { true }
```
sethherr: Nice, clean addition.

app/models/concerns/geocodeable.rb
```diff
@@ -0,0 +1,34 @@
...
+    after_validation :geocode, if: :should_be_geocoded?
+
+    attr_accessor :skip_geocoding
+
+    def skip_geocoding?
+      !!skip_geocoding
+    end
+
+    # Override in Geocodeable model.
+    def geocode_data; end
```
sethherr: Would it make more sense to default to `@geocode_data ||= address`, since most of the models do that?

app/models/mail_snippet.rb
```diff
@@ -50,4 +50,21 @@ def update_organization
...
+
+  def geocode_data
+    address
+  end
+
+  def geocode_columns
+    %i[address]
+  end
+
+  def should_be_geocoded?
```
sethherr: Not sure if this is an improvement, also not sure if it's possible, but can you use `super` here to use the concerns `should_be_geocoded?` ?

app/models/concerns/geocodeable.rb
jmromer: Good catch thanks -- updated in 0ab79e2e

app/models/mail_snippet.rb
jmromer: > not sure if it's possible, but can you use super here to use the concerns should_be_geocoded?

Negative. With concerns, the model methods overwrite rather than override — there's no inheritance to leverage.

PR #1417, opened 2019-11-20
Fix request location zipcode access

jmromer:
[Request geocoding](https://github.com/alexreisner/geocoder#custom-result-handling) exposes a `postal_code` attribute rather than `zipcode`.

Fixes https://app.honeybadger.io/projects/35931/faults/57560267


jmromer:
Makes sense — I'll ship that separately

PR #1418, opened 2019-11-20
Link directly to external registry bike

jmromer:
When notifying a bike owner of a potential match found in an external registry, link directly to the external registry url.

Screen Shot 2019-11-20 at 11 43 38 AM
Screen Shot 2019-11-20 at 11 43 30 AM


app/views/customer_mailer/bike_possibly_found_email.text.haml
```diff
@@ -1,8 +1,10 @@
 = t(".text.we_think_your_stolen_bike", bike_title: @bike.title_string)
```
sethherr: Could you add the name of the external registry (in bold or italic) to this sentence?

PR #1421, opened 2019-11-20
Organization dashboard improvements

sethherr:


sethherr:
Add some graphing:

Screen Shot 2019-11-20 at 11 08 28 AM

app/models/organization.rb
```diff
@@ -128,6 +128,12 @@ def parent?; child_ids.present? end
 
   def child_organizations; Organization.where(id: child_ids) end
 
+  def nearby_organizations
+    return self.class.none unless regional? && search_coordinates_set?
+    self.class.where.not(id: child_ids).near(search_radius).reorder(id: :asc)
+    # nearbys(search_radius).where.not(id: child_ids)
```
sethherr: This makes tests pass but then fails when actually loading the dashboard page. Any suggestions @jmromer ?

jmromer: Can't repro yet — what kind of a failure are you seeing? (both L133 and L134 work for me)

Screen Shot 2019-11-20 at 2 55 47 PM

jmromer: @sethherr 
The following works for me:
```rb
    nearbys(search_radius).where.not(id: child_ids).reorder(id: :asc)
```

The `.reorder(id: :asc)` is needed because by default geocoder tries to order by `distance`, which isn't present in this case.

I'm also seeing `self.class.where.not(id: child_ids).near(search_radius).reorder(id: :asc)` work too though.

sethherr: Thanks!

sethherr: See comment https://github.com/bikeindex/bike_index/pull/1424#issuecomment-557053391

PR #1423, opened 2019-11-21
Drop Vagrant support

jmromer:
Drops Vagrant support (no longer supported?)


realchrisolin:
I haven't been keeping a constant eye on PRs, but I haven't seen an alternative come through (e.g. Docker) and this still works. Why are we killing it again?

sethherr:
Does the vagrant setup currently work?

jmromer:
> this still works

👍

realchrisolin:
Even though this is closed, I figure it's worth this is a topic worth hashing out. The Vagrant environment was created because it's somewhat unintuitive for someone unfamiliar with the BI codebase to set up the postgres database correctly, full stack devs notwithstanding. I know that Vagrant is becoming less and less common in favor of Docker these days, so maybe there's a valid argument for moving off of Vagrant?

I'm not much of a (Ruby) dev, even after spending nearly a year trying to pick up the language/fundamentals. I'm happy to pick up the slack elsewhere if there's a need.

jmromer:
> I know that Vagrant is becoming less and less common in favor of Docker these days, so maybe there's a valid argument for moving off of Vagrant?

I think migrating to Docker makes sense but, given resource constraints, I also think migrating _from_ Vagrant only makes sense if someone in the OSS community contributes the Docker setup. If the Vagrant setup is still functional, I think it makes sense to leave it for now.

PR #1424, opened 2019-11-21
Include current_organization in organization dashboard counts

sethherr:
Follows #1421

### DIFF
--- a/app/controllers/admin/dashboard_controller.rb
+++ b/app/controllers/admin/dashboard_controller.rb
@@ -1,5 +1,6 @@
 class Admin::DashboardController < Admin::BaseController
   def index
+    set_period # graphing set up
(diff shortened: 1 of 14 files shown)


sethherr:
Switched to bounding box, as per the [Geocoder documentation](https://www.rubydoc.info/gems/geocoder/1.3.7) (under _Advanced Querying_):

> This can also dramatically improve query performance, especially when used in conjunction with indexes on the latitude/longitude columns. Note, however, that returned results do not include `distance` and bearing `attributes`. Note that `#near` performs both bounding box and radius queries for speed.

We don't actually need the distance or bearing attributes - and we also want this to be performant. Switching to `bounding_box` fixed an issue I was having rendering the graph, and seems to speed up rendering by a non-trivial amount - but I haven't done rigorous testing on this.

Issue #1080, opened 2019-07-25
Rename BikeCode


sethherr:
There are only about 10 spokecards that still exist. There are about 100k QR stickers that exist, and there continue to be more added.

To reduce confusion BikeCode should be renamed to something that references the fact that they're stickers (`BikeSticker`? `IdSticker`?) and anything referencing spokecode should be removed up (I believe the spokecards can still work after this, they just don't need to be called spokecards)

PR #1426, opened 2019-11-21
Upgrade Rails

jmromer:
Upgrades Rails to 5.2.4, along with concomitant upgrades to key dependencies.

### Notable changes

- Sprockets upgraded to v4
- Webpacker and Webpack upgraded to v4
- Doorkeeper upgraded to v5
- Spring dropped in favor of bootsnap (included by default with Rails 5)
- core-js upgraded to v3

Deprecations and interface changes are pointed out in comments below.

#### Rails 4.2 -> 5.0
- https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-4-2-to-rails-5-0
- https://hashrocket.com/blog/posts/how-to-upgrade-to-rails-5

#### Doorkeeper changes
- https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions#from-4x-to-5x

#### Rails 5.0 -> 5.1
- https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-0-to-rails-5-1

#### Rails 5.1 -> 5.2
- https://guides.rubyonrails.org/upgrading_ruby_on_rails.html#upgrading-from-rails-5-1-to-rails-5-2

#### Webpacker 4 upgrade guide

- https://github.com/rails/webpacker/blob/master/docs/v4-upgrade.md


.babelrc
```diff
@@ -1,28 +0,0 @@
-{
-  "presets": [
```
jmromer: `.babelrc` settings now live in `babel.config.js`

.circleci/config.yml
```diff
@@ -31,110 +32,182 @@ jobs:
...
+
+      - type: cache-save
+        name: "Dockerize: cache save"
+        key: cache-{{ .Environment.CACHE_VERSION }}-dockerize-0.6.1
+        paths:
+          - ~/.cache/dockerize
+
+
+      - type: cache-restore
+        name: "Git: cache restore"
```
jmromer: The updates to the CircleCI config add descriptive labels to caching steps, concentrate cache directories in `~/.cache`, and caches dependencies more effectively.

README.markdown
```diff
@@ -50,7 +50,6 @@ This explanation assumes you're familiar with developing Ruby on Rails applicati
 
 | Toggle in development | command                      | default  |
 | ---------             | -------                      | -------  |
-| Spring                | `bin/rake dev:spring`        | disabled |
```
jmromer: This PR drops Spring since [bootsnap](https://blog.bigbinary.com/2018/01/01/rails-5-2-adds-bootsnap-to-the-app-to-speed-up-boot-time.html) replaces it.

app/controllers/admin/organizations_controller.rb
```diff
@@ -106,7 +106,7 @@ def sortable_columns
   end
 
   def kind_for_organizations
-    # Legacy enum issue so excited for TODO: Rails 5 update
+    # TODO: Refactor to incorporate Rails 5 enum changes
     Organization::KIND_ENUM[params[:search_kind].to_sym] || 0
```
jmromer: Question: 

What's the issue alluded to here?

app/controllers/admin/recoveries_controller.rb
```diff
@@ -47,7 +47,7 @@ def sortable_columns
 
   def recovery_display_status_searched
     return StolenRecord::RECOVERY_DISPLAY_STATUS_ENUM.values if params[:search_recovery_display_status] == "all"
-    # Legacy enum issue so excited for TODO: Rails 5 update
+    # TODO: Refactor to incorporate Rails 5 enum changes
     recovery_display_status_parameter = (params[:search_recovery_display_status] || "waiting_on_decision").to_sym
```
jmromer: Same as above — not sure what issue is being referenced here

app/controllers/api/v2/bikes_search.rb
```diff
@@ -102,10 +102,10 @@ def set_proximity
           use :stolen_search
         end
         get "/count", root: "bikes", each_serializer: BikeV2Serializer do
-          { "declared_params" => declared(params, include_missing: false) }
+          # { "declared_params" => declared(params, include_missing: false) }  # TODO: what is this line doing?
```
jmromer: I think this line can be removed?

app/controllers/api/v3/organizations.rb
```diff
@@ -44,14 +44,14 @@ def allowed_write_organizations
 
           permitted = declared(params, include_missing: false)
           organization = Organization.new(
-            permitted.to_hash.except("locations").merge(auto_user_id: current_user.id)
+            permitted.to_h.except("locations").merge(auto_user_id: current_user.id)
```
jmromer: Calling `to_hash` on `params` is deprecated in Rails because params no longer inherit from `Hash`. `to_h` is defined, however.

app/controllers/application_controller.rb
```diff
@@ -34,7 +34,7 @@ def cors_preflight_check
       headers["Access-Control-Allow-Methods"] = "POST, GET, OPTIONS"
       headers["Access-Control-Allow-Headers"] = "*"
       headers["Access-Control-Max-Age"] = "1728000"
-      render text: "", content_type: "text/plain"
+      render plain: ""
```
jmromer: `render plain: ` replaces `render text:`

app/controllers/application_controller.rb
```diff
@@ -79,7 +79,7 @@ def requested_locale
   end
 
   # Around filter to ensure locale (language and timezone) are set only per request
-  def set_locale
+  def set_locale(&action)
```
jmromer: The Rails guides have been updated to account for the possibility of locale setting leaking across requests. This change updates `set_locale` to match the suggested method signature.

app/controllers/bikes_controller.rb
```diff
@@ -341,7 +341,7 @@ def ensure_user_allowed_to_edit
   end
 
   def update_organizations_can_edit_claimed(bike, organization_ids)
-    organization_ids = (organization_ids || []).map(&:to_i)
+    organization_ids = organization_ids.presence.to_a.map(&:to_i)
```
jmromer: `nil.to_a` returns `[ ]`

app/controllers/feedbacks_controller.rb
```diff
@@ -10,24 +10,27 @@ def create
...
-        redirect_to :back
+        redirect_back(fallback_location: help_path)
       else
         redirect_to help_path
       end
     else
       @page_errors = @feedback.errors
+      render :index and return if request.referer.blank?
+
       re_path = Rails.application.routes.recognize_path(request.referer)
```
jmromer: `Rails.application.routes.recognize_path` now raises an exception if passed `nil`

app/controllers/public_images_controller.rb
```diff
@@ -36,7 +36,7 @@ def edit
 
   def is_private
     @public_image.update_attribute :is_private, params[:is_private]
-    render nothing: true
+    head :ok
```
jmromer: `render nothing: true` deprecated in favor of `head`

app/helpers/money_helper.rb
```diff
@@ -4,6 +4,8 @@ module MoneyHelper
 
   # TODO: Add a bank implementation that fetches conversion rate values
   Money.default_bank.add_rate(:USD, :EUR, 0.88)
+  Money.rounding_mode = BigDecimal::ROUND_HALF_UP
+  Money.locale_backend = :i18n
```
jmromer: Setting these values explicitly to silence deprecation warnings.

app/javascript/packs/application.js
```diff
@@ -1 +1,4 @@
+import "core-js/stable";
+import "regenerator-runtime/runtime";
+
```
jmromer: see: [Webpacker 4 usage guide](https://github.com/rails/webpacker#usage)

app/models/bike_sticker_batch.rb
```diff
@@ -31,8 +31,8 @@ def code_number_length_or_default
 
   # Shouldn't occur anymore, but included for legacy diagnostic purposes
   def duplicated_integers
-    bike_code_integers.map do |int|
-      next unless bike_code_integers.count(int) > 1
+    bike_sticker_integers.map do |int|
```
jmromer: Renaming for consistency with the renaming introduced in #1080

app/models/export.rb
```diff
@@ -17,7 +17,8 @@ class Export < ApplicationRecord
 
   before_validation :set_calculated_attributes
 
-  attr_accessor :timezone, :avery_export # permit assignment
```
jmromer: `avery_export=` defined twice

app/models/feedback.rb
```diff
@@ -21,7 +21,7 @@ def self.bike(bike_or_bike_id = nil)
 
   def self.feedback_types
     # Quick semi-hack to get a list of types, good enough till it's not ;)
-    @feedback_types ||= pluck("DISTINCT feedback_type").reject(&:blank?)
+    @feedback_types ||= pluck(Arel.sql("DISTINCT feedback_type")).reject(&:blank?)
```
jmromer: In Rails 5, raw sql emits a warning unless wrapped in `Arel.sql()`

app/models/handlebar_type.rb
```diff
@@ -2,7 +2,7 @@ class HandlebarType
   include Enumable
 
   SLUGS = {
-    drop: 5,
+    drop_bar: 5,
```
jmromer: `drop` conflicts with a built-in method definition, so this renames the slug to `drop_bar`. Not wedded to it if another term would be preferable

app/models/organization.rb
```diff
@@ -29,21 +29,24 @@ class Organization < ApplicationRecord
...
+
   has_many :memberships, dependent: :destroy
-  has_many :mail_snippets
   has_many :users, through: :memberships
+
+  has_many :creation_states
+  has_many :created_bikes, through: :creation_states, source: :bike
+
+  has_many :locations, inverse_of: :organization, dependent: :destroy
+  has_many :mail_snippets
```
jmromer: The order of definition seems to matter for these now? This change groups associations so dependencies are defined when dependent associations are defined.

app/models/feedback.rb
sethherr: I believe there is a pluck distinct method

app/services/bike_creator_builder.rb
```diff
@@ -20,7 +20,14 @@ def add_required_attributes(bike)
   end
 
   def new_bike
-    bike = Bike.new(@b_param.bike.except(*BParam.skipped_bike_attrs))
+    bike_attrs =
+      @b_param
+        .bike
+        .map { |k, v| [k, v.presence] }
```
jmromer: Ensure no nils or blank strings are passed

app/services/bike_finder.rb
```diff
@@ -18,7 +18,7 @@ def self.find_matching(serial:, owner_email:)
         .joins("LEFT JOIN user_emails ON user_emails.user_id = users.id")
         .where("users.email = ? OR user_emails.email = ?", email, email)
         .select(:id)
-        .uniq
+        .distinct
```
jmromer: AR `uniq` is deprecated in favor of `distinct`

app/services/bike_searcher.rb
```diff
@@ -148,7 +148,7 @@ def fuzzy_find_serial
     bike_ids = fuzzy_find_serial_ids
     # Don't return exact matches
     bike_ids = bike_ids.uniq - matching_serial.map(&:id)
-    Bike.where("id in (?)", bike_ids)
+    Bike.where(id: bike_ids)
```
jmromer: Replacing `id in (?)` with `id: ` here and below resolves SQL generation errors caused by ambiguous table bugs.

app/uploaders/pdf_uploader.rb
```diff
@@ -1,8 +1,5 @@
 class PdfUploader < ApplicationUploader
   include ::CarrierWave::Backgrounder::Delay
-  include CarrierWave::MimeTypes
-
-  process :set_content_type
```
jmromer: The need for these lines is obviated by the CarrierwaveUploader upgrade.

app/views/admin/payments/index.html.haml
```diff
@@ -11,7 +11,7 @@
 
 :ruby
   payments_counts = Payment.kinds.map do |kind|
-    # Legacy enum issue so excited for TODO: Rails 5 update
+    # TODO: Refactor to incorporate Rails 5 enum changes
     payment_enum = Payment::KIND_ENUM[kind.to_sym]
```
jmromer: Not sure what issue is being referenced here

bin/rails
```diff
@@ -1,9 +1,4 @@
 #!/usr/bin/env ruby
-begin
-  load File.expand_path('../spring', __FILE__)
-rescue LoadError => e
-  raise unless e.message.include?('spring')
-end
-APP_PATH = File.expand_path('../../config/application', __FILE__)
+APP_PATH = File.expand_path('../config/application', __dir__)
```
jmromer: Remove spring binstubs

config/initializers/action_controller_parameters_shim.rb
```diff
@@ -0,0 +1,11 @@
...
+  # DEPRECATION WARNING: #to_hash unexpectedly ignores parameter filtering, and
+  # will change to enforce it in Rails 5.1.
+  # Enable # `raise_on_unfiltered_parameters` to respect parameter filtering, which is the
+  # default in new applications.
+  # For the existing deprecated behaviour, call #to_unsafe_h instead.
+
+  def to_hash
+    to_unsafe_h
+  end
+end
```
jmromer: TODO: Probably want to remove this and fix whatever failures ensue.

db/migrate/20180624192035_add_deleted_at_to_bike_organizations.rb
```diff
@@ -1,4 +1,4 @@
-class AddDeletedAtToBikeOrganizations < ActiveRecord::Migration
+class AddDeletedAtToBikeOrganizations < ActiveRecord::Migration[4.2]
```
jmromer: Rails 5 requires the migration version number to be appended to the migration class name

db/migrate/20191216054404_add_confidential_to_applications.rb
```diff
@@ -0,0 +1,13 @@
...
+
+class AddConfidentialToApplications < ActiveRecord::Migration[5.0]
+  def change
+    add_column(
+      :oauth_applications,
+      :confidential,
+      :boolean,
+      null: false,
+      default: false, # Default to false since we're using public clients
+    )
```
jmromer: > Doorkeeper::Application now has a new boolean column named confidential that is true by default and has NOT NULL CONSTRAINT. This column is required to allow creating Public & Private Clients as mentioned in Section 8.5 of draft-ietf-oauth-native-apps-12 of OAuth 2 RFC which was previously unavailable. If you are migrating from the Doorkeeper <= 5.0, then you can easily add this column by generating a proper migration file using the following command: rails g doorkeeper:confidential_applications.

> [IMPORTANT]: all the applications (clients) starting from 5.0 / 4.4.x releases are considered as private by default. You need to manually change confidential column to false if you are using public clients, otherwise your mobile or other applications will not be able to authorize. See #1142 for more details.

https://github.com/doorkeeper-gem/doorkeeper/wiki/Migration-from-old-versions#database-changes-1

spec/controllers/admin/ambassador_tasks_controller_spec.rb
```diff
@@ -26,7 +26,7 @@
 
     describe "#edit" do
       it "renders the edit template with the found ambassador task" do
-        get :edit, id: FactoryBot.create(:ambassador_task).id
+        get :edit, params: { id: FactoryBot.create(:ambassador_task).id }
```
jmromer: Pre-Rails 5:
```rb
    post "/widgets", :widget => {:name => "My Widget"}
```

Post-Rails 5:

```rb
    post "/widgets", :params => { :widget => {:name => "My Widget"} }
```
https://relishapp.com/rspec/rspec-rails/docs/request-specs/request-spec#specify-managing-a-widget-with-rails-integration-methods

spec/controllers/admin/bulk_imports_controller_spec.rb
```diff
@@ -10,7 +10,7 @@
     let!(:bulk_import) { FactoryBot.create(:bulk_import) }
     it "renders" do
       get :index
-      expect(response).to be_success
+      expect(response).to be_ok
```
jmromer: The `successful?` predicate method is deprecated.

spec/controllers/admin/organizations/invoices_controller_spec.rb
```diff
@@ -70,21 +68,32 @@
...
-          expect(invoice.active?).to be_truthy
-          expect(organization.is_paid).to be_truthy
-          expect(organization.paid_feature_slugs).to eq(%w[abandoned_bike_messages passwordless_users])
-          expect(invoice.paid_feature_ids).to match_array([paid_feature1.id, paid_feature2.id])
-          expect(invoice.amount_due).to eq 0
-          # TimeParser isn't storing records perfectly - for now, just ignoring since fix can be separate
-          expect(invoice.subscription_start_at.to_i).to be_within(1.day).of 1536202800
-          expect(invoice.subscription_end_at.to_i).to be_within(1.day).of 1536202800 + 2.years
-          expect(invoice.notes).to eq params[:notes]
+          Sidekiq::Testing.inline! do
```
jmromer: Execute this test synchronously in order to test after_commit callbacks without manually triggering them.

Gemfile
```diff
@@ -1,12 +1,20 @@
...
-gem "rails", "4.2.11"
+gem "rack", "2.0.8"
+gem "rails", "5.2.4"
 
 gem "active_model_serializers", "~> 0.9.3"
 gem "bcrypt", "~> 3.1.7"
+gem "bootsnap"
 gem "jquery-rails"
-gem "pg"
+gem "pg", "~> 1.0"
```
jmromer: pg is pinned to `~> 1.0` here to fix / minimize breaking interface changes. Same with the dependencies below.

Gemfile
```diff
@@ -69,36 +77,35 @@ gem "axlsx", "~> 3.0.0.pre" # Write Excel files (OrganizationExports), on pre b/
...
+gem "swagger-ui_rails", github: "bikeindex/swagger-ui_rails", branch: "bike_index_0.1.7"
 gem "wine_bouncer"
 
 # Secure things
 gem "rack-throttle"
 gem "secure_headers", "~> 2.5.0"
 
 # Frontend
-gem "bootstrap", "~> 4.0.0.alpha3" # Bootstrap 4 - used for revised stylesheets
+gem "bootstrap", "4.0.0.alpha4" # Bootstrap 4 - used for revised stylesheets
```
sethherr: Note (so nobody needs to check what up again) - this locks to the the current version in the Gemfile.lock - it does not change the gem version.

app/controllers/admin/organizations_controller.rb
sethherr: I am pretty sure you can just use the string, without having to cast it to the actual integer.

So this line can become slightly simpler and a lot more readable:

```ruby
Organization.pos_kinds.include? ? params[:search_kind] : "no_pos"
```

And the line above this can become `return %w[no_pos does_not_need_pos]`

These updates don't need to be addressed in this PR. I can deal with them afterward. It's mainly stuff that will make things more readable (at least from my perspective)

app/controllers/admin/recoveries_controller.rb
sethherr: Answered above, and as stated above, I'd prefer to merge this PR and then deal with those updates, since they don't block shipping this (and fewer changes is better than more changes).

Sorry that the string wasn't perfectly consistent, I was figuring I'd just be grepping through, it's nice that you've made it consistent.

app/controllers/api/v2/bikes_search.rb
```diff
@@ -44,7 +44,6 @@ def set_proximity
           use :search_bikes
         end
         get "/", root: "bikes", each_serializer: BikeV2Serializer do
-          { "declared_params" => declared(params, include_missing: false) }
```
sethherr: I think it was critical that `include_missing` was passed here - but since this didn't break any tests, I'm assuming that an update made this no longer relevant. Or perhaps, include missing wasn't actually necessary.

app/views/layouts/doorkeeper.html.haml
```diff
@@ -5,7 +5,7 @@
     = stylesheet_link_tag 'revised', media: 'all'
     <link href='https://fonts.googleapis.com/css?family=Open+Sans:400,400italic,300italic,700' rel='stylesheet' type='text/css'>
     <link href='https://fonts.googleapis.com/css?family=Montserrat:400,700' rel='stylesheet' type='text/css'>
-    = javascript_include_tag 'application'
+    = javascript_include_tag 'application', skip_pipeline: true
```
sethherr: Why does this need to skip pipeline? Does the pipeline no longer do anything useful? e.g. fingerprinting?

No need to answer this before shipping, just something I'm curious about

app/workers/after_user_create_worker.rb
```diff
@@ -51,13 +51,16 @@ def associate_ownerships(user, email)
...
     memberships = Membership.unclaimed.where(invited_email: email)
     return false unless memberships.any?
-    memberships.pluck(:id).each_with_index do |m_id, index|
-      if index == 0 # We want to do the first one inline so we can redirect the user to the org page
-        ProcessMembershipWorker.new.perform(m_id, user.id)
-      else
-        ProcessMembershipWorker.perform_async(m_id, user.id)
-      end
+
+    first, *rest = memberships.pluck(:id)
```
sethherr: Was this added to Rails, or is it a new ruby feature? it's nifty regardless.

config/secrets.yml
```diff
@@ -0,0 +1,25 @@
+# Be sure to restart your server when you modify this file.
```
sethherr: Is this file being used?

spec/controllers/bikes_controller_spec.rb
```diff
@@ -96,7 +96,7 @@
       end
       context "passed all permitted params" do
         let(:query_params) do
-          {
+          ActionController::Parameters.new(
```
sethherr: Why are you switching this hash to permitted params? It seems weird that you're calling `permit` in the test.

spec/controllers/bikes_controller_spec.rb
```diff
@@ -1501,30 +1509,9 @@ def expect_bike_transferred_but_unclaimed(bike, user)
         expect(bike.bike_organization_ids).to match_array([organization.id, organization_2.id])
         expect(bike.editable_organizations.pluck(:id)).to eq([organization_2.id])
       end
-      context "empty organization_ids_can_edit_claimed" do
```
sethherr: I think this is an important test case? This tests that you can remove all the editable organizations

app/controllers/api/v3/organizations.rb
sethherr: I believe the root cause of this is that the services are tested using hashes rather than permitted parameters (which I think is reasonable, maybe even a good thing).

Something cludgy that might work is switching the `.to_hash` to `.as_json` - but, this shouldn't be resolved before shipping this PR. We can discuss.

app/controllers/api/v2/bikes_search.rb
jmromer: Line 47 is a no-op (likewise L67, L79, L185) since the hash literal isn't returned — I got the impression these might have been the response payloads for their respective endpoints originally but then more was added...?

sethherr: Ok! Ignore

spec/controllers/bikes_controller_spec.rb
jmromer: Agreed, but I'm accommodating the stub, which now expects a params object instead of a hash:

```rb
expect(Bike).to receive(:searchable_interpreted_params).with(query_params, ip: "special") { {} }
```

I think it's definitely worth refactoring though — it might shed some light on the path helper / to_hash issues.

sethherr: Ahhhh. Ok. sorry.

app/workers/after_user_create_worker.rb
jmromer: Just a refactor made while I was debugging — the first/rest array-deconstruction pattern is just FP-flavored Ruby.

I intended it to be temporary but left it in because I thought it expressed intent more clearly (process the first synchronously, the rest asynchronously). Happy to revert if it seems too terse though.

spec/controllers/bikes_controller_spec.rb
jmromer: IIRC this was an instance of [this issue](https://github.com/rails/rails/issues/27215#issuecomment-263972010) but I'll do a manual test to see if we still have the expected behavior.

> This is the expected behavior. There is no way to encode an empty array using an HTML form (e.g. Content-Type: url-form-encoded). The reason your tests passed in Rails 4.2 is because controller tests did not encode their parameters, they simply passed the hash through directly. In Rails 5 it encodes them. If your controller cares about empty arrays, it's likely that you are dealing with JSON requests. You can do that in your test with as: :json. If you're just dealing with form input, you will never receive an empty array.

app/workers/after_user_create_worker.rb
sethherr: No, I like it a lot!

spec/controllers/bikes_controller_spec.rb
sethherr: Well, we need to solve this test case.

First solution off the top of my head (provided the issue you remarked upon is correct): add a hidden field to that form which makes the controller know the organization ids are being passed - so if no ids are passed, it defaults to an empty array.

Let's create an issue for this and resolve after this is merged, I would prefer to ship this first.

sethherr: But - can you put this test back in, and mark it pending, before merging?

db/migrate/20191216054404_add_confidential_to_applications.rb
sethherr: Is it possible to run this migration prior to merging this? If we can not upgrade and migrate simultaneously, I think it would be nice.

app/views/layouts/doorkeeper.html.haml
jmromer: Good nudge, thanks! The `:skip_pipeline` option was a step in the wrong direction — there's a bug on master here — it should be `javascript_pack_tag` since webpacker is handling that `application.js`. The jQuery source is also missing:

When I visit http://localhost:3001/oauth/applications on master, I see the following in the JS console:

```
GET http://localhost:3001/javascripts/application.js net::ERR_ABORTED 404 (Not Found)
applications:5507 Uncaught ReferenceError: $ is not defined
    at applications:5507
```

The original line raises an `Sprockets::Rails::Helper::AssetNotFound` exception on this branch because of the Sprockets upgrade.

Fixed in 66bd3546

config/secrets.yml
jmromer: Good nudge thanks — in routes.rb and (with ad32c195) in secret_token.rb:

```
config/routes.rb
6:Sidekiq::Web.set :session_secret, Rails.application.secrets[:secret_key_base]

config/initializers/secret_token.rb
1:Bikeindex::Application.config.secret_key_base = Rails.application.secrets[:secret_key_base]
```

spec/controllers/bikes_controller_spec.rb
jmromer: Agreed on validating the use case is still satisfied — no change there, it still works, afaict:

demo

We have a `:organization_ids_can_edit_claimed_present` hidden field on the form presently:

```haml
-# app/views/bikes/edit_groups.html.haml L13-15 (8ae68fdc)

- if @bike.bike_organizations.any?
  = hidden_field_tag :organization_ids_can_edit_claimed_present, true
  .related-fields.no-divider-row
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/8ae68fdc/app/views/bikes/edit_groups.html.haml#L13-L15)]</sup>

The removed test case can't pass in Rails 5, and on `master` it's a spuriously passing test. I think this test is validating the use case in question?:

```rb
# spec/controllers/bikes_controller_spec.rb L1512-1526 (992e9c53)

context "organization_ids_can_edit_claimed_present" do
  it "updates the bike with the allowed_attributes, marks no organizations can edit claimed" do
    put :update, params: { id: bike.id, bike: allowed_attributes, organization_ids_can_edit_claimed_present: "1" }
    expect(response).to redirect_to edit_bike_url(bike)
    expect(assigns(:bike)).to be_decorated
    bike.reload
    expect(bike.hidden).to be_falsey
    allowed_attributes.except(*skipped_attrs).each do |key, value|
      pp value, key unless bike.send(key) == value
      expect(bike.send(key)).to eq value
    end
    expect(bike.bike_organization_ids).to match_array([organization.id, organization_2.id])
    expect(bike.editable_organizations.pluck(:id)).to eq([])
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/992e9c53/spec/controllers/bikes_controller_spec.rb#L1512-L1526)]</sup>

jmromer: I'll add it back in just in case though 👍

sethherr: Ha. Historical Seth was as smart as current Seth, too bad I don't remember what he did.

Thinking about it - I bet this test was a test I left in because I'm lazy - I probably initially tested it this way, and then figured out that the test didn't represent how things worked, added the new test but didn't delete the old, unrepresentative test. 

Basically, I'm a forgetful idiot. Feel free to delete. Closing the issue.

PR #1428, opened 2019-11-22
Upgrade Sprockets to v4

jmromer:
Upgrades Sprockets to 4.0 

### Migration notes

- Asset precompilation moved from assets.rb initializer to `app/assets/config/manifest.js`: 2a090ac5
- Forks swagger-ui_rails gem 0.1.7 to update its sprockets shims: 17351891

### Screenshots

Screen Shot 2019-11-26 at 10 58 08 PM
Screen Shot 2019-11-26 at 10 58 14 PM
Screen Shot 2019-11-26 at 10 58 19 PM
Screen Shot 2019-11-26 at 10 59 05 PM


app/assets/javascripts/application_revised.js
```diff
@@ -11,7 +11,7 @@
 //= require mustache
 //= require Stickyfill
 //= require select2/select2.full.js
-//= require selectize/standalone/selectize.js
+//= require external_scripts/selectize.js
 //= require external_scripts/selectize_placeholder_plugin.js
```
jmromer: Selectize is broken 

Screen Shot 2019-11-22 at 5 46 42 PM
Screen Shot 2019-11-22 at 5 46 52 PM

app/assets/javascripts/application_revised.js
```diff
@@ -39,5 +39,5 @@
 //= require external_scripts/jquery.fileupload.js
 
 // Our actual scripts:
-//= require init.coffee
+//= require init
```
jmromer: Double-check this still works

app/assets/stylesheets/documentation_v2.scss
```diff
@@ -1,4 +1,4 @@
-// = require swagger-ui
+//= require swagger-ui
```
jmromer: Swagger UI is broken. This patch adds an empty file as an expedient since the gem hasn't been updated since 2015. We may need to fork it.

Screen Shot 2019-11-22 at 5 47 31 PM
Screen Shot 2019-11-22 at 5 47 38 PM

app/javascript/packs/bikes_search/components/BikeSearch.js
```diff
@@ -4,6 +4,7 @@ import React, { Fragment, Component } from "react";
 
 import Loading from "../../Loading";
 import honeybadger from "../../utils/honeybadger";
+import TimeParser from "../../utils/time_parser";
```
jmromer: Pre-existing bug that now raises an exception since there's a stricter separation between webpack and sprockets-managed JS.

app/assets/config/manifest.js
```diff
@@ -0,0 +1,23 @@
+//= link_tree ../images
+
+//= link_tree ../javascripts/external_scripts/
+
+//= link embed.js
+//= link embed_user.js
+//= link documentation_v2_and_v3.js
+//= link application_revised.js
+//= link i18n.js
+//= link registrations.js
```
jmromer: Screen Shot 2019-11-23 at 12 06 45 AM

app/assets/javascripts/application_revised.js
sethherr: Does it make sense to switch to webpack only scss at this point?

We still will need sprockets for the email scss...

To make the switch, we'd need to bump the version of bootstrap that the main app uses. This would take a bunch of work, but might be worth it? (see #1005)

app/assets/stylesheets/documentation_v2.scss
sethherr: I am happy locking the documentation v2/v3 styles and js at this point - ie creating compiled versions of the css and js, and importing those, rather than gems, etc.

If we redesign the documentation, I'm happy figuring out a newer solution at that point, rather than extending what we have now.

(note: if we do make a compiled version, we should put a comment at the top with the URL of the PR where the compilation happens. This would have saved me headaches if I did it for `og_application.css`)

app/assets/javascripts/application_revised.js
jmromer: My hunch is that it makes sense to err on the side of incrementalism and finish migrating to Webpack once Rails is upgraded, but I'm not totally set on that yet.

From what I'm seeing so far the Rails update may compel tackling the [Bootstrap upgrade](https://github.com/bikeindex/bike_index/pull/1429) anyway — there are breaking changes in the way assets are sourced that lead to a lot of undefined mixins / variables, which breaks asset compilation. 

My plan has been to update Sprockets, then Bootstrap, then Rails, in that order. I'm updating the second step to "handle" Bootstap, where "handle" is one of the following approaches:

1) Minimal: Figure out how to accommodate the changes in Sprockets so assets compile as they do now, without changes to Bootstrap (which may or may not be possible w/o forking)
2) Expedient: Temporarily vendorize Bootstrap stylesheets to fix asset compilation and return to the Bootstrap upgrade later
3) Thorough: Update Bootstrap, which might require backing out some customizations and [incrementally migrating](https://getbootstrap.com/docs/4.0/migration/) (potentially time-consuming)
4) Aggressive: Lean into the already partially completed migration from Sprockets to Webpack (risky, but also potentially time-saving? I think so at least, but I'd need to explore this a bit more to know exactly how.)

There's a lot of wilderness in our assets directory — are the Bootstrap files there in fact customized? Or just vendorized for some other reason?

app/assets/stylesheets/documentation_v2.scss
jmromer: Sweet — that'll save some time.

Not sure if by _compile_ you mean vendorize or vendorize + minify, but in case it's the latter — I had to un-minify `og_application.css` to track down some bugs in it — I think it's better if we let the asset pipeline handle the minification (or set it up to do so if it doesn't already).

sethherr: Agreed @jmromer, just vendorize - they shouldn't be minified, that was a bad choice on my part. I _would_ prefer it to all be in one file, or one directory, and have it clearly documented in the root js/css files what is happening, so that we know why it is the way it is, and don't accidentally use the vendorized code.

app/assets/javascripts/application_revised.js
sethherr: As per my other comment - I think option 2 is a good choice here, so long as there are comments explaining what is happening and why.

app/assets/config/manifest.js
jmromer: Resolved

Gemfile
```diff
@@ -82,7 +86,7 @@ gem "doorkeeper", "~> 3.1.0"
 gem "grape", "~> 0.19.1"
 gem "grape-active_model_serializers", "~> 1.4.0"
 gem "grape-swagger", "0.11"
-gem "swagger-ui_rails"
+gem "swagger-ui_rails", github: "jmromer/swagger-ui_rails", branch: "bike_index_0.1.7"
```
jmromer: see: https://github.com/jmromer/swagger-ui_rails/pull/1

jmromer: we should fork to bikeindex

.circleci/config.yml
```diff
@@ -53,7 +53,14 @@ jobs:
             tar -xzf git.tgz
             cd git-${GIT_VERSION}
 
+            export DEFAULT_HELP_FORMAT="man"
+            autoconf
+            ./configure
+
             sudo make prefix=/usr/local all
+            sudo make install
+            sudo make man && sudo make install-man
```
jmromer: This builds git with the ability to work with submodules (currently missing)

app/assets/stylesheets/registrations.scss
```diff
@@ -1,7 +1,7 @@
 /*
  * Revised manifest file
  * things from rails-assets
- *= require selectize/selectize.bootstrap3.scss
+ *= require selectize/selectize.bootstrap3
  *= require dataTables/bootstrap/3/jquery.dataTables.bootstrap
```
sethherr: FWIW, I'm 90% sure we don't need dataTables anymore. Not sure if there is a better time to remove it than now 🤷‍♂️

config/application.rb
```diff
@@ -51,11 +51,14 @@ class Application < Rails::Application
     end
 
     config.generators do |g|
-      g.factory_bot true
-      g.helper false
-      g.javascripts false
-      g.stylesheets false
+      g.factory_bot nil
```
sethherr: Can we get the factory_bot to auto generate? I think that is the existing behavior, which I think is nice.

PR #1437, opened 2019-11-27
Build CI Git with submodules

jmromer:
Adds submodule support to the Git we use on CI. (Submodule support is needed when a gem fetch from GitHub contains submodules we also want to fetch.)

Extracted from #1428


jmromer:
> if this is only required because of assets that could be vendored instead, I feel like it would be easier to vendor the assets than add this complexity.

This fixes an earlier oversight — git being built without the submodule command will create headaches down the line.

PR #1441, opened 2019-11-27
Upgrade Sprockets to v4

jmromer:
Upgrades Sprockets to 4.0 

Migration notes

- Asset precompilation moved from assets.rb initializer to `app/assets/config/manifest.js`: 2a090ac5
- Forks swagger-ui_rails gem 0.1.7 to update its sprockets shims: 17351891


sethherr:
Note: this is a redo of #1428, which was reverted in #1440

jmromer:
I'm probably going to close this in favor of shipping the sprockets upgrade together with the Rails 5 upgrade. I haven't been able to track down the cause of the asset compilation failure blocking deployment for this branch but I *have* been able to successfully deploy Rails 5 branch (which includes the sprockets upgrade) to staging, so there's perhaps a configuration dependency between Sprockets 4 and Rails 5 that I'm overlooking. I'm investigating a bit further but I don't want to go any further down the rabbit hole than necessary.

PR #1476, opened 2020-01-24
WIP: Improve inconsistent slugifying

sethherr:
This is low priority, was touching something else and saw it, figured I'd touch it and add some more tests - but I'm not finishing this right now.

---

Make slugifying more consistent.

When/if this is merged, we'll need to update all objects stored in the database with slugs.

I would _love_ if we stopped having 3 different methods of sluggifying things.


sethherr:
Doesn't bring enough value for the hassle.

PR #1490, opened 2020-01-31
Organization registration of abandoned bikes without owners

sethherr:


sethherr:
Have to merge prematurely

PR #1491, opened 2020-01-31
Migrate to Rails.application.credentials

jmromer:
Rails 5.2 introduced `credentials.yml.enc` as the default way to manage app secrets, replacing `secrets.yml`:

https://medium.com/cedarcode/rails-5-2-credentials-9b3324851336

This patch replaces `secrets.yml` with `credentials.yml.enc` and requires the master key be set in prod. 
NB: This migration isn't strictly required since Rails 5 is backward-compatible with secrets.yml, so we can close this if that'd be preferable.

The master key env var is set on C66.


sethherr:
Two things:

1. I think the encrypted secrets file makes a lot of sense, but I've been curious about the best practices of `credentials.yml.enc` in public repositories and haven't found solid answers. 
  - Is it a good practice to commit the encrypted file to github publicly? If not, what's the expectation?
  - Are the secrets in it secure if it's public?
  - Is it possible to decrypt the master key based on having the encrypted output, being able to view change history for it and having a large amount of time?
2. I would like the Bike Index repository to be able to be downloaded and run on someone's computer with a minimal amount of fuss. If we're adding this, we should add stuff to bin/setup to generate a default secrets file.

jmromer:
> I would like the Bike Index repository to be able to be downloaded and run on someone's computer with a minimal amount of fuss.

Agreed. This seems dispositive because afaict we would have to share a master key file (infeasible) or provide an env variable in `.env` (which would obviate some of the benefit of migrating to the new secrets system). I'll close this in favor of https://github.com/bikeindex/bike_index/pull/1495

PR #1497, opened 2020-02-05
Replace unicorn with puma

jmromer:
- Replaces the application server with puma, which in Rails 5 replaces unicorn in prod and webrick in development.

- Removes unicorn-worker-killer — do we need a replacement? (ie, was it added to remedy a memory consumption issue? 🤔)

(NB: Not strictly required — we can punt on this if it introduces elective complexity)


Procfile
```diff
@@ -1,3 +1,3 @@
 # Running puma with 4,16 threads and 3 workers
-custom_web: bundle exec unicorn_rails -c config/unicorn.rb -E $RAILS_ENV -D
+custom_web: bundle exec puma -e $RACK_ENV -b unix:///tmp/web_server.sock --pidfile /tmp/web_server.pid -d
```
jmromer: ref: https://help.cloud66.com/rails/how-to-guides/rack-servers/puma-rack-server.html

jmromer:
Sounds good. There is a puma version of the same (https://github.com/schneems/puma_worker_killer) but the docs actually recommend against using it unless absolutely necessary.
I'll keep an eye on Honeybadger — if anything a memory leak would be interesting to debug :]

PR #1512, opened 2020-03-02
Simplify Geocodeable, normalize on geocoding by address field

jmromer:
We currently have the following `Geocodeable` models, each with the listed location-value columns (in addition to `latitude` and `longitude`):

<details>
<summary>MailSnippet</summary>

- address

</details>

<details>
<summary>StolenRecord</summary> 

- city
- country_id
- state_id
- street
- zipcode

</details>

<details>
<summary>Bike</summary>

- city
- country_id
- zipcode

</details>

<details>
<summary>TwitterAccount</summary>

- address
- city
- country
- state
</details>

<details>
<summary>Location</summary>

- city
- country_id
- state_id
- street
- zipcode

</details>

<details>
<summary>User</summary>

- city
- country_id
- state_id
- street
- zipcode

</details>

This PR is to simplify the `Geocodeable` mixin by consistently geocoding using an `address` column, adding one to the above models as needed. 

For models where the `address` value is derived from other fields, we set the value in a `before_validation` `set_address` hook.

This patch renames the existing `StolenRecord#address` method to `#display_address` in order to retain its current functionality.


spec/models/location_spec.rb
```diff
@@ -9,10 +9,6 @@
   end
 
   describe "address" do
-    it "strips the non-digit numbers from the phone input" do
```
jmromer: Removing this example since its description appears to be copy-pasted from the preceding one, and the assertion is reproduced the following example.

app/models/user.rb
```diff
@@ -320,20 +331,8 @@ def userlink
     end
   end
 
-  def address(skip_default_country: false)
-    country_string = country&.iso
-    country_string = nil if skip_default_country && country_string == "US"
```
jmromer: `User#address` isn't currently invoked anywhere except for in `address_hash`

app/models/concerns/geocodeable.rb
```diff
@@ -4,34 +4,26 @@ module Geocodeable
...
+    # By default:
+    #  - we skip geocoding if the `skip_geocoding` flag is set.
+    #  - geocode if address is blank or has changed
+    #
+    # Overwrite this method in inheriting models to customize skip-geocoding
+    # logic.
     def should_be_geocoded?
-      return false if skip_geocoding?
-      geocode_data.present? && any_geocode_columns_changed?
+      !skip_geocoding? && (address.blank? || address_changed?)
```
sethherr: Shouldn't this be `address.present?` ?

spec/models/bike_spec.rb
```diff
@@ -1335,7 +1335,7 @@
       end
       let(:bike) { FactoryBot.create(:stolen_bike) }
       it "does not get out of integer errors" do
-        expect(bike.listing_order).to be < 10000
+        expect(bike.listing_order).to be < 1e10
```
sethherr: While this is the limit for out of range, I feel like testing that it's well below the limit is good.

If you'd prefer `1e4` here, that's fine.

jmromer: No prob  — reverted in dea790b1
The test currently fails intermittently but it's pretty rare for it to exceed `10_000`.

app/models/concerns/geocodeable.rb
jmromer: Good catch thanks — I think this was an incomplete DeMorgan's Law application but I just ended up confusing myself. Fixed and added a test via `StolenRecord` in d30d8b7b5554d83

spec/models/bike_spec.rb
sethherr: If it is actually flaky, maybe switch it to 1e5? I would much prefer non-flaky, I just also think having some extra space up there would be great.

app/models/bike.rb
```diff
@@ -809,30 +824,12 @@ def propulsion_type_name
...
-  end
-
   # Take lat/long from associated geocoded model
-  # Only geocode if no lat/long present and geocode data present
+  # Only geocode if lat/long are blank and address present
   def should_be_geocoded?
     return false if skip_geocoding?
     return false if latitude.present? && longitude.present?
-    geocode_data.present?
+    address.present?
```
sethherr: I'm a little confused by this, it seems circular - but maybe I'm missing something?

- Before validation check this method (`should_be_geocoded?`)
- This line checks if address is present
- If address *is* present, we set the address.

app/models/concerns/geocodeable.rb
sethherr: Thanks for the test! Definitely makes it clear

sethherr:
After this is merged, I would like to do two things:

1. Makes the location attributes on all the models the same (probably `city`,`country_id`,`state_id`,`street`,`zipcode` and `address`) - including Organization
2. Refactor `set_location_info` on Bike and remove duplicated code - since we will be assigning from the same attribute names, it can be a much smaller method

app/models/bike.rb
jmromer: Good call-out—it's actually L830 that's operative in that situation (in tests) but using `:should_be_geocoded?` muddies that a bit. Fixed in 8d6205df

PR #1527, opened 2020-03-26
POS manual setting

sethherr:


sethherr:
Memoizing doesn't seem to speed things up and I worry it could be surprising and frustrating to someone sometime, so ignoring.

PR #1534, opened 2020-04-01
Add location appointments

sethherr:


sethherr:
As per https://github.com/bikeindex/bike_index/pull/1535#issuecomment-611062349

> This doesn't seem like something that enough users want to justify working on it 😢
>
> I'll continue to ask around and see if there is something from this that would be useful, but closing for now.

I'm pulling out the management stuff that was generally useful, and then closing this.

PR #1535, opened 2020-04-02
Day schedules

sethherr:
Part of #1534, separated out to make easier to deal with


sethherr:
This doesn't seem like something that enough users want to justify working on it 😢

I'll continue to ask around and see if there is something from this that would be useful, but closing for now.

---

I'm a fan of the business logic of this for scheduling though.

- `LocationSchedule` is either a reference (a day of the week), or a specific date
- Specific dates inherit the schedule from the reference, to make it easier to set the specific differences on that day
- There would be a calendar view that would show the days that are specifically modified, and give the user the ability to click on a given day and add a specific schedule for it

Issue #1509, opened 2020-02-12
Create Parking Notifications


sethherr:
This issue is intended to link together all the work on parking notifications, since it is sprawling.

First, created abandoned records (#1490, #1493) - but after conversations with users, realized this didn't solve the issue (and that organization messages don't solve the issue either).

So, decided to create parking notifications

- Renamed abandoned records to Parking notifications with #1503
- Parking notifications creation frontend #1504
- Parking Notifications time and bike search #1508

PR #1536, opened 2020-04-03
Immediate merges from Location Appointments

sethherr:
Part of #1509


sethherr:
Things that I want included now from #1534

PR #1546, opened 2020-04-11
Consistify location-related fields, simplify Bike#set_location_info

jmromer:
Ensures all `Geocodeable` models have the following fields (in addition to lat/lon and address):

  - street
  - city
  - state_id
  - zipcode
  - country_id

<details>
<summary>Schema migration</summary>

```rb
# db/migrate/20200410183949_add_missing_location_fields_to_geocodeable_models.rb L8-29 (4248b3ca)

def change
  change_table :mail_snippets do |t|
    t.column :street, :string
    t.column :city, :string
    t.column :zipcode, :string
    t.belongs_to :state, index: true
    t.belongs_to :country, index: true
  end

  change_table :bikes do |t|
    t.column :street, :string
    t.belongs_to :state, index: true
  end

  change_table :twitter_accounts do |t|
    t.column :street, :string
    t.column :zipcode, :string
    # data-migrate current state and country to foreign keys
    t.belongs_to :state, index: true
    t.belongs_to :country, index: true
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/4248b3ca/db/migrate/20200410183949_add_missing_location_fields_to_geocodeable_models.rb#L8-L29)]</sup>
</details>

<details>
<summary>Data migration</summary>

```rake
# lib/tasks/data_migrations.rake L5-18 (947ebb49)

desc "Migrate state and country fields from names to foreign keys."
task up: :environment do
  puts "Migrating..."

  TwitterAccount.transaction do
    TwitterAccount.find_each do |ta|
      ta.state_id = State.fuzzy_find(ta[:state])&.id
      ta.country_id = Country.fuzzy_find(ta[:country])&.id
      ta.save
    end
  end

  puts "done."
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/947ebb49/lib/tasks/data_migrations.rake#L5-L18)]</sup>
</details>

Refactors `Bike#set_location_info` to leverage the consistent fields:

```rb
# app/models/bike.rb L656-695 (1b10da8e)

# Select the source from which to derive location data, in the following order
# of precedence:
#
# 1. The current parking notification, if one is present
# 2. The current abandoned record, if one is present
# 3. The current stolen record, if one is present
# 4. The creation organization, if one is present
# 5. The bike owner's address, if available
# 6. The request's IP address, if given
def find_location_info(geolocation = nil)
  location_records = [
    current_parking_notification,
    current_stolen_record,
    creation_organization,
    owner,
    geolocation,
  ]

  location_records
    .compact
    .map { |rec| Geohelper.location_from_result(rec) }
    .find { |rec| rec[:country] && (rec[:city] || rec[:zipcode]) }
end

# Set the bike's location data (lat/long, city, postal code, country, etc.).
def set_location_info(request_location: nil)
  # ensure current_stolen_record is set if available
  find_current_stolen_record

  # select a source of location info
  location_record = find_location_info(request_location)
  return unless location_record.present?

  self.address = location_record[:address]
  self.city = location_record[:city]
  self.country = location_record[:country]
  self.latitude = location_record[:latitude]
  self.longitude = location_record[:longitude]
  self.zipcode = location_record[:zipcode]
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/1b10da8e/app/models/bike.rb#L656-L695)]</sup>

### TODO

- [ ] Follow-on migration dropping `twitter_accounts.state` and `twitter_accounts.country` (see #1547)


app/models/bike.rb
```diff
@@ -652,67 +653,46 @@ def registration_location
...
-      self.zipcode = current_stolen_record.zipcode
-      self.address = current_stolen_record.display_address(force_show_address: true)
-    elsif location_info_present?(creation_organization)
-      self.latitude = creation_organization.location_latitude
-      self.longitude = creation_organization.location_longitude
-      self.city = creation_organization.city
-      self.country = creation_organization.country
-      self.zipcode = creation_organization.zipcode
-      # This is a hack, need to fix
-      self.address = [city, zipcode, country&.name].select(&:present?).join(" ").presence
```
jmromer, line 699: L699 is now equivalent to `self.address = creation_organization.address`

app/models/bike.rb
```diff
@@ -652,67 +653,46 @@ def registration_location
...
-      self.zipcode = owner.zipcode
-      self.address = owner.address
-    elsif location_info_present?(request_location)
-      self.latitude = request_location.latitude
-      self.longitude = request_location.longitude
-      self.city = request_location.city
-      self.country = Country.fuzzy_find(request_location&.country_code)
-      self.zipcode = request_location.postal_code
-      # Need to test what the responses from this look like - probably slightly different
-      # self.address = current_parking_notification.address
```
jmromer, line 714: L714 (uncommented) is here in case we need it but I can also kill it: https://github.com/bikeindex/bike_index/pull/1546/files#diff-998a6bdfbdc9035d6f2c1a089867c8efR677

app/models/concerns/geocodeable.rb
```diff
@@ -3,6 +3,13 @@
 module Geocodeable
   extend ActiveSupport::Concern
 
+  # Does the given object have enough location info to set location info on a
```
sethherr, line 6: Why are there objects that don't give enough information, but still have a location?

Couldn't this just check if the lat & long are present? And  then be generalized, e.g. `has_location?` - which would be the same method that is checked to determine whether to geocode in the first place?

app/models/request_geolocation_info.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+# Decorates a Geocoder:Result object to consistify its interface
+# with that of `Geocodeable` ActiveRecord objects.
+# see: app/models/concerns/geocodeable.rb
+class RequestGeolocationInfo < SimpleDelegator
```
sethherr, line 6: I think there is already a class that is suppose to do this - `Geohelper` - which also is suppose to manage the differences between test responses, production responses, different Google responses, and Maxmind responses (we're also going to be adding/switching to mapbox soon, so it will need to handle all that as well)

spec/factories/twitter_accounts.rb
```diff
@@ -7,15 +7,14 @@
...
-    state { "NY" }
-
     address { "278 Broadway, New York, NY 10007, USA" }
     latitude { 40.7143528 }
     longitude { -74.0059731 }
 
     after(:build) do |twitter_account, _evaluator|
-      usa = Country.find_by(iso: "US") || FactoryBot.build(:country_us)
-      twitter_account.country = usa
+      new_york = State.find_by(abbreviation: "NY") || FactoryBot.build(:state_new_york)
```
sethherr, line 15: Something cool that could be done with all these location traits, now that everything has the same attributes - we could only define the traits once, in the `spec/factories.rb` file - https://stackoverflow.com/questions/23204677/share-traits-across-different-factories

And then we wouldn't need to include Los Angeles, nyc, chicago, amsterdam - anywhere else.

app/models/bike.rb
sethherr: Having worked with this now for a little while, I'm not sure there is a strong benefit to having an address column. Can't we just generate it when it needs to be displayed?

I don't think it has a drawback to being there, other than the fact that we have `street`, `address` and `display_address` methods/attributes - which is confusing.

app/models/request_geolocation_info.rb
jmromer: This intent for this class is primarily to map `country` messages to `country_code` and `zipcode` to `postal_code`, to match the other receivers, which are ActiveRecord objects. I don't see a way for `Geohelper` to do that.

app/models/concerns/geocodeable.rb
jmromer: > Why are there objects that don't give enough information, but still have a location?

Dirty data and/or different requirements for different objects, I imagine. 

For `Bike` records specifically, we expect a country and city or country and zipcode:

```rb
    object&.country.present? &&
      (object&.city.present? || object&.zipcode.present?)
```

A given accompanying record we're pulling location info from (stolen record, organization, etc) to set a bike's location may not satisfy that condition.

We're using more than just the Bike latitude and longitude (city, country, etc) elsewhere.

app/models/bike.rb
jmromer: > Can't we just generate it when it needs to be displayed?

That's the previous behavior, but for the purposes of simplifying the `Geocodeable` implementation (in particular the `geocode_columns` method below) we pre-generate and store an address. 
```rb
# app/models/concerns/geocodeable.rb L16-36 (2b78f947)

  # Customize by overwriting in the Geocodeable model.
  def geocode_data
    @geocode_data ||= address
  end

  # An array of symbols of the db columns upon which `geocode_data`
  # depends.These are checked for changes before geocoding using
  # `should_be_geocoded?`. By default, empty, so we don't check for changes.
  def geocode_columns; []; end

  def any_geocode_columns_changed?
    return true if geocode_columns.blank?
    geocode_columns.any? { |col| public_send("#{col}_changed?") }
  end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/2b78f947/app/models/concerns/geocodeable.rb#L16-L36)]</sup>

But if the set of fields for `Geocodeable` are common by convention, we can alternatively drop `address` and specify `geocode_columns` in `Geocodeable`.

app/models/request_geolocation_info.rb
sethherr: Is there a way to put it into `Geocodeable`?

jmromer: No prob — updated in 1b10da8e : 

```rb
# app/services/geohelper.rb L61-83 (1b10da8e)

# Extract location info from the given result object `result`,
# expected to be an ActiveRecord or Geocoder::Result object.
#
# Returns a Hash
def location_from_result(result)
  location = {
    address: result.address.presence,
    city: result.city.presence,
    latitude: result.latitude.presence,
    longitude: result.longitude.presence,
  }

  if result.respond_to?(:country_code)
    location.merge(
      country: Country.fuzzy_find(result.country_code),
      zipcode: result.postal_code.presence,
    )
  else
    location.merge(
      country: result.country.presence,
      zipcode: result.zipcode.presence,
    )
  end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/1b10da8e/app/services/geohelper.rb#L61-L83)]</sup>

spec/factories/twitter_accounts.rb
jmromer: Nice! good catch, thanks — updated in 3c31ff6f

app/models/bike.rb
sethherr: > But if the set of fields for `Geocodeable` are common by convention, we can alternatively drop `address` and specify `geocode_columns` in `Geocodeable`.

I _think_ this would be good. Do you think it would make things simpler or more complicated? If it's ultimately going to make things more confusing, it's not worth doing. The only reason to do it would be to make things easier to understand, I don't think there is anything actually wrong with the current implementation.

Might make sense to work on this in a different PR, this PR seems 🎉, but  ¯\\\_(ツ)\_/¯

app/models/stolen_record.rb
```diff
@@ -89,24 +88,15 @@ def recovering_user_owner?; recovering_user.present? && bike&.owner == recoverin
...
-      (force_show_address || show_address) ? street : nil,
-      city,
-      state&.abbreviation,
-      zipcode,
-      country_string,
-    ].reject(&:blank?).join(", ")
+    address(
+      street: (force_show_address || show_address),
+      country: [(:skip_default if skip_default_country)]
+    ).presence
```
jmromer, line 99: WIP

app/models/parking_notification.rb
```diff
@@ -116,26 +116,12 @@ def repeat_number
...
-  end
-
-  # Proxy method for now
-  def display_address(skip_default_country: true, force_show_address: false)
-    address(skip_default_country: skip_default_country, force_show_address: force_show_address)
+    Geocodeable.address(
+      self,
+      street: (force_show_address || show_address),
+      country: [(:skip_default if skip_default_country), :optional],
+    ).presence
```
jmromer, line 124: WIP

app/models/bike.rb
jmromer: That makes sense — that adds more logic to `Geocodeable` but reduces duplication overall. 27e79db6 drops the previously added `address` columns and extracts address-string-building logic to a single `Geocodeable` class method. 

<details>
<summary><code>Geocodeable.address</code></summary>

```rb
# app/models/concerns/geocodeable.rb L41-84 (d3d9d0af)

# Build an address string from the given object's location data.
#
# The following keyword args accept booleans for inclusion / omission in the
# string: `street` `city` `state` `zipcode`, `country`.
#
# By default, the country is included as an ISO abbreviation and is required
# (an empty string is returned if no country is available.)
#
# The `country` keyword arg also accepts a list of options to customize
# output:
#
# - :iso or :name for the format
# - :optional to make the country optional
# - :skip_default to omit the country name if it's the default country (US)
#
# Returns a String.
def self.address(obj, street: true, city: true, state: true, zipcode: true, country: [:iso])
  return "" if obj.blank?

  include_country =
    country && !(obj.country&.default? && country.include?(:skip_default))

  country_name =
    if include_country
      country_format = country.find { |e| e.in? %i[iso name] } || :iso
      country_name = obj.country&.public_send(country_format)

      country_is_required = !country.include?(:optional)
      not_enough_info = country_is_required && country_name.blank?
      return "" if not_enough_info

      country_name
    end

  [
    street && obj.street,
    city && obj.city,
    [
      state && obj.state&.abbreviation,
      zipcode && obj.zipcode,
    ].reject(&:blank?).join(" "),
    country_name,
  ].reject(&:blank?).join(", ")
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/d3d9d0af/app/models/concerns/geocodeable.rb#L41-L84)]</sup>
</details>

PR #1562, opened 2020-04-21
Docker setup for development

wwwbruno:
Hi, I noticed here (https://github.com/bikeindex/bike_index/pull/1423) that you were opening for a Docker setup for a development environment, so I wanted to help with that!

## Concerns
- Does not work inside Docker or VM: 
  - `letter_opener` https://github.com/ryanb/letter_opener#remote-alternatives
  - `terminal-notifier` https://github.com/julienXX/terminal-notifier/issues/234


docker-compose.yml
```diff
@@ -0,0 +1,38 @@
...
+    command: bundle exec foreman start -f Procfile_development
+    depends_on:
+      - postgres
+      - redis
+    environment:
+      REDIS_URL: redis://redis:6379
+      PGHOST: postgres
+      PGUSER: postgres
+    env_file:
+    - docker.env
```
wwwbruno, line 15: I did not want to duplicate the .env file, but I found out that the format: `VARIABLE: value` is not valid for docker, needs to be `VARIABLE=value` see: https://docs.docker.com/compose/env-file/

And maintaining two files almost identical is pretty hard. I'll look further for another approach, or if anyone has a suggestion.

sethherr: Hey Bruno! Switching the dotenv file to use equals (which works fine for the app).

I'll merge #1564 as soon as it passes tests

Thanks for taking a look at this 😀

sethherr: merged @wwwbruno 🙂

wwwbruno: Thank you for the fast response @sethherr

Procfile_docker
```diff
@@ -0,0 +1,4 @@
+web: bundle exec puma -C config/puma.rb -e development -p 3001
+webpacker: ./bin/webpack-dev-server # Live update webpack js
+log: tail -f log/development.log
+hard_worker: bundle exec sidekiq
```
wwwbruno, line 4: I'm getting an error when trying to use `rerun` gem inside a Docker container in Mac.

```
hard_worker.1 | /usr/local/bundle/gems/terminal-notifier-2.0.0/vendor/terminal-notifier/terminal-notifier.app/Contents/MacOS/terminal-notifier: 1: /usr/local/bundle/gems/terminal-notifier-2.0.0/vendor/terminal-notifier/terminal-notifier.app/Contents/MacOS/terminal-notifier: Syntax error: "(" unexpected
```

I could not understand the issue but for now, I just created another `Procfile` to be able to keep working.

I'll revisit this issue later and put more effort into trying to understand and solve it.

wwwbruno: The issue was `rerun` calling `terminal-notifier`.

It's only possible to use `terminal-notifier` on the host machine, not inside Docker. https://github.com/julienXX/terminal-notifier/issues/234

So passing `--no-notify` option solves the issue. 406ba5b3

wwwbruno:
@sethherr In general my Pull Request is ready to review, but my specs are failing. I removed the Docker related files https://github.com/bikeindex/bike_index/pull/1562/commits/4da009af382212d708aef2704ae93d726c3e1ac7, but still fails.

I managed to make it pass locally, updating the `STOP_HELING_BASE_URL` variable on the `.env` file to a valid URL. But it's working with the current way on `master`.

```diff
diff --git a/.env b/.env
index 882893f3..64eb173d 100644
--- a/.env
+++ b/.env
@@ -34,7 +34,7 @@ TWITTER_ACCESS_SECRET=something
 ALLOWED_WRITE_ORGANIZATIONS=foo,bar
 TRANSLATION_IO_API_KEY=something
 LIVE_EXTERNAL_API_SPECS=false
-STOP_HELING_BASE_URL=something
+STOP_HELING_BASE_URL=http://something.org
 VERLOREN_OF_GEVONDEN_BASE_URL=something
 STOLEN_ALERT_TWITTER_1_USERNAME=test
 STOLEN_ALERT_TWITTER_1_ACCESS_TOKEN=test
```

Do you have any idea what could be?

sethherr:
I believe it is a CircleCI permissions issue.

I'm working on a branch that will skip the tests that depend on CircleCI variables unless the variables are present, which should solve this issue for you. I'll comment here again when it merges.

Thanks again for working on this!

Procfile_docker
```diff
@@ -0,0 +1,4 @@
+web: bundle exec puma -C config/puma.rb -e development -p 3001
+webpacker: ./bin/webpack-dev-server # Live update webpack js
+log: tail -f log/development.log
+hard_worker: bundle exec rerun --background --dir app,db,lib --pattern '{**/*.rb}' --no-notify -- bundle exec sidekiq
```
sethherr, line 4: Is there a reason that you are using a `Procfile_docker` rather than the development version, since they're the same?

config/database.yml
```diff
@@ -1,5 +1,7 @@
 development:
   adapter: postgresql
+  host: <%= ENV['PGHOST'] || 'localhost' %>
```
sethherr, line 3: Because of the way my postgres install is set up, having these fallbacks breaks it (homebrew uses your username for your postgres user).

Removing the fallback resolved the issue for me (done in #1575) - hopefully that will work for you here too.

Though if you set the variable `POSTGRESQL_USERNAME` rather than `PGUSER` it might work without needing to add these keys?

sethherr:
#1575 has been merged! If you merge in master, tests should pass on this branch @wwwbruno!

Procfile_docker
wwwbruno: They are almost the same, on the Docker version I added the `--no-notify` option.

I wrote here more details on what I found out.
https://github.com/bikeindex/bike_index/pull/1562#pullrequestreview-398024453

config/database.yml
wwwbruno: Good point, actually with `PGUSER` already works without the keys. It was working from the test database already! Probably I needed when doing initial configuration for debugging, but don't need anymore!

Procfile_docker
sethherr: Ahh, ok, thanks for the explanation!

wwwbruno:
@sethherr from my side I believe it's everything good. Do you have any other concerns or suggestions?

sethherr:
Nope! I think it's good, merging now.

Thanks for bumping @wwwbruno 🙂

PR #1587, opened 2020-05-08
Redirect after sign in to bike for writing message

### LINKED ISSUE #1445: Redirect after sign in to bike for writing message
When clicking the "write a user a message" on a stolen bike, you are redirected to sign in if not signed in. After signing in, you're redirected to your home rather than back to the bike

wwwbruno:
Fixes #1445

- On https://github.com/bikeindex/bike_index/blob/master/app/controllers/concerns/controller_helpers.rb#L117 it expects a string starting with `BASE_URL` or `/`.  So, just adding a `/` on the `data-redirect` attribute, already redirects the user without any other change.
- In order to redirect the user even when the password fails, I merged the `return_to` param on the login form path. So when the login fails, the `redirect_to` is not lost.

### DIFF
--- a/app/views/bikes/show.html.haml
+++ b/app/views/bikes/show.html.haml
@@ -220,7 +220,7 @@
           - if @bike.abandoned
             %p
               = t(".is_this_your_bike_send_proof_of_ownership")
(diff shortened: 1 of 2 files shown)


sethherr:
I'm merging it now, since it seems like it's good to go.

Issue #952, opened 2019-06-27
Add live (or nearly live) currency conversion


sethherr:
[Initial discussion in #951](https://github.com/bikeindex/bike_index/pull/951#discussion_r298286911)

Not a high priority right now, but it might be someday.

jmromer:
related: https://github.com/RubyMoney/money#implementations

PR #1591, opened 2020-05-17
Add exchange rate model for currency conversions

jmromer:
This patch implements a lingering `Money`-related `TODO` from `MoneyHelper`, adding a currency conversion backend that updates an `exchange_rates` table daily.

```rb
# app/workers/update_exchange_rates_worker.rb L3-14 (9e2c8b56)

class UpdateExchangeRatesWorker < ScheduledWorker
  prepend ScheduledWorkerRecorder
  sidekiq_options queue: "low_priority", retry: false

  def self.frequency
    1.day
  end

  def perform
    ExchangeRateUpdator.update
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/9e2c8b56/app/workers/update_exchange_rates_worker.rb#L3-L14)]</sup>

```rb
# app/services/exchange_rate_updator.rb L3-15 (f55ebd33)

class ExchangeRateUpdator
  def self.update
    payload = ExchangeRateApiClient.new.latest
    rates = payload.fetch(:rates)
    base_iso = payload.fetch(:base)

    results = rates.map do |target_iso, multiplier|
      ExchangeRate.add_rate(base_iso, target_iso, multiplier)
    end

    results.none?(&:nil?)
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/f55ebd33/app/services/exchange_rate_updator.rb#L3-L15)]</sup>

```rb
# app/models/exchange_rate_api_client.rb L3-6 (d1f7e639)

class ExchangeRateApiClient
  attr_accessor :base_iso, :base_url, :cache_key

  BASE_URL = ENV.fetch("EXCHANGE_RATE_API_BASE_URL", "https://api.exchangeratesapi.io")
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/d1f7e639/app/models/exchange_rate_api_client.rb#L3-L6)]</sup>

### post-deploy

```
bundle exec rails exchange_rates_update
```

```rake
# lib/tasks/rake_tasks.rake L62-66 (f55ebd33)

task exchange_rates_update: :environment do
  print "\nUpdating exchange rates..."
  is_success = ExchangeRateUpdator.update
  print is_success ? "done.\n" : "failed.\n"
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/f55ebd33/lib/tasks/rake_tasks.rake#L62-L66)]</sup>

### USD
USD

### EUR (pre-update)
EUR pre-update

### EUR (post-update)

EUR post-update

Resolves https://github.com/bikeindex/bike_index/issues/952

### Alternative implementation

For a less invasive implementation, we can instead merge #1592, which uses the eu_central_bank gem. One drawback there (arguably. arguably not, considering we don't deploy to ephemeral instances) is that the gem uses an XML file stored on disk to cache fetched conversion rates.

The APIs used in both implementations are free.

### DIFF
--- a/app/controllers/admin/exchange_rates_controller.rb
+++ b/app/controllers/admin/exchange_rates_controller.rb
@@ -1,0 +1,76 @@
+# frozen_string_literal: true
+
+class Admin::ExchangeRatesController < Admin::BaseController
+  include SortableTable
+
+  before_action :find_exchange_rate, only: %w[edit update]
+
+  def index
+    @exchange_rates =
+      ExchangeRate.where(filter_params).order(sort_column => sort_direction)
+  end
+
+  def new
+    @exchange_rate = ExchangeRate.new
+  end
+
+  def create
+    @exchange_rate = ExchangeRate.new(exchange_rate_params)
+
+    if @exchange_rate.save
+      redirect_to admin_exchange_rates_url
+    else
+      flash.now[:error] = @exchange_rate.errors.full_messages.to_sentence
+      render :new
+    end
+  end
+
+  def edit; end
+
+  def update
+    if @exchange_rate.update(exchange_rate_params)
+      redirect_to admin_exchange_rates_url
+    else
+      flash.now[:error] = @exchange_rate.errors.full_messages.join("\n")
+      render :edit
+    end
+  end
+
+  def destroy
+    exchange_rate = ExchangeRate.find_by(id: params[:id])
+
+    if !exchange_rate&.destroy
+      flash[:error] = "Could not delete exchange rate."
+    end
+
+    redirect_to admin_exchange_rates_url
+  end
+
+  private
+
+  def exchange_rate_params
+    params.require(:exchange_rate).permit(:from, :to, :rate)
+  end
+
+  def find_exchange_rate
+    @exchange_rate = ExchangeRate.find(params[:id])
+  end
+
+  def sortable_columns
+    %w[to from rate updated_at]
+  end
+
+  def default_direction
+    "asc"
+  end
+
+  def filter_params
+    params
+      .permit(:search_to)
+      .to_h
+      .map { |k, v| [k[/(?<=search_).+/].to_sym, v] }
+      .to_h
+  end
+
+  helper_method :filter_params
+end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::Base
   protect_from_forgery
 
   around_action :set_locale
+  rescue_from Money::Bank::UnknownRate, with: :localization_failure
 
   ensure_security_headers(csp: false,
                           hsts: "max-age=#{20.years.to_i}",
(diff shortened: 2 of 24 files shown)


sethherr:
Hey, two questions/requests:

- Could you make an admin view that would give us the ability to see current exchange rates displayed, and their updated times?
- Could you add Canada as a currency, with tests?

Also - are there any drawbacks to this that you can see or think of? Any places that this could cause problems or break?

jmromer:
No prob! Admin view added in e9a1f5cb

Can you elaborate on what you mean by "add Canada as a currency"? The updator already sources the USD to CAD exchange rate, if that's what you mean.

The main point of failure with either implementation is the application code requesting an exchange rate that isn't available, in which case an exception is raised:

```rb
# spec/helpers/money_helper_spec.rb L59-64 (d641a1fd)

context "given an unknown exchange rate" do
  it "raises UnknownRate" do
    expect { money_usd(1.00, exchange_to: :CAD) }
      .to(raise_error(Money::Bank::UnknownRate))
  end
end
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/d641a1fd/spec/helpers/money_helper_spec.rb#L59-L64)]</sup>

Currently the only place we convert currencies is on the landing page when displaying recoveries value in EUR:

```haml
-# app/views/welcome/index.html.haml L2-6 (49eea684)

%header.jumbotron.root-landing-banner-header.bg-image-bike-hero
  .container
    .root-landing-header-text
      %h2= t(".value_of_bikes_recovered", value: as_currency(Counts.recoveries_value))
      %h1.upcase= t(".hero")
```
<sup>[[source](https://github.com/bikeindex/bike_index/blob/49eea684/app/views/welcome/index.html.haml#L2-L6)]</sup>

```
bike_index money-ar % rg -l as_currency
app/views/welcome/index.html.haml
app/helpers/money_helper.rb
spec/helpers/money_helper_spec.rb
```

sethherr:
Thanks for the explanations! I wasn't sure what the difference was between this and #1592.

Just verifying (only necessary to say something if this is inaccurate) - this is not in any way touching places where we collect money or show charges to people (e.g. [cities_packages](https://bikeindex.org/cities_packages)) - this is only for the homepage.

If so, I think having extra flexibility would be nice - and being able to display Canadian dollars if relevant for the user would be cool, since we're building up a strong user base in Canada. If #1592 won't give us that ability, then this is the better solution.

When would the `Money::Bank::UnknownRate` error be raised, and would it be possible to fallback to USD instead of 500ing for a user, if that would be the result of throwing the error?

jmromer:
> this is not in any way touching places where we collect money or show charges to people

Correct (at least currently, we don't convert currencies for transactions)

> being able to display Canadian dollars if relevant for the user would be cool, since we're building up a strong user base in Canada.

Gotcha, thanks! I'll make an issue for that since I could see that growing.

> When would the Money::Bank::UnknownRate error be raised

Given the current state of the application, I don't think it's likely to happen in practice. Once the exchange rates are seeded the first time, someone would have to delete the USD -> EUR exchange rate record for that exception to be raised, which would have to happen in the console.

At the DB level, the `exchange_rates` table has no associated `on_delete: :cascade` instructions  and the model has a `before_destroy` callback defined to stop deletions of rates required for 18n, so the only way I can see it happening is via the console with a `skip_callbacks` invocation. It would take some doing, in other words.

> would it be possible to fallback to USD instead of 500ing for a user, if that would be the result of throwing the error?

Probably 🤔 I ran into some complications attempting that because of where `Money.default_bank` is expected to be set (the class instance level of `MoneyHelper`). But I'll give it another shot — I just worked around it the first time.

jmromer:
No worries — added one in 14856a0

Screen Shot 2020-05-21 at 3 17 23 PM

PR #1639, opened 2020-06-22
Rename master branch to main

sethherr:
See https://tools.ietf.org/id/draft-knodel-terminology-00.html#rfc.section.1.1.1

There are a lot of additional references, [here is one](https://www.zdnet.com/article/github-to-replace-master-with-alternative-term-to-avoid-slavery-references/)


sethherr:
Also of note: translations aren't synced in this branch, to test that the translation sync process still works post update.

PR #1646, opened 2020-06-24
Fix future date issue

sethherr:
The month was off by 1 when translating the date!

Running this in console and getting the shown responses back was wild!

```ruby
Time.current
=> "Wed, 24 Jun 2020 11:20:48 CDT -05:00"
I18n.l(Time.current)
=> "Wed, 24 Jul 2020 11:02:51 -0500" # wtf?
I18n.l(Time.current.beginning_of_year)
=> "Wed, 01 Feb 2020 00:00:00 -0600" # WHOAH, WHAT!!?
I18n.l(Time.current, format: :dotted)
=> "2020.6.24" # and this is CORRECT!
```

I found an answer in this [stack overflow question](https://stackoverflow.com/questions/15897171/i18n-l-displays-wrong-month-one-month-later).

In PR #1630 I removed the nil columns because they were causing problems with the translation normalizing. Reverting that - putting a nil row back in `month_names` and `abbr_month_names` fixed the issue.


sethherr:
So... still running into the issue that led to deleting those nil values - where my local `bundle exec bin/rake prepare_translations` doesn't normalize in the same way that CircleCI does.

I tried to figure out a quick fix but couldn't. I opened an issue in the i18n-tasks gem.

PR #1647, opened 2020-06-24
CircleCI 2 parallelism

sethherr:
Try out returning to parallelism 2

In https://github.com/bikeindex/bike_index/pull/1642 I switched to parallelism 3, because I was worried that jobs were taking too long to process. Unfortunately, this means there is only space for 1 job at a time (with our 4 workers).

So I'm going to benchmark - rerunning this 10 times on CircleCI and posting the results here.

Related: #1648 (parallelism 4)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     working_directory: ~/bikeindex/bike_index
-    parallelism: 3
+    parallelism: 2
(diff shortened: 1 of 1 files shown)


sethherr:
## Runs

(only the build time)

1. 06:26
2. 09:07
3. 07:43
4. 05:29
5. 05:54
6. 07:05
7. 05:35
8. 06:21
9. 10:17
10. 04:55

PR #1648, opened 2020-06-24
CircleCI 4 parallelism

sethherr:
Try out parallelism 4

In #1642 I switched to parallelism 3, because I was worried that jobs were taking too long to process. Unfortunately, this means there is only space for 1 job at a time (with our 4 workers).

So I'm going to benchmark - rerunning this 10 times on CircleCI and posting the results here.

Related: #1647 (parallelism 2)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   build:
     working_directory: ~/bikeindex/bike_index
-    parallelism: 3
+    parallelism: 4
     shell: /bin/bash --login
     environment:
       RAILS_ENV: test


sethherr:
## Runs

(only the build time)

1. 05:49
2. 
3. 
4. 
5. 
6. 
7. 
8. 
9. 
10.

sethherr:
Seems like the timing has more to do with the time of day than the parallelism - though more parallelism definitely helps.

Only being able to run one job at a time, with only nominal benefits in speed, seems like it's not worth it.

I'm sure there's a way to improve the parallelism performance, but doesn't seem important. Closing in favor of #1647

PR #1687, opened 2020-07-24
Add shop display for appointments

sethherr:
This replaces #1662 (and pulls in some of the changes from it).


sethherr:
Should use #1746 instead.

Issue #1255, opened 2019-09-12
Re-add strava authentication


sethherr:
Strava authentication is broken.

It looks like the gems are unmaintained [gem in Bike Index now](https://github.com/thogg4/omniauth-strava), [maybe official gem](https://github.com/tildedash/omniauth-strava-oauth2).

Because this is causing problems for our users, I hid the auth button in #1254

At some point we may want to re-add it (or we can close this PR if it no longer seems relevant)

sethherr:
No longer a priority

realchrisolin:
Any chance you could query the prod DB to see how many users had Strava OAuth set up? I was using it to login and I doubt I'm the only one. While I'm smart enough to know to request a password reset email to get back into my account, I feel like this going to inadvertently lock users out of their accounts temporarily. I figure most people are going to try logging in to report a stolen bike and the experience of not being able to (seemingly) log back in to their account will be understandably frustrating.

PR #1732, opened 2020-08-17
Re-add strava integration

sethherr:
Currently seems to be blocked because [Strava stopped returning emails](https://developers.strava.com/docs/changelog#january-17-2019)

- https://groups.google.com/g/strava-api/c/qS7BH0EvDoU
- https://github.com/dblock/strava-ruby-cli

Related: #1255


sethherr:
Closing because it doesn't seem like strava is going to change this.

PR #1746, opened 2020-08-27
Shop display using handlebars

sethherr:
Use [handlebars](https://handlebarsjs.com/) for templating


sethherr:
Closing because we aren't doing this right now - and if we decide to do this, this PR will require significant re-working.

Issue #1844, opened 2020-12-11
Trying to build compose from main is failing. 


joshchernoff:
Just forked and trying to spin up on main  

```
bike_index main % docker-compose build --no-cache app
Building app
[...]
Step 8/9 : RUN gem install bundler --conservative && bundle install
 ---> Running in 756836816ff0
Successfully installed bundler-2.2.0
1 gem installed
Your Ruby version is 2.5.8, but your Gemfile specified 2.7.2
ERROR: Service 'app' failed to build : The command '/bin/sh -c gem install bundler --conservative && bundle install' returned a non-zero code: 18
```

PR #1845, opened 2020-12-11
Bump the ruby version in the docker image to match the Gemfile.lock

joshchernoff:
Per #1844 trying to address the docker build ruby version mismatch and the Gemfile.lock

Is possible this PR is a noop and the solution is the fix the Gemfile.lock to build built with the correct version of ruby. 

I'm assuming the current version of ruby is 2.7 but just was not been updated in the respective docker files.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM ruby:2.5.8
+FROM ruby:2.7.2
 WORKDIR /src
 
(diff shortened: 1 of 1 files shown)


sethherr:
Thanks!

You're correct, the ruby version should be 2.7

I will merge this once it passes tests

PR #1947, opened 2021-04-23
Small performance improvements

sethherr:


sethherr:
Unclear why circleci isn't running the job, it should pass now so merging

PR #2056, opened 2021-08-19
Add a few relevant things to .gitignore

willbarrett:
Adding a few little bits that other people might wind up with when setting up the project that shouldn't be in source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,6 +47,7 @@ yarn-debug.log*
 yarn-error.log
 .yarn-integrity
 .nvmrc
+package-lock.json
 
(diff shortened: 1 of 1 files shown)


sethherr (approved):
👍

willbarrett:
@sethherr I ignored `package-lock.json` as well as it isn't checked in - that one I'm not sure about. I think the yarn lock does something similar, so probably fine, but lmk if I should un-ignore it and and add it to the project.

sethherr:
yarn lock does the same thing! Use yarn instead!

PR #2062, opened 2021-08-21
Upgrade secure_headers gem

willbarrett:


sethherr:
The test that is broken is important, we need to be able to provide iframes off of our current domain.

I think there might be some CORs stuff that is important too.

Also FYI - this is one of the few places where I'm not confident about our test coverage, because of differences in between production and development, and this will warrant some click testing before and after shipping.

willbarrett:
Yes, it also appears that the basic usage of `secure_headers` has changed since the version that's currently in use - so, this will require re-implementation in order to support the upgrade. Usage seems consistent between the target version and most recent, so the right way forward here is likely to upgrade to most recent, add the missing test coverage, and re-implement, maybe not necessarily in that order.

PR #2063, opened 2021-08-21
Upgrade bootstrap gem

willbarrett:


sethherr:
FYI: this isn't that simple - see #1005 for the start.

There are _A BUNCH_ of style changes between the version of bootstrap that we have and `4.6` (the one you have here). Like, most of the custom SCSS will not work right.

Also, the way we want to move forward is via webpacker (we have bootstrap 4.3.1 in `package.json`, which is what is used in the admin styles).

If this gem blocks upgrading other important things, the easy way forward is to vendor the exact version of bootstrap that is used (`4.0.0.alpha4`) and remove the gem.

willbarrett:
Yup, didn't think it would be. Consider this PR an experiment to see what breaks where.

PR #2067, opened 2021-08-23
Ignore facebook specs when the FB ads integration is not configured

willbarrett:
I found that I had a couple of test failures due to `Facebook::AdsIntegration` being an undefined constant. I think it makes sense to ignore these things when the configuration is not done to allow folks to have a green suite out of the gate. Let me know if you disagree and I'll jump through the hoops to get it set up.


sethherr (approved):
👍

PR #2072, opened 2021-09-10
finish migrating bike registration_info

sethherr:


sethherr:
Ran these commands:

```ruby
BParam.where("(params -> 'bike' ->> 'student_id') IS NOT NULL").order("RANDOM()").any? { |bp| bp.created_bike&.registration_info&.dig("student_id")&.blank? }

BParam.where("(params -> 'bike' ->> 'bike_sticker') IS NOT NULL").any? { |bp| bp.created_bike&.registration_info&.dig("bike_sticker")&.blank? }
```

Both returned false, so I think migration is finished well enough

sethherr:
This now includes #2037

PR #2074, opened 2021-09-16
Bump ruby version

sethherr:


sethherr:
Not worth fixing right now

PR #2123, opened 2021-12-26
downgrade active_model_serializers

sethherr:
`active_model_serializers` after version `0.8` don't actually support caching - see Discourse Gemfile


sethherr:
As per #471 - add some caching in too!

sethherr:
Need to remove `gem 'grape-active_model_serializers'` and replace with just calling `ActiveModel::ArraySerializer`

PR #2139, opened 2022-01-03
rename cmodel_name attribute

sethherr:


sethherr:
replaced by #2140 (`model_name` is a reserved word in rails)

PR #2153, opened 2022-01-11
Fix logout redirect for Bikehub

willbarrett:
By default, redirect back to `bikehub.com` so that folks won't then get immediately redirected back to the Bikehub login page.


sethherr (commented):
LGTM

sethherr (approved):

PR #2178, opened 2022-02-01
Add Instagram accounts to user model, user edit/show pages

### LINKED ISSUE #813: Add instagram handle to users
It should have the same options as twitter handle and be editable in the same place ([/edit_my_account?edit_template=sharing](https://bikeindex.org/edit_my_account?edit_template=sharing))

This is full-stack - you'll have to add an migration to add `instagram` to the database and make it behave basically the same as the `twitter` attribute, including adding a form field - but, since you have the twitter attribute as an example it seems pretty approachable.

barefootford:
I'll look at this again in the next day or two, but I think we are mostly there.
Let me know if there is anything that is wildly or mildly off.

I did a little bit of simplifying on `#show_twitter_and_website` but it ends up producing the same thing (I think) without having to manually handle the html string. Happy to add a basic `User` test too that checks for the new columns, but didn't see anything equivalent right now so wasn't sure if you generally skip these safety specs, ie `User.new(...); expect(user.instagram).to be_present` etc

Screen Shot 2022-02-01 at 3 20 18 PM

Screen Shot 2022-02-01 at 3 19 44 PM

This closes #813

### DIFF
--- a/app/controllers/my_accounts_controller.rb
+++ b/app/controllers/my_accounts_controller.rb
@@ -109,9 +109,9 @@ class MyAccountsController < ApplicationController
     pparams = params.require(:user)
       .permit(:name, :username, :notification_newsletters, :notification_unstolen,
         :additional_emails, :title, :description, :phone, :street, :city, :zipcode, :country_id,
-        :state_id, :avatar, :avatar_cache, :twitter, :show_twitter, :website, :show_website,
-        :show_bikes, :show_phone, :my_bikes_link_target, :my_bikes_link_title, :password,
-        :password_confirmation, :preferred_language,
+        :state_id, :avatar, :avatar_cache, :twitter, :show_twitter, :instagram, :show_instagram,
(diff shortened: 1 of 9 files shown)


sethherr:
This is great @barefootford !

I do generally skip basic safety tests like `expect(user.instagram).to be_present` - but I'm a big fan of request specs. It would be great if you added a request spec for the update action that tested that `instagram` was assigned in the [bike_index/spec/requests/my_accounts_request_spec.rb](https://github.com/bikeindex/bike_index/blob/main/spec/requests/my_accounts_request_spec.rb).

If you wanted to go above and beyond, you could put the other sharing options in there (`website`, `twitter`, `show_website`, etc) since they aren't currently tested in request specs :)

barefootford:
Yep, a request spec sounds great. I'll update it later today

sethherr (approved):
Looks great!

(Sorry, I didn't catch the notification that this was ready for review, I'll set something up privately with you for better communication).

Is this ready to merge?

barefootford:
Yep ready to go 🐿

PR #2192, opened 2022-03-08
Grape Upgrade

willbarrett:


config/locales/en.yml
```diff
@@ -2221,7 +2221,7 @@ en:
     - Fri
     - Sat
     abbr_month_names:
-    - 
+    -
```
sethherr, line 2224: Ooof. This used to happen on my computer when normalizing things - I'm not sure why it changed, but now it doesn't add this whitespace. I used to go in to this file and manually correct it, there didn't seem to be an easy answer when I last looked at the gem.

willbarrett: Thanks. I was wondering what was up here. No idea why this is happening - I haven't touched this file. I'll do the hand-hacked version for now. Is it just adding an ending newline character?

sethherr: It's removing the space that's added on these two lines 😢

sethherr (approved):
LGTM!

PR #2248, opened 2022-05-27
Update sticker search code to search more predictably

sethherr:
- If it looks like you're searching from the start of the string, don't match things in the middle
- update this for lookup too


sethherr:
To migrate:

```ruby
BikeStickerBatch.where(code_number_length: nil).each do |batch|
  batch.update(code_number_length: batch.calculated_code_number_length)
end
BikeStickerBatch.all.each do |batch|
  batch.bike_stickers.update_all(code_number_length: batch.code_number_length)
end
```

PR #2257, opened 2022-06-20
Update circleci convenience image

sethherr:
Replaces #2205


sethherr:
Pipeline values require version 2.1 schema, workflows require 2 - seems like circleci is a little bit of a shitshow.

Reverting to 2, removing the pipeline parameters, requires manually searching cache-key in the config

PR #2269, opened 2022-06-23
Remove winebouncer

sethherr:


app/controllers/api/base.rb
```diff
@@ -10,18 +10,98 @@ def parameters(request, _)
...
+      end
+    end
+
+    # Before grape actions happen
+    def before
+      @endpoint_auth_declaration = env["api.endpoint"]&.options&.dig(:route_options, :authorizations)
+      return unless endpoint_protected?
+      @request = ActionDispatch::Request.new(env)
+      doorkeeper_authorize!
+      self.class.doorkeeper_access_token = doorkeeper_access_token
```
sethherr, line 86: This feels a little weird, but I'm not sure how to share this information in other places without doing this.

Thoughts?

willbarrett: Session storage could be another way to handle this

willbarrett (approved):

PR #2364, opened 2023-07-28
Update carrierwave_background to a new commit to remove warning

sethherr:
0.4.2, the current version of `carrierwave_backgrounder` (released in 2015), raises a bunch of warnings:

```
WARN: Job arguments to CarrierWaveStoreWorker do not serialize to JSON safely. This will raise an error in
Sidekiq 7.0. See https://github.com/mperham/sidekiq/wiki/Best-Practices or raise an error today
by calling `Sidekiq.strict_args!` during Sidekiq initialization.
```

https://github.com/lardawge/carrierwave_backgrounder/pull/307 offers a solution, this references the commit from that version.

PR #2366, opened 2023-07-28
Add test for calculated_mnfg_name

sethherr:
Add a test to catch the change from the Nokogiri update (see #2364 for discovery)

This spec should be failing (it fails locally), but circleci doesn't fail. It's time to switch off of circleci.

```
# ./spec/workers/organization_export_worker_spec.rb:205:in `block (4 levels) in <main>'
```

Error actually started in 47baf5cd7768ada692ed2e3f73cb3ef22248e5e2, about 1 month ago.

I'm not necessarily opposed to the new version, but not having it throw an error is unacceptable.


sethherr:
Seems like the `Gemfile.lock` change didn't break the CircleCI cache. Really makes me trust what's going on there.

Ultimately, I think I'm ok with this change, so I'm updating the tests to reflect that. BUT - the fact that this wasn't caught was _so_ unacceptable

sethherr:
WTF! It still doesn't fail. CircleCI is dead.

sethherr:
Arg, after a brief investigation it's clear switching of circleci will be a pain. So... dealing with it for now.

PR #2376, opened 2023-08-02
2355 User additional link

eld120:


eld120:
@sethherr this specific test fails in CI and locally (a lot of tests fail locally and I have a lot to dig through to understand what is going on) but this specific test and the ExchangeRateAPIClient class haven't really been changed in the past year. I'm unsure if/how this could be related or interact with my proposed changes - that isn't to say that I want to bypass or ignore this failing test as much as I'm curious to hear your thoughts on this potential regression. Thanks for the help.

`rspec ./spec/models/exchange_rate_api_client_spec.rb:7 # ExchangeRateAPIClient#latest returns a hash with :base and :rates keys`

Cheers!

eld120:
@sethherr  Let's give this a shot

sethherr (approved):
Looks good to me!

PR #2404, opened 2023-09-11
Include scope in authorize button from oauth applications page

sethherr:
Per [comment on forum](https://discuss.bikeindex.org/t/dev-cannot-authorize-oauth-endpoint/4476) - missing the scope param causes this link to fail.


aokj4ck:
Thank you!

PR #2408, opened 2023-09-19
Bump rails to 6.1.7.6

sethherr:
Printing out a bunch of error messages: 

```
ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:68: warning: already initialized constant Net::ProtocRetryError
ruby/2.7.0/net/protocol.rb:66: warning: previous definition of ProtocRetryError was here
ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:214: warning: already initialized constant Net::BufferedIO::BUFSIZE
ruby/2.7.0/net/protocol.rb:206: warning: previous definition of BUFSIZE was here
ruby/gems/2.7.0/gems/net-protocol-0.2.1/lib/net/protocol.rb:541: warning: already initialized constant Net::NetPrivate::Socket
ruby/2.7.0/net/protocol.rb:503: warning: previous definition of Socket was here
```

See: https://stackoverflow.com/questions/67773514/getting-warning-already-initialized-constant-on-assets-precompile-at-the-time - adding `gem "net-http"` fixed the issues.


sethherr:
Some minor change in how things were escaped caused a spec failure, because it was matching the escaped sequence.

I don't think this is a problem, so I just updated the matcher, but 🤷

PR #2410, opened 2023-09-19
Add DatabaseSelector

sethherr:
Use rails DatabaseSelector to set read mode.


sethherr:
It seems like manually specifying when to go into read mode has pretty well balanced the load - I no longer think switching this on is necessary.

PR #2448, opened 2023-10-24
Add log searcher service

sethherr:
- Read log files to find bike searches
- Put the searches in the analytics database


sethherr:
Based on benchmarking, ripgrep is 10x faster than normal grep.

I ran the grep with it all combined into one regex, it's faster (~15%) but it seems to miss some lines. Doesn't seem worth it for the initial version, it can be improved later.

PR #2476, opened 2023-11-05
Rename ParamsNormalizer to InputNormalizer

sethherr:
Also rename method `strip_or_nil_if_blank` to `string` (for consistency with `boolean`)


sethherr:
Failing spec is because of Daylight Savings Time

PR #2477, opened 2023-11-05
Migrate frame model worker, reassign manufacturer if other is found

sethherr:


sethherr:
Failure is from Daylight Savings Time

PR #2489, opened 2023-11-16
Fix bin/setup

sethherr:
Fix issue caused by #2424


eld120:
@sethherr This fixed the issue I was having in development - thanks!

PR #2505, opened 2023-12-04
Enabling Norwegian on the front-end

eld120:
Norwegian translation/localization files are live/not deleted - assuming tests path what else should we be wary of before this goes live?


eld120:
Working on updating missed tests

sethherr (approved):
LGTM!

PR #2509, opened 2023-12-07
Spanish Translation

eld120:
Initial config and initial tests for Spanish Translation of Bike Index


eld120:
We're going to need to run `rake translation:sync` before the new tests for the Spanish translation will pass, we'll then need to commit the updated translation and localization files. A little different than the Norwegian translation but should work per the translation.io docs

sethherr:
Synced!

Please do the same thing that was done with the norwegian translations, where the option was hidden from the frontend - and then we can merge this PR.

PR #2510, opened 2023-12-09
Translation choices/nomenclature

eld120:
Wasn't paying close enough attention previously to the translation choices nomenclature. Changed to the following:

```
Native Spelling (localization setting)

# for example if localization is set to 'en'
English 
Nederlands (Dutch)

# for example if localization is set to 'nl'
English (Engels)
Nederlands 
```
- The localized translation is always in parentheses for translation choices
-  The language choice before the parentheses is always written in it's native script and spelling so that native speakers can easily identify their preference


sethherr:
This PR looks like it updates the spec to be what we want it to be!

However, main is broken right now. I updated the spec to match the current situation (in #2512), and merged it.

sethherr:
Also, there are now some specs that need to be uncommented in `spec/requests/locale_detection_request_spec.rb`

eld120:
I looked at the spec in `spec/requests/locale_detection_request_spec.rb` and the tests that were commented out seemed to be mistakenly labelled as `nb/norwegian` where they should have been `es/spanish` - basing this off of the expected test results. Updated and running tests

eld120:
@sethherr it's been a while, I've pushed a couple new translation targets to the config. I'd like to get back to writing tests for these translations before any new translations are merged but in the interim do you think you could run `rake translation:sync` on the `seya/translation-choices` branch so that our translators can start reviewing translations?

sethherr:
Done!

Thanks for your work @eld120!

eld120:
@sethherr I absolutely don't remember how to get translation.io to auto-populate the translation key/values - I'll track that down and loop you in as well as our contributors. 

Thanks for the help!

PR #2555, opened 2024-07-09
BIB-29: Update carrierwave due to code injection vulnerability [Draft]

torresga:
This PR upgrades `carrierwave` due to a security issue in the current version:
```
Name: carrierwave
Version: 0.11.2
CVE: CVE-2021-21305
GHSA: GHSA-cf3w-g86h-35x4
Criticality: High
URL: https://github.com/carrierwaveuploader/carrierwave/security/advisories/GHSA-cf3w-g86h-35x4
Title: Code Injection vulnerability in CarrierWave::RMagick
Solution: upgrade to '~> 1.3.2', '>= 2.1.1'
```

I originally updated to 1.3.4 as to limit the incompatibilities between the current and the new `carrierwave` version. However, there was an incompatibility in the `carrierwave` dependency `ssrf_filter`, which raised this OpenSSL error: `SSL_connect returned=1 errno=0 state=error: sslv3 alert handshake failure`. We couldn’t upgrade `ssrf_filter` because it was locked to be less than 1.1.0 due to a constraint in [`carrierwave` 1.3.4](https://rubygems.org/gems/carrierwave/versions/1.3.4).

We found https://github.com/arkadiyt/ssrf_filter/issues/53 , which gave us the idea that we would need to upgrade `ssrf_filter` to at least 1.1.1. Carrierwave ~> 2.2.0 updates `ssrf_filter` to 1.1.2, which fixed the bug present in `ssrf_filter` 1.0.8. This fixes the last two test failures we were seeing.

PR #2557, opened 2024-07-17
Update Carrierwave, second try

sethherr:
Redo of PR #2555 - which was reverted because it caused an error from Carrierwave Backgrounder.


sethherr:
Accidental merge, reopened with #2567

PR #2560, opened 2024-08-05
[Bib-40] Addressing send vulnerabilities

londerson:
Screenshot 2024-08-05 at 14 51 49

Brakeman was reporting 12 dangerous send vulnerabilities. The process here was to correct them using these practical strategies: https://brakemanscanner.org/docs/warning_types/dangerous_send/

After these changes, no more warnings were reported.

### DIFF
--- a/app/controllers/admin/feedbacks_controller.rb
+++ b/app/controllers/admin/feedbacks_controller.rb
@@ -33,7 +33,7 @@ class Admin::FeedbacksController < Admin::BaseController
     feedbacks = Feedback
     if params[:search_kind].present? && permitted_kinds.include?(params[:search_kind])
       @search_kind = params[:search_kind]
-      feedbacks = feedbacks.send(@search_kind) unless @search_kind == "all"
+      feedbacks = feedbacks.public_send(@search_kind) unless @search_kind == "all"
     else
       @search_kind = "all"
     end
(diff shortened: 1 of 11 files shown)


sethherr (commented):
All these sends are for whitelisted methods and the updated code doesn't do the same thing.

app/controllers/admin/feedbacks_controller.rb
```diff
@@ -32,8 +32,8 @@ def permitted_kinds
   def matching_feedbacks
     feedbacks = Feedback
     if params[:search_kind].present? && permitted_kinds.include?(params[:search_kind])
```
sethherr, line 34: `permitted_kinds` is a method that lists the permitted kind scopes that can be called on the Feedback model.

This line verifies that the method is whitelisted (that the `permitted_kind` includes it)

The value of `@search_kind` is a scope that is called directly on `Feedback` (not an array, not nil).

londerson:
That's true @sethherr It seems to be a false-positive. For some reason, Brakeman doesn't identify when the whitelisted methods aren't explicitly inline (we are delegating to functions that check hashes inside the model). Just in case, I switched from `send` to `public_send`, just to add one more safeguard level with the permitted methods.

I noticed that one test isn't passing through CI buildings. Running in my machine, sometimes this test fail randomly. Trying to looking a way to re-run on Circle CI I couldn't find this option available. How could we proceed?

Screenshot 2024-08-05 at 19 25 54

sethherr:
"Rerun failed tests" is setup, but I've found that sometimes CircleCI just refuses to use it. In those cases, I just click rerun from failed (which I just did)

sethherr (approved):

PR #2561, opened 2024-08-06
BIB-41 - Address dangerous redirects

londerson:
I'm focusing on solving redirects that where identified as dangerous by Brakeman, ensuring that is no risk of the client do any changes, explicitly marking it as a safe host to avoid breaches.

### DIFF
--- a/app/controllers/concerns/controller_helpers.rb
+++ b/app/controllers/concerns/controller_helpers.rb
@@ -141,22 +141,25 @@ module ControllerHelpers
       target = session[:return_to] || cookies[:return_to] || params[:return_to]
       session[:return_to] = nil
       cookies[:return_to] = nil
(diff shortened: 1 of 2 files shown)


mateusdeap (approved):

londerson:
We get a dangerous redirect warning on Brakeman, so it was necessary to ensure that it wasn’t possible to insert malicious parameters into the URL. One is a false flag, and for two of them, I added a safeguard. Because the method handle_target after conditional didn’t pass on CodeClimate, I did a refactoring. The lines that Brakeman was complaining about were 152 (added safeguard), 154 (false-positive) on app/controllers/concerns/controller_helpers.rb, and line 12 on app/controllers/discourse_authentication_controller.rb

PR #2564, opened 2024-08-15
Add spec to verify process_in_background for carrierwave_backgrounder

sethherr:
This is the test to demonstrate the problem with upgrading https://github.com/bikeindex/bike_index/pull/2557


sethherr:
It seems like the test still passes when carrierwave is upgraded ([PR for carrierwave upgrade](https://github.com/bikeindex/bike_index/pull/2557)), and it passes in click testing.

I'm not sure why this passes even though it didn't work in the upgrade - maybe something to do with image uploading vs local images?

You can upload an image at [admin/recovery_displays/new](http://localhost:3042/admin/recovery_displays/new) and test that it works (note, after you create the recovery display, you'll be directed to `admin/recoveries` - which won't include your new `recovery_display` - you'll have to navigate to `admin/recovery_displays` to see it).

sethherr:
Merging with the carrierwave upgrade because something is better than nothing.

PR #2569, opened 2024-08-18
Update Carrierwave, third try

sethherr:
Redo of PR https://github.com/bikeindex/bike_index/pull/2555 - which was reverted because it caused an error from Carrierwave Backgrounder.

This raises errors on production every time it's deployed.

---

I may just give up, because ultimately, I'd like to switch to active storage [using Cloudflare R2](https://kirillplatonov.com/posts/activestorage-cloudflare-r2/)

(which would mean I'd remove `carrierwave_backgrounder`, and do the migration to active storage after Rails is updated)

Here is [an article on migrating](https://huy-hoang.medium.com/from-carrierwave-to-active-storage-b2fd3e71407f), here is an article on [background processing image variants](https://dev.to/zimski/prevent-activestorage-to-kill-your-rails-app-18p3)


sethherr:
There is [this unresolved issue](https://github.com/carrierwaveuploader/carrierwave/issues/2739) for `NoMethodError: undefined method `copy'` for the carrierwave 3.0.7 - so I'm trying downgrading.

When deployed with Carrierwave 2.2.6, I saw this error:

> NoMethodError: undefined method `key' for nil:NilClass

[This stack overflow](https://stackoverflow.com/questions/64916838/undefined-method-key-for-nilnilclass-in-userscontroller-create) says adding `attr_writer :image_cache` solved the issue.

PR #2572, opened 2024-08-23
Conditionally add vehicle type to Partial registration form

sethherr:


sethherr:
somehow, not registering the latest commit :/

Replaced by #2574

PR #2573, opened 2024-08-23
[BIB-22] Reactivate Brakeman checks in CI

mateusdeap:
Previously we had configured brakeman to skip checks.

This PR removes that config, moves `brakeman.ignore` to the correct folder and ignores all false positives.


etagwerker (changes requested):
@mateusdeap We want the Brakeman check to display a red, failing mark if a PR is introducing insecure code. https://github.com/bikeindex/bike_index/blob/main/.circleci/config.yml#L22

config/brakeman.ignore
```diff
@@ -0,0 +1,1417 @@
...
+        "type": "method",
+        "class": "Organized::ImpoundClaimsController",
+        "method": "available_impound_claims"
+      },
+      "user_input": "params[:search_status]",
+      "confidence": "High",
+      "cwe_id": [
+        77
+      ],
+      "note": "@sethherr: All these sends are for whitelisted methods (from https://github.com/bikeindex/bike_index/pull/2560)"
```
etagwerker, line 92: Small nitpick here. Could you change this (and all other mentions) to "allowlisted"? Thanks!

etagwerker (approved):

PR #2578, opened 2024-09-03
BIB-45: Dual boot Rails 6.1 and Rails 7

torresga:
# What

This PR adds the dual boot code to dual boot both Rails 6.1 and Rails 7.0 locally and in CI.

Running rails c and rails s should work as expected with no errors.
The CI should run the tests for Rails 7.0.

You can append `BUNDLE_GEMFILE=Gemfile.next` to a command, i.e. `BUNDLE_GEMFILE=Gemfile.next rails c` to run whatever command using Rails 7.


.circleci/config.yml
```diff
@@ -554,7 +554,7 @@ jobs:
           name: Setup Database
           command: |
             bundle exec rails db:create --trace
-            bundle exec rails db:structure:load --trace
```
torresga: This will need to be changed in the `build` job when we make Rails 7.0 the only version, as `rails db:structure:load` was deprecated and Removed in Rails 7.0.

mateusdeap:
Looks good to me. Just make sure to mark it as ready for review

bin/setup
```diff
@@ -23,8 +23,8 @@ chdir APP_ROOT do
   system('bin/yarn')
 
   puts "\n== Preparing and seeding database =="
-  # NOTE: only loading db:structure on primary, because primary_replica is the same db in dev
-  system "bin/rake db:create db:structure:load:primary db:migrate db:seed"
+  # NOTE: only loading db:schema on primary, because primary_replica is the same db in dev
+  system "bin/rake db:create db:schema:load:primary db:migrate db:seed"
```
sethherr, line 27: You need to convert the existing `structure.sql` files to schemas

sethherr (changes requested):

bin/setup
torresga: Hi @sethherr, we don’t technically need to use a schema.rb file since the `rails db:schema:{load, dump}` tasks work with either Ruby or SQL (see https://github.com/rails/rails/commit/5488686851978c64346a0c03d2b6541cf2a94879#diff-1ed2907b3b8f148c2533558a77673ffa , the `config.active_record.schema_format` is set to `:sql` in `application.rb`.) If you would prefer that we switch back to using a schema.rb, we can invest time into that.

sethherr: Oh, cool! Good to know.

sethherr (approved):
I’ll test this out today

PR #2615, opened 2025-01-21
bump doorkeeper

sethherr:
Replaces #2454


sethherr:
Screenshot 2025-01-20 at 6 07 33 PM


Seems to be a change in error message - which is surmountable, but maybe not the problem that needs addressing immediately

PR #2619, opened 2025-01-21
Update grape-swagger

sethherr:


sethherr:
This doesn't seem easily possible.

Replaced by #2621 - which converts it to be static (so grape-swagger can be removed)

PR #2623, opened 2025-01-22
Post Ruby 3 upgrade cleanup

sethherr:
For the Ruby 3 upgrade (#2605), we added translation methods so that the upgrade would be backwards compatible with ruby keyword argument changes (see #2609, #2612 and https://github.com/bikeindex/bike_index/pull/2608)

This PR fixes the things that were changed there


sethherr:
Reverting the merge with #2667 - doing the find and replace in a different branch makes more sense.

PR #2626, opened 2025-01-22
Add Jsbundling

sethherr:
Currently, assets are broken because webpacker is no longer supported.

This tries to switch to jsbundling.


sethherr:
Seems like too big of a lift for something that we want to remove (and switch to importmaps).

Replaced by #2627

PR #2706, opened 2025-02-21
Install dartsass

sethherr:


sethherr:
Unable to get this to work, the build still fails locally with:

```bash
bin/rails dartsass:build --verbose --trace

Error reading app/assets/stylesheets/revised.scss : Cannot open file.
bin/rails aborted!
Command failed with exit 66: .asdf/installs/ruby/3.3.7/bin/ruby
```

Even when the revised.scss file is completely empty!

The other files compile OK, but for some reason I can't get that one to.

PR #2713, opened 2025-02-24
Rename TheftAlert to PromotedAlert

sethherr:


sethherr:
Actually, after doing this - I think `TheftAlert` more accurately describes what they are - considering we may have other varieties of promoted alerts in the future.

PR #2720, opened 2025-03-06
Updated stolen processor handles new stolen images

sethherr:


sethherr:
Closing and opening a PR that only includes the changes for stolen bike images (rather than also the header tag updates, which are now being handled separately in #2721)

PR #2730, opened 2025-03-10
Add knapsack pro

sethherr:
Add multiple CI nodes on GitHub Actions and use Knapsack Pro to handle test splitting


app/views/admin/stripe_prices/_table.html.haml
```diff
@@ -4,7 +4,7 @@
       %th
         = sortable "created_at", render_sortable: render_sortable
       %th
-        = sortable "membership_level", "Kind", render_sortable: render_sortable
+        = sortable "membership_level", "Level", render_sortable: render_sortable
```
sethherr, line 7: Small drive by fix!

PR #2787, opened 2025-04-19
Drop user address attrs

sethherr:
The Address Record model was added in #2780, and users were updated to use that. This removes the legacy address attributes on the User model.


sethherr:
Not ready for this yet, just seeing what breaks

PR #2804, opened 2025-04-23
Rename "search bikes" to "search" in primary menu

sethherr:
Also fix a primary menu height bug

### DIFF
--- a/spec/services/spreadsheets/manufacturers_spec.rb
+++ b/spec/services/spreadsheets/manufacturers_spec.rb
@@ -71,15 +71,15 @@ RSpec.describe Spreadsheets::Manufacturers do
         expect_target_manufacturer(manufacturer.reload)
       end
       context "with existing mnfg with alternate_name" do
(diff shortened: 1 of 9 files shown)


spec/services/spreadsheets/manufacturers_spec.rb
```diff
@@ -79,7 +79,7 @@ def expect_target_manufacturer(mnfg)
       end
       context "with no matching manufacturer" do
         let!(:manufacturer) { nil }
-        it "creates" do
+        it "creates", :flaky do
```
sethherr, line 82: Super annoying. Tried to figure it out twice - #2798 and #2799 but it still happens, so just mark it.

sethherr: Updated to create a manufacturer in the let block (rather than nil). Hopefully that solves the flakiness?

sethherr: didn't solve the flakiness. Marked it flaky

PR #2842, opened 2025-06-11
Assign end_at if marketplace listing isn't current

sethherr:


spec/requests/bikes/update_request_spec.rb
sethherr, line 1: Rather than fixing this, I'm going to switch to handling updates in the `MarketplaceListingsController` and handle it there. So ignoring for now.

PR #2856, opened 2025-06-19
Add search logging for marketplace searches

sethherr:
Also fix `LogSearcher::Reader` for `/search/registrations` which has been broken since it switched from `/bikes` 🤦


app/services/log_searcher/reader.rb
```diff
@@ -26,7 +26,7 @@ def searches_regex
     # If no time is included, it returns all the lines from the file
     def rgrep_command_str(time = nil, log_path: nil)
       log_path ||= DEFAULT_LOG_PATH
-      "rg '#{searches_regex}' '#{log_path}'" + time_rgrep(time)
+      "rg '#{searches_regex}' '#{log_path}'" + time_rgrep(time) + " | sort -u"
```
sethherr, line 29: `sort -u` should make a significant performance improvement - uniqing first.

PR #2872, opened 2025-08-05
Improve apt install in github actions

sethherr:
Prior to this PR, the "Install system dependencies" build step was regularly taking 1.5 minutes.

Caching all the apt-get with `awalsh128/cache-apt-pkgs-action` reduces that time to under 10 seconds.

🔥🎉


sethherr:
Just caching imagemagick (via `mfinelli/setup-imagemagick@v6`), the step time drops to under 30s

PR #2899, opened 2025-08-30
Add to URI tips about setting up URI redirect for bike_index_ios client

aokj4ck:
# Description

- This is necessary for deeplinks and OAuth sign-in to work for iOS clients
- Should link to https://github.com/bikeindex/bike_index_ios?tab=readme-ov-file#quick-start

### DIFF
--- a/app/views/doorkeeper/applications/_form.html.haml
+++ b/app/views/doorkeeper/applications/_form.html.haml
@@ -20,7 +20,9 @@
           Include
           %code
             = Doorkeeper.configuration.native_redirect_uri
-          for local tests
+          for local tests.
+      %p
+        Include <code>bikeindex:\/\/</code> for #{link_to "iOS app deeplinks", "https://github.com/bikeindex/bike_index_ios?tab=readme-ov-file#quick-start", tabindex: -1}.
       %p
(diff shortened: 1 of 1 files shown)


app/views/doorkeeper/applications/_form.html.haml
```diff
@@ -20,7 +20,9 @@
           Include
           %code
             = Doorkeeper.configuration.native_redirect_uri
-          for local tests
+          for local tests.
+      %p
+        Include <code>bikeindex:\/\/</code> for #{link_to "iOS app deeplinks", "https://github.com/bikeindex/bike_index_ios?tab=readme-ov-file#quick-start", tabindex: -1}.
```
sethherr, line 25:
```suggestion
        Include <code>bikeindex://</code> for #{link_to "iOS app deeplinks", "https://github.com/bikeindex/bike_index_ios?tab=readme-ov-file#quick-start", tabindex: -1}.
```

Should fix, currently it looks like this:

Screenshot 2025-09-01 at 1 10 45 PM

Thanks for adding `tabindex: -1`!

aokj4ck: Oh you're right, I was caught up in the iOS .xcconfig escaping (to prevent `//` from becoming a comment) and removing the escapes is right here, good catch

PR #2900, opened 2025-09-05
Fix: Compress assets with pngcrush

aokj4ck:
# Description

I want to reduce the bytes transferred in PNG assets to speed up using/developing Bike Index.

- Save 5_503_737 bytes
- This compression tosses metadata that could be important to preserve on a per-file basis
- Visually changes should be imperceptible and the 3 GitHub diff options should help

## Method

```bash
brew install pngcrush
find . -type f -name '*.png' -exec pngcrush -ow -rem allb -f 9 -brute -reduce "{}" \;
```


public/apple-touch-icon.png
aokj4ck, line 1: 55% size reduction for this asset

sethherr:
I need to verify that nothing useful is lost in this compression!

I already had processed most of the assets using ImageOptim, I'm surprised this resulted in such a large reduction!

sethherr (approved):
👍 🔥 🗜️

PR #2911, opened 2025-09-17
Added address_record for bike

sethherr:
Address record was created and added for Users in https://github.com/bikeindex/bike_index/pull/2780

This adds address record to Bikes

- [x] Update bike create to use address record (and have BParam handle it)
- [x] Update bike update to use address record, add test to verify it doesn't create duplicate address records

#2922 is the follow up issue to track cleaning up things after the backfill and migration

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -12,7 +12,8 @@ RSpec.describe User, type: :model do
     end
 
     it "is valid" do
-      expect(address_record.reload).to have_attributes target_attrs
+      # TODO: allow the user_id to include - #2922
+      expect(address_record.reload).to have_attributes target_attrs.except(:user_id)
       expect(address_record.to_coordinates.map(&:round)).to eq([52, 5])
       expect(user.to_coordinates).to eq(address_record.to_coordinates)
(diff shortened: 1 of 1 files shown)


app/views/bikes_edit/bike_details.html.haml
```diff
@@ -168,34 +168,11 @@
...
-                  = f.text_field :city, class: "form-control", required: true, placeholder: t(".city")
-
-              .form-group.row
-                = f.label :zipcode, class: "form-well-label"
-                .form-well-input
-                  = f.text_field :zipcode, class: "form-control", required: true, placeholder: t(".zipcode")
+            -# It's sus to make queries in a page - but I think it's more important to call this at the correct place
+            - @bike.find_or_build_address_record(country_id: current_country_id)
+            = f.fields_for :address_record do |address_form|
+              = render(LegacyFormWell::AddressRecord::Component.new(form_builder: address_form, no_street: !BikeServices::Displayer.edit_street_address?(@bike, current_user)))
```
sethherr, line 174: TODO: Verify that you don't want to pass in an organization here.

PR #2912, opened 2025-09-17
Add address_record to ownerships

sethherr:
Split from #2911 because this can be done entirely behind the scenes, prior to implementing address record on bikes.

... which will make adding address record on bikes easier


db/migrate/20250917185540_add_address_record_to_ownerships.rb
```diff
@@ -0,0 +1,7 @@
+class AddAddressRecordToOwnerships < ActiveRecord::Migration[8.0]
+  def change
+    add_reference :ownerships, :address_record, index: true
+    add_reference :bikes, :address_record, index: true
```
sethherr, line 4: NOTE: adding address_record to bike here (rather than in #2911) so that the migration happens before the logic is added to prevent attribute out of date issues when deploying the update.

PR #2923, opened 2025-09-22
Add ReActionView and Herb

sethherr:
- https://github.com/marcoroth/reactionview
- https://reactionview.dev/overview
- https://herb-tools.dev/overview


marcoroth:
Awesome, even better! Either way, let me know. I'm more than happy to help you adopt it as I'm looking to get more feedback from people adopting and using it. 

Plus, the fact that this repo is open source is tremendously helpful, so thank you for doing this in the open! 🙏🏼

PR #3335, opened 2026-03-31
Fix PSU sticker bug

sethherr:
not actually sure this fixes it, waiting on confirmation


sethherr:
Haven't heard more about this, closing

PR #3425, opened 2026-04-15
Add registration-sticker search to org multi search

sethherr:
Adds a **Serials / Registration Stickers** toggle to the org multi-search page (shown only when the org has `bike_stickers` enabled), so staff can paste a batch of sticker codes the same way they already paste serial numbers.

- Toggling swaps the placeholder, button label, and search endpoint, clears any current results/chips, and writes `search_kind` to the URL so a reload re-runs the matching search.
- Sticker search looks up codes across **all** organizations and renders matches through the existing registration-search table, redacting private fields for bikes outside the current org (same as cross-org serial results). Chips link to their result group; "view all" deep-links to the org stickers index for that code.
- The "Search all bikes" checkbox is forced on and disabled while stickers is the active kind (with a tooltip explaining why), since sticker search always spans every org; it re-enables on serial search.
- A query that normalizes to a blank code (e.g. a code-less `bikeindex.org/bikes/` URL) returns no matches, rather than falling through to `sticker_code_search`'s `all` and surfacing every org's bikes.
- Both kinds share one path: the results component takes `query:` + `search_kind:`, a single turbo_stream template serves both, `MultiResultChip` takes one resolved `label:`, and the JS runs each lookup through one `searchItem`. Sticker lookups reuse the `BikeSticker.claimed` scope, and the placeholder/button copy lives in one place (injected as Stimulus values).


sethherr:
## Screenshots

Screenshots of the Serials / Registration Stickers toggle (Hogwarts org).

### Desktop

**Serials (default)**

Multi search — Serials mode

**Registration Stickers** — the "Search all bikes" checkbox is checked + disabled, since sticker search always spans every organization

Multi search — Registration Stickers mode

### Mobile

**Serials (mobile)**

Multi search — Serials mode (mobile)

**Registration Stickers (mobile)**

Multi search — Registration Stickers mode (mobile)

PR #3430, opened 2026-04-17
Remove Functionable test monkey-patch

sethherr:
Drops `spec/support/functionable.rb`, the monkey-patch that made Functionable modules permissive enough for RSpec stubs and `stub_const`. Test-overridden values (Strava `STRAVA_DOORKEEPER_APP_ID`, `STRAVA_WEBHOOK_VERIFY_TOKEN`) are now read from `ENV` at call time so tests can set env vars instead of patching constants. `BlobUrl#local_storage?` checks the blob's service name directly, removing the need to stub `LOCAL_STORAGE`. Affected specs were restructured to avoid mocking Functionable modules: pushing real data into Redis (LogSearcher), using `geocoder_default_location` (GeocodeHelper), and asserting side effects rather than expecting method calls (Autocomplete::Loader, Images::StolenProcessor).

## Test plan
- [ ] CI passes


sethherr:
I think actually, I'm ok with breaking functionable a little bit in testing - since I think using constants instead of using ENV variables is a better solution

PR #3433, opened 2026-04-17
Add Cloudflare Turnstile to unauthenticated auth forms

sethherr:
Wraps the four unauthenticated POST endpoints attackers actually target — sign in, magic-link request, password-reset request, and resend confirmation — with a Cloudflare Turnstile challenge. Real users solve the invisible widget transparently; bots fail the verify call and get redirected back with a flash error. Token-based endpoints (\`sign_in_with_magic_link\`, \`update_password_with_reset_token\`) and signed-in actions (\`my_account#update\`, \`user_emails#resend_confirmation\`) keep rack_attack as their only defense — Turnstile adds nothing there. The widget and verification are no-ops unless both \`CLOUDFLARE_TURNSTILE_SITE_KEY\` and \`CLOUDFLARE_TURNSTILE_SECRET_KEY\` are set, so dev/test work unchanged.

## Test plan
- [ ] Set both env vars in production after creating the Turnstile site key in the Cloudflare dashboard
- [ ] Verify the widget renders and is solvable on each of the four forms
- [ ] Confirm a POST without a valid token is rejected with the bot-check flash error
- [ ] Confirm the in-flow resend-confirmation link still works for unconfirmed users (skipped via \`unless: unconfirmed_current_user.present?\`)

### DIFF
--- a/app/controllers/concerns/sessionable.rb
+++ b/app/controllers/concerns/sessionable.rb
@@ -51,6 +51,14 @@ module Sessionable
     store_return_and_authenticate_user(translation_key: :create_account, flash_type: :info)
   end
 
+  def verify_turnstile!
+    return true if Integrations::CloudflareTurnstile.verify(params[:"cf-turnstile-response"], ip: forwarded_ip_address)
+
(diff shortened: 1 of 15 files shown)


sethherr:
closing, not doing for now

PR #3438, opened 2026-04-21
Retire shared/_period_select partial in favor of UI::PeriodSelect

sethherr:
This PR finishes the migration to `UI::PeriodSelect::Component` (extracted in #3451) by deleting the legacy partial and its locale keys, fixing a `UI::ButtonLink` bug uncovered by the migration, and adding integration coverage that pinned down a `search_email` regression.

## Summary

- **Retire the partial.** Delete `app/views/shared/_period_select.html.erb` and the `shared.period_select.*` keys from all five locale files; migrate the remaining call sites (organized search, admin graphs subnav, etc.) to `UI::PeriodSelect::Component`.
- **Inline the one odd caller.** `organized/parking_notifications/index.html.haml` rendered the partial with `skip_submission: true` (a flag the component doesn't model); it's now inlined locally so the rest of the migration could complete.
- **Persist `search_email` through period clicks.** Add an integration spec exercising every period button + the custom date range against the organized registrations search; this also verifies that `?search_email=…&period=year` round-trips through the form.
- **Fix `UI::ButtonLink#html_class`.** The keyword was being captured into `@html_options` and read back as `@html_options[:class]` (always nil), so `period-select-standard` was silently dropped on the rendered link. Promoted to a real keyword arg, matching `UI::Button`.
- **Quiet a flaky CI test.** The select2 combobox spec hits the production autocomplete API and was timing out at 10s on GitHub Actions through both rspec-retry attempts; bump the wait to 30s (passes locally in <3s, so successful runs are unaffected).


sethherr:
Period select with the custom date range form expanded on `/o/hogwarts/bulk_imports`:

period_select_custom

PR #3487, opened 2026-05-01
Fix Doorkeeper cookie access to use AUTH_COOKIE_KEY

aokj4ck:
- In production environment cookie[:auth] is the right key
- but in development this key is mismatched and should use the dev key w/ the port number 

See:
> app/controllers/concerns/controller_helpers.rb:10:  AUTH_COOKIE_KEY = Rails.env.development? ? :"auth_#{ENV.fetch("DEV_PORT", 3042)}" : :auth
- now Doorkeeper uses the right key from ControllerHelpers `AUTH_COOKIE_KEY`


aokj4ck:
This does not change the test suite behavior for `:auth` symbols

PR #3515, opened 2026-05-09
Add Promoted section to marketplace index

sethherr:
Surfaces Bike Index member listings on the marketplace and promotes them on the index.

- **Promotes member listings**: listings whose seller has an active membership are partitioned to the top of `/search/marketplace` on page 1; standard listings paginate independently below. The section is hidden when there are none, and both partitions keep the existing `published_at` order.
- **"Bike Index member" badge**: each member listing shows a purple pill — right-aligned to the price on the index cards, and on the bike show page (right-aligned to the price on desktop, stacked above it on mobile).
- **Cached `seller_member`**: a denormalized boolean (with a partial index) on `marketplace_listings`, set when a listing is created or its seller changes. `AfterUserChangeJob` keeps existing current listings in sync, and `Membership` enqueues that job on status change so membership grants and expirations propagate.
- **Empty state**: the standard list's "no results" message is suppressed when only member listings matched the search.

----

TODO after shipping:

- `Backfills::MarketplaceListingSellerMemberJob.perform_async`


sethherr:
## Screenshots

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
|  |  |
| Member listings (badged) sort first, then standard listings, in one continuous grid — at this 3-column width, row 2 is 1 member + 2 standard | |

### /bikes/:id — member listing show page

| Desktop | Mobile |
| --- | --- |
|  |  |
| The badge right-aligns to the price on desktop and stacks above it on mobile | |

PR #3537, opened 2026-05-13
Skill docs: screenshots-in-comment + default org slug

sethherr:
Two unrelated skill doc tweaks bundled on this branch:

- `pr` skill: post the `## Screenshots` section as a PR comment rather than appending it to the description, so the body stays focused on the human-written summary. Re-runs locate the existing screenshots comment via the issues-comments API and PATCH it in place.
- `frontend-screenshots` skill: note that org-scoped URLs (`/o/<slug>/...`) default to the `hogwarts` slug when the caller doesn't specify one.


sethherr:
## Screenshots

### /about

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3553, opened 2026-05-15
Convert organized impound_records table to a view component

sethherr:
Replace the organized `impound_records/_table` partial with `Org::ImpoundRecordsTable::Component`, which renders through `UI::Table::Component`. Follows the same pattern as #3550.

- Per-row `canupdate-<kind>` classes on the multiselect column are load-bearing for `.multiselect-cell.canupdate-X input` in `application.js`'s multi-update behavior, so `UI::TableColumn` now lets `classes:` accept a callable that receives the record (`th_classes` skips callables — header classes still come from `header_classes:`).
- `current_organization` is captured as a template local so cell blocks can reach it across `UI::Table`'s `instance_exec` boundary.

PR #3559, opened 2026-05-16
Drive impound records index multi-update with Stimulus

sethherr:
Replaces the legacy vendored `application.js` multi-update behavior on the organized impound records pages with Stimulus.

- Two controllers: `org--impound-update` shows the kind-specific field for the selected update kind (shared by the index and the impound record show page); `org--impound-update-multi` adds the index-only extras — open/close panel toggle, checkbox column, per-kind checkbox enable/disable, empty-submit validation, and `multi_update` URL state.
- The update form now lives in the shared `Org::ImpoundRecordUpdateForm::Component`, so the show page is Stimulus-driven too.
- Index dropdowns → `UI::Dropdown`; the multi-update toggle → `UI::Button` and is now a real open/close toggle ("update multiple records" ↔ "hide update").
- Form validation: an empty submit shows a `UI::Alert` error, and `transferred_to_new_owner` requires the new-owner email.
- Per-row update kinds move from `canupdate-<kind>` classes to a `data-update-kinds` attribute; select-all reuses the shared `table-multi-checkbox` controller.
- Reverts #3553's `UI::TableColumn` callable-`classes:` support, now unused.

The legacy `BinxAppOrgImpoundRecords` still ships in the vendored bundle and runs on these pages; the id/class hooks it keyed on are renamed so it's an inert no-op.


sethherr:
## Screenshots

### `/o/:org/impound_records`

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### `/o/:org/impound_records?multi_update=true` — panel rendered open

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3560, opened 2026-05-16
Fix collapsed whitespace in RadioButtonGroup toggle labels

sethherr:
The org registration search status/sticker/address toggle buttons stopped showing spaces between words (e.g. `only not impounded` rendered as `onlynotimpounded`).

- `Form::RadioButtonGroup::Component` labels are `inline-flex`, so each text run and inline element (`<strong>`) became a separate flex item and the whitespace between them was stripped — wrap the label content in a `span` so it is a single flex item.
- Add a `with_html_labels` component preview and a spec asserting spaces around inline markup are preserved (existing previews only used single-word labels).

### DIFF
--- a/app/components/form/radio_button_group/component.rb
+++ b/app/components/form/radio_button_group/component.rb
@@ -49,7 +49,7 @@ module Form
             class: "tw:sr-only",
             form: @form,
             data: @data) +
-            option[:label].html_safe
+            tag.span(option[:label].html_safe)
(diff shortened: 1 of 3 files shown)


sethherr:
## Screenshots

### /rails/view_components/form/radio_button_group/component/with_html_labels

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3582, opened 2026-05-21
Replace admin fancy-selects with Form::Combobox

sethherr:
Converts the remaining single-value `fancy-select` dropdowns across `admin/` views to render `Form::Combobox::Component` (introduced in #3570), so admin selects use the new hotwire_combobox-based control instead of selectize.

- 10 admin views updated: payments, bulk imports, organization roles, organizations form, bike sticker batches (new/reassign), bikes edit (cycle type, manufacturer, color, creation org, stolen-record country/state), missing-manufacturer, social posts.
- Drops the `.fancy-select.unfancy` wrapper class on a `referral_source` text_field in `admin/payments/new` (was misapplied — not a select).
- Multi-select / grouped fancy-selects (`news/edit` content tags, `bikes/edit` bike_organization_ids, `organizations/index` features-and-settings) are deferred — hotwire_combobox multiselect needs a `multiselect_chip_src` endpoint that doesn't exist yet.
- Shared `bikes_edit/bike_fields/_primary_activity` partial untouched (used by user-facing edit too).

PR #3606, opened 2026-05-26
Fix admin org parent_organization_id combobox

sethherr:
The vendored `application_standalone.js` admin bundle's compiled `AdminEdit` controller (source `app/javascript/scripts/pages/admin/organization_form.js`, long gone from the repo) hard-coded `$('#organization_parent_organization_id').selectize().enable()` on construction. When the field was a `<select>` that was a no-op against an already-selectized element; #3582 turned it into a hw_combobox `<input>`, so the bare `.selectize()` *initialized* selectize on the combobox's text input — wrapping it in a `.selectize-control` inside the `.hw-combobox__main__wrapper`, stranding the chevron outside the inner input and routing keystrokes to the wrong field.

Fix: port the two behaviors AdminEdit owned (ambassador-kind disable, stolen-message-kind area reveal) to a new Stimulus controller at `app/javascript/controllers/admin/organization_form_controller.js`. Sidestep the legacy bundle's dispatcher by overriding `@page_id` to `"admin_organization_form"` for `Admin::OrganizationsController#new`/`#edit` — same pattern `InfoController` already uses — so the `#admin_organizations_(new|edit)` selector matches nothing and `new AdminEdit(form)` never fires. No vendored-bundle edits.

Also patches a race in `CollapseUtils`: when `show()` interrupted an in-flight `hide()`, `isVisible(el)` still returned true (the `tw:hidden!` finalizer hadn't fired yet) so `show()` bailed early, then the pending `setTimeout` from `hide()` landed `tw:hidden!` on the now-supposedly-visible element. Now both methods cancel any pending finalizer at entry, and `show()`'s early-return also checks for the in-progress `tw:scale-y-0` flag.


sethherr:
## Screenshots

### /admin/organizations/example-organization/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3633, opened 2026-05-30
Eager-load search results via turbo-frame src; drop JS auto-submit, fix initial kind counts

sethherr:
Switches every search form that uses the shared `search--form` Stimulus controller to load its results via a native eager `<turbo-frame src="…">` instead of a JS-driven form auto-submit. Each results frame gets a `src` (`request.original_fullpath`) on the initial shell render, so Turbo fetches its contents in a separate async request the moment the frame connects — right after the form, header, and page chrome have rendered.

- **Scope:** `search/registrations`, `search/marketplace`, and the organized `registrations`, `graduated_notifications`, and `impound_records` pages. Each controller already branches on `@render_results` (false for the HTML shell, true for the frame's own `turbo_request?` fetch), so one action serves both the shell and the eager-frame response. The form/combobox is skipped on frame requests (`unless turbo_frame_request?`) — Turbo discards it, so re-rendering it is wasted work.
- **JS cleanup:** removed the auto-submit machinery and the dedicated `popstate` handler. After back/forward, Turbo's snapshot restoration keeps the frame correct; `reloadFrameIfUrlStale` (on `turbo:load`) is the fallback that re-points a genuinely stale frame at the address bar. Kept the `search_no_js` removal, stale-`[busy]` clearing, time localization, and hidden-field sync. Pages whose form sits outside the frame opt out of Turbo's snapshot cache (`no-cache` meta) so back-nav re-fetches fresh.
- **Kind counts:** the auto-submit used to fetch the per-kind counts on connect; the eager flow no longer submits, so counts are fetched directly on connect — fixing the blank counts on the marketplace's initial render. Redundant fetches are deduped via a `data-counts-query` marker that rides along in Turbo's cached DOM.
- **No-JS:** without JS the eager `src` never fires, so results render server-side from the `search_no_js` param, and the loading spinner ships hidden (revealed by JS) so a no-JS user never stares at a spinner that can't resolve.
- **Tests:** integration specs now drive the real user path (visit `/`, click the nav link) and the no-JS fallback. The eager `src` deliberately doesn't rewrite history on initial load, so a default search sits at the clean base path.


sethherr:
@/tmp/sc_body.md

PR #3637, opened 2026-06-02
Re-initialize legacy JS on turbo:load so the parking notification form opens

sethherr:
The "New parking notification" button on the bike show page didn't open after a Turbo navigation. The legacy `BikeIndex` page scripts initialize only on `$(document).ready`, which Turbo doesn't fire when it swaps `<body>` on a visit or snapshot restore — so `BikesShow` never re-instantiated and the button's click handler was never bound. This also initializes on `turbo:load`, guarded by a plain variable so the two don't double-bind on the initial load (where both fire) and cleared on `turbo:before-render` so each visit/restore re-initializes; `$(document).ready` still covers pages that don't load Turbo, such as the doorkeeper layout. The integration spec now creates a parking notification through the bike show UI before exercising the index filters.


sethherr:
## Screenshots

### /bikes/50 — "New parking notification" form, opened

This PR changes only JS init timing (`init.coffee`), not the form's markup, so there's no visual difference from `main` — these just confirm the form opens. The fix is behavioral: after a Turbo navigation `BikesShow` now re-initializes via `turbo:load` and binds the button's click handler (previously it bound only on `$(document).ready`, which Turbo doesn't fire after a `<body>` swap).

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3646, opened 2026-06-04
Rework authorized applications index to use UI::Table

sethherr:
Reworks the `/oauth/authorized_applications` index to use `UI::Table` instead of the hand-rolled Bootstrap table.

- Fixes the mismatched column count — the header declared four columns but the body only rendered three.
- Renders `created_at` through `UI::Time` so the timestamp localizes to the viewer instead of a raw `strftime`.
- Loads the modern importmap JS in the doorkeeper layout so `UI::Time`'s `localizeTime` span actually converts (the layout previously loaded only legacy JS, which localizes `convertTime`); this also enables Stimulus on doorkeeper pages and fixes the same latent issue on `/oauth/applications`.
- Converts the view from the deprecated HAML to ERB.


sethherr:
## Screenshots

### /oauth/authorized_applications

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3653, opened 2026-06-05
Consolidate search index views into Search::ResultsFrame::Component

sethherr:
Dedup pass over the search index views, which had accumulated copy-pasted Turbo scaffolding.

- **Results-frame ViewComponent:** the `search-results-frame-wrapper` skeleton (eager turbo-frame + no-JS hidden spinner + loading overlay) was repeated across all five search views. Extracted `Search::ResultsFrame::Component`; each view passes `frame_id`/`loading_text` and keeps its per-page chrome and results partial in the render block.
- **No-cache opt-out, now uniform:** every page using the component has the same form-outside-frame + `turbo_action="advance"` pattern, so any of them can restore a stale frame on back/forward. The component now emits the `turbo-cache-control` no-cache meta unconditionally — extending the opt-out to `search/registrations`, `search/marketplace`, and `organized/registrations`, which previously (inconsistently) lacked it. Verified by removing it: the back-nav specs restore stale form/frame state without it.

Per-page chrome (period select, render-chart link) now sits inside the component's `@render_results` branch, so it no longer emits into the JS shell's throwaway frame placeholder — the eager `src` fetch replaces it regardless, so no user-visible change; the no-JS path renders chrome + results as before.

Covered by a component spec and the existing integration specs across all five pages (JS + no-JS), with an assertion that the no-cache meta reaches the head.

### DIFF
--- a/app/components/search/results_frame/component.html.erb
+++ b/app/components/search/results_frame/component.html.erb
@@ -1,0 +1,15 @@
+<div class="search-results-frame-wrapper">
+  <%= helpers.turbo_frame_tag @frame_id, src: @src do %>
+    <% if @render_results %>
+      <%= content %>
+    <% else %>
+      <div hidden data-search-loading>
+        <%= render UI::LoadingSpinner::Component.new(text: @loading_text) %>
+      </div>
+    <% end %>
+  <% end %>
+
+  <div class="search-loading-overlay">
+    <%= render UI::LoadingSpinner::Component.new(text: @loading_text) %>
+  </div>
+</div>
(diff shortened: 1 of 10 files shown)


sethherr:
## Screenshots

Pure refactor — output is identical; these document visual parity (`main` 👆 / this branch 👇).

### /o/cannondale/impound_records

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/cannondale/graduated_notifications

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3659, opened 2026-06-06
Surface 429 throttling during search

sethherr:
Active searching can trip Rack::Attack's throttle, and a 429 went unsurfaced: the eager results turbo-frame just left a stuck spinner (Turbo only logs the error) and the kind-count fetch swallowed it silently, leaving the tab counts blank.

- Added a hidden rate-limit notice above the results frame (`Search::ResultsFrame::Component`) that's revealed whenever a search request comes back 429, and hidden again on the next successful render.
- `search--form` now intercepts the results frame's 429 via `turbo:before-fetch-response`, shows the notice, and clears the dangling spinner.
- The kind-count fetch checks for 429, signals the notice through a `search:rate-limited` window event, and gained a `.catch` so other errors no longer reject silently.

### DIFF
--- a/app/components/search/results_frame/component.en.yml
+++ b/app/components/search/results_frame/component.en.yml
@@ -1,0 +1,6 @@
+---
+en:
+  components:
+    search:
+      results_frame:
+        rate_limited: Too many searches too fast — you're being rate limited. Wait a moment, then try again.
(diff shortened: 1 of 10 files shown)


sethherr:
## Screenshots

The rate-limit notice (revealed when a search request comes back 429):

### /search/registrations

| Desktop | Mobile |
| --- | --- |
| search rate-limit alert — desktop | search rate-limit alert — mobile |

PR #3679, opened 2026-06-10
CI: consolidate test matrix into one Test results check

sethherr:
Replaces the commented-out EnricoMi aggregate job with a single `Test results` check that merges every matrix node's junit XML into one `GITHUB_STEP_SUMMARY`.

- One stable check (`Test results`) to point branch protection at, instead of the per-node matrix names — its name survives resizing the matrix.
- Renders a totals table (per-node + matrix total) and fails if any node failed or any junit file is missing.
- Needs no write permissions, so forked PRs work; per-failure PR annotations still come from `RSpec::Github::Formatter`.
- Each failing example gets a collapsible block with its **full backtrace** (from the junit `<failure>`/`<error>` text, not a truncated message), so every failure across the whole matrix is readable in one place.


sethherr:
Meh, doesn't really make things better

PR #3685, opened 2026-06-10
Replace hotwire-livereload with Mata

sethherr:
Replaces `hotwire-livereload` with [Mata](https://railsdesigner.com/meet-mata/), which live-reloads via Rack middleware (SSE + idiomorph DOM morphing) instead of an ActionCable channel and per-layout `hotwire_livereload_tags`.

- A thin `MataExceptLookbook` subclass skips script injection on `/lookbook`, which has its own reloader — so we no longer need to pin the gem version (the prior reason for the `~> 1.4.1` pin and the dependabot ignore, both removed).
- Dropped `hotwire_livereload_tags` from the application, payments, and admin layouts; the middleware injects globally, so admin pages now reload too.
- Mata's file-watcher thread + in-memory SSE registry must live in the request-serving process, so development now runs **single-process Puma**: Puma's worker count is driven by `WEB_CONCURRENCY`, which `bin/env` defaults to `0` locally (deployed envs set their own — staging `1`, production `3`/Cloud66 — and keep `preload_app!`).
- The old `.env WEB_CONCURRENCY=1` never reached Puma (Puma parses its config before dotenv loads `.env`), so the default moved to `bin/env`, which is exported into the shell before Puma boots. `bin/env` is also no longer loaded in staging — only production was excluded before, but it sets dev/test-only defaults.

Verified with Playwright: editing a view morphs the page in place (`[Mata] Page morphed successfully`), `/lookbook` is untouched, and the dev server boots single-mode via plain `bin/dev`.


sethherr:
Closing because hotwire-livereload works with JS assets and I don't think this does.

PR #3751, opened 2026-06-27
Drop the dead Integration model

eld120:
Removes the dead `Integration` model. It backed the old Facebook/Strava/GlobalID OmniAuth social login — OmniAuth is gone and the login buttons are long commented out, so nothing creates `Integration` records anymore and the table and model are vestigial.

- Delete the model, its table (migration), spec, factory, and the Facebook/Strava auth-hash fixtures. `integration_data_tweet.json` stays — it's used by `SocialPost`, unrelated to this model.
- Drop the now-unused `User has_many :integrations` and the merge-job line that moved integrations between users.
- Repoint the credibility scorer's `user_connected_to_strava` badge to the live `StravaIntegration` association. It previously read the dead `Integration` records, so it was effectively never awarded — this restores the intended behavior (note: the badge starts being awarded again).


app/services/credibility_scorer.rb
```diff
@@ -132,7 +132,7 @@ def self.user_badges(user)
     badges += [:user_sent_in_bike_tip] if Feedback.where(user_id: user.id).stolen_tip.any?
     badges += [:user_supporter] if user.payments.any?
     badges += [:long_time_user] if user.created_at < Time.current - 2.years
-    badges += [:user_connected_to_strava] if user.integrations.strava.any?
+    badges += [:user_connected_to_strava] if user.strava_integration.present?
```
sethherr, line 135: Oooh, nice addition 🙂

sethherr (approved):

PR #3754, opened 2026-06-27
Seed Ctype secondary_name values to match live data

sethherr:
Seed `Ctype` and `Cgroup` data so a fresh `db:seed` matches the live component names — restoring the `secondary_name` values the create loop was silently dropping, and normalizing display casing.

- Set `secondary_name` when creating ctypes and add the six missing values (Ebike battery, Ebike motor, Pannier, Cargo bike side rails, seat bag, Top tube bag).
- Title-case component type and group names, leaving `"unknown"` (`Ctype.other`) and `"Additional parts"` (`Cgroup.additional_parts`) lowercase since both are matched by exact string in app code.


sethherr:
Solved better by #3756

PR #3758, opened 2026-06-28
Add SAML 2.0 SSO Service Provider (config, metadata, login)

eld120:
Adds a generic, multi-tenant SAML 2.0 Service Provider so an organization can log its members in through its own IdP. Org-agnostic — configured per-organization from the admin panel.

## What's included

- **Models**: `OrganizationSamlConfiguration` (per-org IdP entityID / SSO URL / signing cert, behind the `saml_sso` feature) and `SsoIdentity` (links an IdP subject to a Bike Index user for account linking).
- **SP endpoints**: published SP metadata (`/sso/:org_slug/metadata`), SP-initiated login (`/sso/:org_slug/init` → IdP), and the ACS callback (`/sso/:org_slug/callback`) with full assertion validation via `Saml::AssertionProcessor`. The shared SP keypair comes from ENV.
- **Provisioning**: an assertion for an email with no Bike Index account provisions and links one through the passwordless-org path — org role granted, no email sent; an assertion for an email that already has an account links to it. `saml_sso` triggers that behavior through `Organization#passwordless_user_creation?` rather than granting the `passwordless_users` feature, so an SSO org doesn't pick up an unrelated feature (or its "passwordless" admin labels) as a side effect.
- **Login integration**: SSO plugs into the identifier-first login (#3791) — an SSO-managed email entered on `/session/new` is handed straight to the IdP. One `before_action` covers identify/create/create_magic_link, so no password is ever checked and no new magic link is ever issued for such an email.
- **One SSO org per email domain**: without it `saml_email_matching` picks by name order, silently routing logins to the wrong org's IdP. Non-SSO orgs may still share a domain.
- **Local verification**: `docs/sso-local-testing.md` points at the throwaway Keycloak IdP in [bikeindex/saml-idp-test](https://github.com/bikeindex/saml-idp-test), whose README is the runbook. Deterministic coverage lives in the request specs, which sign assertions in-process with no container.

## Known gaps, deliberately left for follow-ups

- Members who already had a Bike Index account before their org enabled SSO get signed in and linked, but not added to the org — only newly provisioned accounts get the org role.
- `sign_in_with_magic_link` submits a token rather than an email, so the forced-SSO redirect has nothing to match on and can't cover it. A link issued before the org moved to SSO still signs in until that token expires.
- Production SP keypair provisioning.


eld120:
## Screenshots

### /admin/organizations/sso-test/edit

| Desktop | Mobile |
| --- | --- |
| admin organization edit, desktop | admin organization edit, mobile |

No `main` comparison: this branch adds migrations `main` doesn't have (plus the `passwordless_user_domain` → `user_email_domain` rename), so the schema is ahead of `main`'s code and the page errors there.

PR #3759, opened 2026-06-28
Add SAML SSO login entry UX (email-first + org-slug)

eld120:
Email first, then follow up step to email magic link / password / SSO

- [ ] If you use a password manager, it auto fills in the password from the first page
- [ ] There is an integration test for this

---

Surfaces the SAML SSO login path (PR3) to users. Two entry points, both falling through to the existing password / magic-link flow when SSO doesn't apply.

- **Email-first auto-detect** on `/session/new`: when the submitted email belongs to an org with live SAML SSO (`passwordless_email_matching` + a configured config), redirect to the IdP instead of asking for a Bike Index password.
- **Org-slug entry** (`GET /sso`): a "Single sign-on" link + form for users whose email domain isn't the org's passwordless domain — submit the org's short name to reach its IdP.
- Adds `Organization#saml_sso_configured?`, a lock icon, and the i18n strings for the new form/link.

Stacked on #3758 — base will retarget to `main` once that merges.


eld120:
## Screenshots

Captured signed-out (both pages are public). No `main` comparison: `/sso` is a new route, and the stacked SAML migrations leave the schema ahead of `main`'s code.

### /session/new

| Desktop | Mobile |
| --- | --- |
| session-new desktop | session-new mobile |

### /sso

| Desktop | Mobile |
| --- | --- |
| sso-entry desktop | sso-entry mobile |

eld120:
already implemented/superceded by PR  #3791

PR #3772, opened 2026-06-28
Make Lexxy editor toolbar configurable (toolbar_buttons:)

sethherr:
Make the Lexxy rich-text editor toolbar configurable per instance.

`Form::TextEditor::Component` now accepts `toolbar_buttons:` — the allowlist of buttons to show (e.g. `%i[bold italic link undo redo]`). Lexxy renders a fixed toolbar with no per-button config, so for each omitted button the component adds a `lexxy-editor--hide-<button>` class that scopes a `display: none` rule in `lexxy_overrides.css` to that editor; hidden commands stay reachable via hotkeys. Omitting the argument keeps the full toolbar. A `custom_toolbar` Lookbook preview and a system spec cover the shown/hidden buttons.

Also makes the `Spreadsheets::Components#to_csv` spec deterministic — it asserted the global `Ctype` table held exactly one row, so a leaked `Ctype.other` record from another spec failed it intermittently in CI; the export is now scoped to the ctype under test.

### DIFF
--- a/app/assets/tailwind/lexxy_overrides.css
+++ b/app/assets/tailwind/lexxy_overrides.css
@@ -25,19 +25,23 @@ lexxy-editor.lexxy-editor--compact .lexxy-editor__toolbar-button {
 /* Per-instance toolbar trimming (Form::TextEditor::Component toolbar_buttons:). The gem renders a
    fixed button set (lexxy.js LexicalToolbarElement.defaultTemplate) with no per-button config, so
    the component adds a lexxy-editor--hide-<button> class for each button the caller omitted; hidden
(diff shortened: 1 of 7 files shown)


sethherr:
## Screenshots

`toolbar_buttons:` trims the Lexxy toolbar per editor. Both states are on this branch — the feature is the difference between them (the `default` preview's toolbar is unchanged from `main`).

### custom_toolbar — `toolbar_buttons: %i[bold italic link undo redo]`

| Desktop | Mobile |
| --- | --- |
|  |  |

### default — full toolbar (no `toolbar_buttons:`)

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3744, opened 2026-06-25
Add organization registration sequences

sethherr:
Adds a versioned, org-editable "registration sequence" — an ordered set of pages (image, title/subtitle, and a rich-text body) an organization shows its registrants, with superuser approval before one goes live.

- **Status is derived, not stored.** A `RegistrationSequence` is draft / active / archived / template, computed from `organization_id` plus `start_at`/`end_at`. Partial unique indexes enforce one draft and one active per organization, and a single global template.
- **Org admins build a draft; a superuser approves it.** Org admins add, edit, drag-to-reorder, and remove pages on the draft under `Organized::RegistrationSequencesController`; `Admin::RegistrationSequencesController#update` calls `make_active!`, which archives the prior active so exactly one stays live.
- **New drafts clone the global template** (`RegistrationSequence.draft_for` / `.template`), seeded with default e-vehicle safety attestation pages in `db/seeds/seed_registration_sequence_template.rb`.
- **Pages** hold a title, subtitle, an Active Storage image, and a single rich-text `body` (HTML) authored with main's shared `Form::TextEditor` (Lexxy) component; body HTML is sanitized on save. Includes the Action Text tables migration the editor relies on.
- **Shared `RegistrationSequencePreview` component** renders the active and upcoming versions identically across the org and admin views.

PR #3784, opened 2026-06-30
Restore and rebuild the registration sequence preview

sethherr:
Restores the organization registration-sequence preview that was dropped from #3744 before it merged, and rebuilds it into a faithful walk-through of what registrants see — alongside editor improvements.

- **Preview** — a "Preview" button opens a paginated walk-through: each page rendered exactly as a registrant sees it and gated the same way, through to the review screen. The page editor shows that same preview inline and marks it stale until you save.
- **The flow and the preview render through the same components** — `Register::PageContent`, `ReviewSummary`, `AcknowledgmentCheck`, `RuleCheckbox` and `UI::DividedCard`, driven by the `register--acknowledgment` controller, so the two can't drift. Each takes only what actually varies: whether a checkbox submits, whether a review row links back.
- **Editing** — new drafts clone the live sequence (not the template), so an edit tweaks what's current; "Discard draft" starts over; pages require a title and at least one rule, added through a new/create flow so blanks aren't persisted; and the index button names what it does ("Create a sequence" / "Copy current sequence and edit" / "Edit sequence").
- The admin index page list starts expanded, and the FAQ-link setting and per-page image upload are wired up.


sethherr:
## Screenshots

### /o/:org/registration_sequences/:id — preview walk-through (new)

| Desktop | Mobile |
| --- | --- |
|  |  |

### /o/:org/registration_sequences/:id?page=4 — the review it ends on (new)

| Desktop | Mobile |
| --- | --- |
|  |  |

### /o/:org/registration_sequences/:id/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3805, opened 2026-07-06
ShortId: keep short ids decimal, add UI::ShortId component

sethherr:
Tightens `ShortId` encoding and adds a component for displaying short ids.

- Short ids whose base36 form is under 3 digits now stay decimal, e.g. bike 36 encodes to `r/36` instead of `r/10`. This fixes a silent round-trip collision, since `decode` treats any all-digit string as a decimal id (the old base36 `"10"` decoded back to bike 10). Large ids are unchanged (`3431156` → `r/21J-HW`).
- Adds `UI::ShortId::Component`, rendering a record's `short_id` as plain monospace text. Accepts a record responding to `short_id`, or a raw string.

### DIFF
--- a/app/components/ui/short_id/component.rb
+++ b/app/components/ui/short_id/component.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+module UI
+  module ShortId
+    # Renders a record's short_id (see ShortIdable) as a monospace code block.
+    # Pass a record that responds to short_id, or a raw short_id string.
+    class Component < ApplicationComponent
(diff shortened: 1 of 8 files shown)


sethherr:
## Screenshots

`UI::ShortId::Component` rendered in Lookbook (plain monospace, no background).

**Default** — base36 short id
UI::ShortId default

**Short decimal id** — under-3-digit ids stay decimal
UI::ShortId decimal

**With extra classes** — `html_class` appended
UI::ShortId with html_class

**Dark mode**
UI::ShortId dark mode

PR #3806, opened 2026-07-06
Fix mark-recovered leaving the bike stolen; add pending spinner

sethherr:
Fixes a bug where clicking "Mark recovered" could reload the page still showing the bike as stolen.

- The report_recovered modal fired an unconditional 500ms redirect right after kicking off the async recovery request. A request slower than 500ms was aborted by the navigation, so the page reloaded still-stolen. Now the redirect happens only once recovery succeeds (via `recoveredRequestCallback`), matching the sibling delete flow.
- Add a "Marking your bike recovered…" spinner, shown after the modal closes until the redirect, so there's feedback during the now-awaited request.
- Adds an integration spec that drives the full path through the UI — report a bike stolen, fill and verify the theft details, then mark it recovered — reproducing the race by delaying the recovery request.


sethherr:
## Screenshots

Pending state after submitting the recovery modal — shown until the redirect to the recovered bike.

### /bikes/:id/edit?edit_template=report_recovered

Marking your bike recovered spinner

PR #3820, opened 2026-07-08
Review-app/staging: banner, deploy fixes, CI + Kamal-config refactor

sethherr:
Review-app + staging improvements across the banner, the deploy steps, and the CI/Kamal plumbing.

**Banner**
- Shows **Staging** instead of **Review app** on the persistent staging deploy (no PR number), plus a "?" next to the label — styled like the email-outbox one — that **links to the deployed commit** on GitHub and shows `current commit: <sha>` on hover. Adds a `tooltip_link` slot to `UI::Tooltip` so a tooltip trigger can be a link.

**Deploy**
- Clears the Rails cache after each per-PR deploy (review apps only — staging untouched).
- Persistent `public/uploads` named volume across roles, so the worker sees the web-written original for version generation; removed on destroy (also fixes a pre-existing storage-volume leak). Added on staging too.
- Destroy cleanup (label + GHCR images) is gated on teardown success, so a transiently-failed destroy no longer strips them while the app/db/volumes stay leaked.

**CI & Kamal config**
- Staging deploys from CI's `dispatch` job (`gh workflow run staging.yml`) instead of a `workflow_run` trigger — note it runs in parallel with tests, so staging no longer waits for the suite to pass.
- Shared `kamal-deploy.yml` reusable workflow owns build + the kamal command; `review-app.yml` and `staging.yml` are thin callers.
- One Kamal config for both targets — `config/deploy.review.yml`'s ERB resolves the `pr-<N>` or `staging` slug and emits the shared-accessories block for review apps only; `config/deploy.staging.yml` deleted.
- `bin/kamal_review` passthrough defaults `--app` to `staging` (was the nonexistent `pr-0`); accessory commands now take an explicit `--app 0`.

Note: the workflow/config plumbing (`secrets: inherit` in the reusable workflow, the `staging` slug + `REVIEW_APP_REDIS_DB=0`) hasn't been exercised by a live run yet — worth one `workflow_dispatch` staging deploy before merge.


sethherr:
## Screenshots

Banner only renders under `REVIEW_APP`, so these are the ViewComponent previews. Desktop shots hover the commit "?" to show its tooltip.

### `page_block/review_app_banner/component/with_pr_link` — review-app banner (with PR)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The commit "?" (styled like the email-outbox one) links to the commit on GitHub; hover shows `current commit: …`. Both the label and commit "?" stay hidden on mobile.

### `page_block/review_app_banner/component/staging` — staging banner (new)

Persistent staging deploy: no PR number, so the label reads **Staging**. New preview, no `main` counterpart.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3827, opened 2026-07-09
Undo an export's bike sticker assignment

sethherr:
Adds an undo for an export's bike sticker assignment, replacing the old "Unassign stickers" action. Undo restores each sticker to the exact state it was in before the export claimed it — bike, claiming user, `claimed_at`, and secondary organization — by deleting the export's `BikeStickerUpdate` rows via the existing `RevertBikeStickerUpdateJob`.

- `PUT ?undo_bike_stickers=1` enqueues a revert per sticker and records `bike_codes_undone`. The button is `:error` colored and confirms first, since reverting erases the history of the assignment.
- A sticker claimed again since the export can't be reverted without erasing that later claim. Those are skipped, their codes stored in `bike_codes_not_undone`, and listed on the show page.
- The exports index badge now reports the state: `stickers` (cyan), `stickers unassigned` (orange), or `stickers restored` (gray).
- `BikeStickerUpdate#following_updates` is extracted from the job, which is the check that decides whether a sticker is revertable.

The "Unassign stickers" button and its `?remove_bike_stickers` param are removed — it unclaimed stickers outright rather than restoring them. `Export#remove_bike_stickers` stays, since export deletion and the export job's ebrake still call it, and `bike_codes_removed` still renders for exports that were unassigned before this change.


sethherr:
## Screenshots

### Export show — assignment in effect

The undo button is the only sticker action. Clicking it raises a native `confirm()`:

> Are you sure you want to undo the sticker assignment? This will remove the history of having assigned these stickers. This can not be undone.

Native browser dialogs are chrome rather than page content, so Playwright can't capture one — the wording above is the literal string from `config/locales/en.yml`, verified firing in the browser (Cancel is a no-op; Accept submits).

| Desktop | Mobile |
| --- | --- |
| show-undo-desktop | show-undo-mobile |

### Export show — after undo

The restored message leads. `BR1` was claimed again after the export, so reverting it would erase that later claim — it's skipped and named. The sticker list collapses behind a disclosure toggle.

| Desktop | Mobile |
| --- | --- |
| show-undone-desktop | show-undone-mobile |

With "Show previously assigned stickers" expanded:

| Desktop | Mobile |
| --- | --- |
| show-undone-expanded-desktop | — |

### Exports index — sticker state badges

`stickers` (cyan), `stickers unassigned` (orange), `stickers restored` (warning/amber).

| Desktop | Mobile |
| --- | --- |
| index-desktop | index-mobile |

PR #3839, opened 2026-07-09
Purge a review app's R2 objects on destroy

sethherr:
Review-app ActiveStorage attachments were written to the shared `cloudflare_dev` R2 bucket and never cleaned up, so every destroyed PR leaked its blobs.

- `bin/kamal_review destroy` now purges the PR's R2 objects through the running app before `app remove` — enumerating only this PR's own blobs (keys are globally unique, so no other app's objects are touched), best-effort like the rest of destroy.
- Corrected `docs/review-apps.md`, which claimed a per-PR `review-app/` prefix that was never implemented; documents the actual split (CarrierWave → local volume, ActiveStorage → shared R2 root) and the new purge.

## Rollout

No action is needed for the mechanism itself: the destroy workflow checks out the **default branch**, so once merged, every review-app teardown — including closing PRs opened before this — runs the new purge automatically. Currently-open review apps are cleaned when they close (as long as the app still boots; a dead app orphans its blobs at destroy, same as today).

The one gap is the **existing backlog** — blobs from PRs destroyed *before* this change. Their databases are already dropped, so nothing can enumerate them, and with no per-PR prefix the shared bucket mixes them with staging's and dev's live blobs at the root. Clearing them is a one-time manual reconciliation: build a keep-set of `ActiveStorage::Blob.pluck(:key)` from staging plus every open review app, list objects in `bikeindex-dev`, and delete those whose key isn't in the set. Optional (dev/sandbox bucket, no PII) — worth doing only if the bucket has grown large.

## Impact

- Only **review-app destroys** change behavior. Staging can't reach this path (`destroy` requires a PR number and rejects `staging`), so staging's blobs are never touched.
- The purge reads only the PR's **own** database and deletes by unique key, so it removes exactly that PR's objects and nothing else's.
- Production R2 (the `bikeindex` bucket, Cloud66) is entirely unaffected — different bucket, different deploy path.
- CarrierWave uploads (the local `_uploads` volume) are unaffected — already dropped with the volume.
- Destroy runs slightly longer (boots `bin/rails runner`, one delete per blob) and is best-effort (`|| true`), so a purge failure never blocks teardown.


sethherr:
## Considered alternative: per-PR R2 prefixes (not pursued)

Documenting this for the future. This PR cleans up review-app R2 blobs by enumerating each PR's own database and deleting by key. An alternative is to give each app a real folder prefix (`review-app/pr-<N>/`) in the bucket. We chose **not** to do this — recording what it would take, and why it wasn't worth it.

### What it would take

**1. A prefix-aware S3 service** — Rails 8.1's `ActiveStorage::Service::S3Service` has no native prefix/root option. Subclass it, accept a `prefix:` config, and override the private `object_for(key)` to prepend it (that one funnel handles all reads/writes/deletes/`exist?`/`url`), plus `delete_prefixed` so variant cleanup stays scoped. Registration is fiddly: ActiveStorage's `Configurator` does `require "active_storage/service/prefixed_s3_service"` then `const_get`, so the file must live at exactly `lib/active_storage/service/prefixed_s3_service.rb` (`lib` is on `$LOAD_PATH`), and `active_storage` must be added to `config.autoload_lib(ignore:)` so Zeitwerk doesn't fight the explicit require over the gem's existing `ActiveStorage::Service` namespace.

**2. `config/storage.yml`** — add a service (e.g. `cloudflare_dev_prefixed`) using the new class with `prefix: <%= ENV["ACTIVE_STORAGE_PREFIX"] %>`. Leave the existing `cloudflare_dev` untouched so local dev keeps using it unprefixed.

**3. `config/environments/staging.rb`** — point `config.active_storage.service` at the prefixed service.

**4. `config/deploy.review.yml`** — set `ACTIVE_STORAGE_PREFIX` from the existing `slug`: `review-app/pr-<N>/` for PRs, `review-app/staging/` for staging.

**5. `app/services/blob_url.rb`** — the real coupling. It builds `File.join(storage_host, blob.key)` from the **bare** key (the DB stores keys unprefixed; the service adds the prefix only to the physical object), so BlobUrl must prepend the same prefix — but only for the `cloudflare_dev`-family host, not production. The service and BlobUrl then both depend on the same prefix ENV. (`BlobUrl.for` is the single funnel for review-app blob URLs — every caller goes through it, nothing else builds a URL from `blob.key` — so this is the only view-layer change.)

**6. Destroy** — becomes simpler and more robust: one `delete_prefixed("review-app/pr-<N>/")` instead of enumerating the DB. Works even when the app can't boot, and makes the backlog identifiable/wipeable by prefix.

**7. Tests** — the service (prefixing + `delete_prefixed`) and BlobUrl-with-prefix.

### The two hard parts

- **Custom-service registration** (1) — workable but non-obvious; reopening the gem namespace from `lib` needs the Zeitwerk-ignore dance.
- **Cutover / existing objects** — the sharp edge. The moment BlobUrl starts prefixing, *every* blob already at the bucket root 404s: its URL now points at `prefix/key` but the object is at `key`. Review apps are ephemeral (redeploy → reseed), but **staging is persistent** — its existing blobs would need a one-time copy (root → `review-app/staging/`) or a reseed. Acceptable on a no-PII sandbox bucket, but it's a real migration step, not a pure code change.

### Why not now

Over the DB-purge in this PR, prefixes buy exactly two things: cleanup that survives an app that won't boot at destroy, and identifiable orphans for the backlog sweep. Steady-state cleanup is already solved. The cost is a custom service, a 5-file change touching production-shared code (`storage.yml`, `BlobUrl`), and a staging cutover — not worth it unless failed-boot orphans become a recurring annoyance or a prefix-wipeable backlog becomes desirable.

PR #3850, opened 2026-07-13
Add redesigned registration flow at /register

sethherr:
Adds a redesigned multi-step public registration flow at `/register`, separate from `/bikes/new` and built from the `kelsey_redesign/register-your-bike.html` prototype (committed here). Builds on #3980, merged in, which carries the shared `UI::Forms::Combobox` and `form-persist` changes.

- `register/new` creates a blank `BParam` (reusing the session's while step 1 is unsubmitted) and redirects into `show`, which renders the whole flow at `/register?b_param_token=…&step=1|2|finished` — clamping the requested step to the registration's progress, so the browser back button returns to step 1 prefilled. The logic lives in `BikeServices::Register`; the controller keeps params, session and redirects. Copy interpolates the cycle type ("Register your e-scooter!") and follows the vehicle type as it changes.
- **Completion forks on creator availability** (`Ownership` requires one): signed-in registrants, and organizations with an auto user, get the bike at step 2. Anyone else lands on `step=finished` awaiting email confirmation — but **that email is no longer sent**, so an anonymous registration with no organization currently has no way to finish, and the copy on steps 1/2 and the finished page still describes the link. `confirm_email` and the `confirmation_token` are intact, and an organization can still resend from its incomplete registrations.
- Entry params on `register/new`: `organization_id` (slug or id, also honored on later steps and posted back from step 1), `status`, `email` (`false` blanks it even when signed in), and `b_param_token=false` to start a fresh registration — what "Start over" links to.
- Step 2 asks for what `bikes/new` asks: phone, address, extra registration number, affiliation and student id, each behind the same `include_field_reg_*` gate and grouped under the organization's own heading. Status is picked in this form rather than known before render, so `register--status-fields` rechecks those gates client-side and disables the fields a status doesn't ask for. Photo upload is offered once there's an account behind it.
- Drafts survive a reload through `form-persist`, keyed per registration; it learns checkboxes, radios, selects and combobox pairs, and announces restores with a `form-persist:restored` window event so dependent UI reconciles. `CleanBParamsJob` moves to 3 days and sweeps never-submitted blank shells.

Intentionally not in this pass: the prototype's organization select UI, member-contact toggle, and model catalog dropdown (model stays free text).


sethherr:
## Screenshots

All three pages are new on this branch, so there's no base to compare against.

### /register?step=1

| Desktop | Mobile |
| --- | --- |
| register step 1 desktop | register step 1 mobile |

### /register?step=2

| Desktop | Mobile |
| --- | --- |
| register step 2 desktop | register step 2 mobile |

### /register?step=finished

| Desktop | Mobile |
| --- | --- |
| register finished desktop | register finished mobile |

PR #3863, opened 2026-07-15
Migrate twbtn buttons to UI::Button and consolidate button colors

sethherr:
Migrates the legacy `twbtn`/`twbtn-blue`/`twbtn-green` CSS button classes to the `UI::Button`/`UI::ButtonLink` components and consolidates the button color palette.

- **twbtn → components** — all 7 call sites (membership/sale/message forms, both search submit buttons, the marketplace "sold to" link, and the search kind-option link) now render `UI::Button`/`UI::ButtonLink`; the unused `.twbtn*` CSS and the legacy Lookbook preview are deleted.
- **Palette** — folds `:outline` into `:secondary` (white base, neutral border) and drops `:outline`; adds a `link_purple` color plus a `LINK_COLORS` list that generalizes the text-only (link) colors.
- **`.twlink`** — `:link` now underlines on hover/active instead of always and shares the `.twlink` class, which is updated to match. Note this makes every `.twlink` link app-wide underline on hover rather than always.


sethherr:
## Screenshots

Branch (after) captures. A before/after against `main` wasn't captured — the local dev Tailwind watcher wouldn't recompile `main`'s `.twbtn*` CSS without a `bin/dev` restart, so the "before" shots would have rendered unstyled.

### /

| Desktop | Mobile |
| --- | --- |
|  |  |

### /search/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3865, opened 2026-07-15
Eager-build CarrierWave versions to fix avatar NoMethodError

sethherr:
Fixes Honeybadger fault 132710180 (`NoMethodError: undefined method '[]' for nil`) on the user profile page. Rendering `@user.avatar` makes CarrierWave build its version uploader subclasses (thumb/medium/large/biggest) lazily on first access, and that build is not thread-safe — under Puma a second thread can read the half-built class before `version_options` is assigned, so `versions[name].class.version_options[:if]` raises in `version_active?`.

- Add `CarrierWaveVersionWarmer.warm!`, which builds every version subclass once at boot (`after_initialize`, single-threaded) so the race can't happen at request time.
- Cover the warmer with a spec asserting each version subclass ends up with `version_options` populated.

### DIFF
--- a/config/initializers/carrierwave.rb
+++ b/config/initializers/carrierwave.rb
@@ -59,3 +59,7 @@ end
 CarrierWave::Backgrounder.configure do |c|
   c.backend :sidekiq, queue: :med_priority, retry: 2
 end
+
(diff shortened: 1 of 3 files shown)


sethherr:
closing until it happens more frequently.

PR #3869, opened 2026-07-16
Default UI::Tooltip to the "?" button trigger

sethherr:
`UI::Tooltip::Component` now renders the "?" info button by default, so a bare `render(UI::Tooltip::Component.new(text:))` produces a styled trigger with no block or slot required.

- Extracted the button styling into a `BUTTON_CLASS` constant and made it the default for both the `tooltip_button` slot and the no-block trigger.
- Callers/previews that hand-rolled the "?" button class no longer need to.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The `?` default trigger renders smaller (`h-4 w-4`, `text-2xs`) than main's `with_tooltip_button` slot examples (`h-5 w-5`, `text-xs`).

PR #3884, opened 2026-07-20
Fix two frontend Honeybadger errors: WebGL init + cross-page turbo-frame

sethherr:
Fixes two frontend errors from Honeybadger.

- **`stolen_map` / WebGL** — WebGL is often unavailable (crawlers, headless browsers, disabled GPUs, blocked CDN), so Mapbox init rejected and was reported as an unhandled promise rejection. The controller now catches it and, instead of leaving a blank canvas, reveals a server-rendered (translatable) message explaining the map couldn't load.
- **`search/form_controller`** — `reloadFrameIfUrlStale` compared only the query string, so a back/forward restoration could point one page's results frame (e.g. `marketplace_results_frame`) at another page's URL (`/search/registrations`), whose response lacks that frame and gets discarded by Turbo. Guard on `pathname` first.

The third error from the same batch (invalid combobox selector) was already resolved on `main` by the general `hw_combobox_patch.js` override in #3860, so it needs no further change.

### DIFF
--- a/app/assets/stylesheets/revised/pages/bikes/show.scss
+++ b/app/assets/stylesheets/revised/pages/bikes/show.scss
@@ -124,6 +124,14 @@
     height: 400px;
   }
 
+  .map-unavailable {
+    min-height: 400px;
+    padding: 2rem 1rem;
+    text-align: center;
+    color: #6b7280;
+    background: #f3f4f6;
+  }
+
   .bike-photos,
   .show-bike-details,
   .ad300x600,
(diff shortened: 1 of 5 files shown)


sethherr:
## Screenshots

### `/bikes/:id` — stolen map WebGL-failure fallback

Normal map rendering is unchanged; this is the new message shown **in place of the map** when WebGL/Mapbox can't initialize (crawlers, headless browsers, disabled GPU, blocked CDN). Before this PR the same failure left a blank box plus an unhandled promise rejection. Captured on a seeded stolen bike with the controller driven into its failure state.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3885, opened 2026-07-20
Use UI::Forms components for the register flow fields

sethherr:
Build the `/register` flow's form fields from the shared `UI::Forms` component library instead of hand-rolled markup, so the registration redesign stays in lockstep with the rest of the app. Stacked on #3850.

- Both register steps render through `UI::Forms::Group` / `Input` / `Checkbox` / `Combobox` / `RadioButtonGroup` rather than bare `f.*` helpers.
- Colors use a combobox with a `UI::ColorSwatch` per option, and the selected swatch shows in the field.
- Frame size uses the full-width `RadioButtonGroup` chips; the motorized + serial toggles use `UI::Forms::Checkbox`.


sethherr:
## Screenshots

Branch-only — the base is `sethherr/new-registration-flow` (not `main`), so the tooling can't auto-capture a before/after. Step 2 is shown with a populated draft (Black + Blue colors, size M).

### /register/new

| Desktop | Mobile |
| --- | --- |
|  |  |

### /register/details

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3891, opened 2026-07-20
Move production deploy to its own main-only workflow

sethherr:
Follow-up to #3890. Moves the Cloud 66 production deploy out of `ci.yml` into its own `deploy-production.yml` so it no longer leaves a skipped "Deploy to production" check on every PR and branch.

- Triggered by CI's `workflow_run` completion on `main`, gated on `conclusion == 'success'` — same "only deploy after CI passes" behavior, but now the whole CI run must be green rather than just `test` + `lint_and_scan`.
- The check now attaches only to `main` commits, so PRs stay clean.


sethherr:
Current tradeoffs are better than this

PR #3898, opened 2026-07-22
Polish the registrations show redesign

sethherr:
UX polish and deduplication for the registrations show redesign, stacked on `sethherr/bike-show-page-flag`.

- **Share button** renders via `UI::Button::Component` (the old markup passed a non-existent `:outline` color and rendered unstyled).
- **View switcher**: a divider separates each organization, and the chip always shows the org name (the warning color still flags the not-registered case) instead of "Not in your org".
- **Public view**: the serial reads "hidden" for a hidden serial regardless of viewer, and the redundant registration-date header is gone from the organized view.
- **Deduped the consumer and org-admin components** into shared sub-components — `BikeIndexCard` and `BikeDetails` for the two summary cards, plus `PrimaryColors`, `Subtitle`, and `Status`. Each view now holds only its genuinely view-specific bits.


sethherr:
## Screenshots

Registrations show redesign (Wrapper Lookbook preview). Branch-only: the preview URLs can't be captured on the base branch, so there's no side-by-side before/after here.

### /rails/view_components/registrations/show/wrapper/component/default

Public view — the **Share** button renders as a proper `:secondary` `UI::Button` with a border and background (previously it passed a non-existent `:outline` color and rendered borderless/transparent).

| Desktop | Mobile |
| --- | --- |
| share button desktop | share button mobile |

### /rails/view_components/registrations/show/wrapper/component/default?view_as=organization

Organized view — the switcher chip shows the org name and role (**Brakebills · Full access**), the registration-date header is gone, a long primary-color name wraps on mobile, and the "This bike on Bike Index" card (now shared with the consumer view) no longer shows a credibility row.

| Desktop | Mobile |
| --- | --- |
| organized view desktop | organized view mobile |

PR #3904, opened 2026-07-22
Add Org::OriginDisplay component with an origin-title tooltip

sethherr:
Extracts the `origin_display` helper into a new `Org::OriginDisplay::Component` and routes every call site through it, so the bike creation origin — label plus its explanatory title — lives in one place.

- Shows the origin title in a `UI::Tooltip` (default `?` button) beside the origin label, instead of a native `title=` attribute.
- Moves the origin-title text mapping out of `BikeServices::Displayer` (the component was its only caller) and removes the old `origin_display` helper.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -52,7 +52,7 @@ The same instinct applies beyond buttons: **check `app/components/ui/` and `app/
 
 ## Tooltips: default `?` button trigger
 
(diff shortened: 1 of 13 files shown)


sethherr:
## Screenshots

The origin (bike creation source) now renders through `Org::OriginDisplay::Component`: the origin label plus a `UI::Tooltip` default `?` button that reveals the origin title on hover/focus. The `main`/before side uses the native `title=` attribute, which doesn't appear in a screenshot.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 (tooltip shown) |
|  |  |

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3909, opened 2026-07-22
Fix creation_description for the unregistered parking notification origin

sethherr:
Follow-up to #3908, which added the "Unregistered Parking Notification"
label but keyed the component off the incidental humanized origin string.

- Fixes the dead branch in `Ownership#creation_description` (`ownership.rb:192`):
  it checked the non-existent origin `unregistered_parking_notification`
  instead of the real enum key `creator_unregistered_parking_notification`,
  so it never fired and the value fell through to the titleized
  "creator unregistered parking notification". It now returns "parking notification".
- Simplifies `Org::OriginDisplay::Component` to key its label and tooltip
  off that stable string (tooltip moved into `EXTENDED_DESCRIPTIONS`).
- Fixes the same never-matching check on the bike show page
  (`bikes/show.html.haml`), which gated the "created to record parking
  notification" note; it now uses the `creator_unregistered_parking_notification?`
  enum predicate so the note renders.

Note: because `creation_description` is consumed raw elsewhere, the org export's
`registration_method` column and the admin ownerships table now read
"parking notification" instead of "creator unregistered parking notification".


sethherr:
## Screenshots

### /bikes/:id  (unregistered parking-notification bike)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The "Not registered to a user" warning and the "created by … to record a Parking Notification" note were previously gated behind a never-matching origin string, so they never rendered for parking-notification bikes. They now appear.

PR #3922, opened 2026-07-23
Add no_redesign override and legacy viewer link

sethherr:
Adds an escape hatch out of the `bike_show_redesign` rollout so users can reach the classic bike page, and reworks the cross-linking between the two views.

- `/bikes/:id?no_redesign=true` renders the classic show page instead of redirecting to the redesign, even when the flag is on.
- The redesign's view_as chooser gains a "View in Legacy Viewer" entry (with a divider above it) when `bike_show_redesign_toggle` is enabled, linking to that override.
- A redesign-default user who lands on the legacy page now sees a tailored invitation ("Legacy viewer…") offering to make legacy their default or jump back to the new viewer, and the opt-in/opt-out copy names the cycle type.


sethherr:
## Screenshots

### /registrations/:id — view_as chooser open

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/:id?no_redesign=true — legacy viewer alert

New state added by this PR (on `main` this URL redirects to the redesign), so no before/after.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3923, opened 2026-07-23
Show the review-app top bar on the local dev server

sethherr:
Renders the review-app top bar on the local dev server too, so a dev environment is never mistaken for production. It shows in purple (matching the dev favicon) — distinct from the review-app green — and carries the email-outbox link and one-click superadmin sign-in.

- The layout passes `"development"` as the banner's `review_app` value when `Rails.env.development?`, and never renders the banner in production at all.
- Purple `Development` accent with no "data is ephemeral" disclaimer (the local DB persists); review-app and sandbox deploys keep the green styling and the disclaimer.
- Set `NO_REVIEW_TOPBAR=true` to suppress the bar on every non-production deploy (dev, review apps, sandbox) — the opt-out is a constant read in the component's `render?`.


sethherr:
## Screenshots

### /

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3930, opened 2026-07-23
Extract parking-notification details component; org-admin polish

sethherr:
Extracts the parking-notification show details into a reusable component and polishes the org-admin registration view.

- New `Org::ParkingNotificationDetails` component renders the details as two `UI::DefinitionList` (`term: :below`) columns inside a `UI::Card`, replacing the old `table-list` markup in `_show_details`. The org-admin "View notifications" panel now reuses it (with `skip_bike:`) instead of its own hand-rolled definition list.
- The org-admin "other registrations" card gains the column-toggle settings, so it shows the same default column set as the registrations search (sharing one settings instance with the table).
- The org-search settings toggle is now a primary `UI::Button`, with its pressed state driven by `aria-pressed`.


sethherr:
## Screenshots

### `/o/brakebills/parking_notifications/:id`

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### `/registrations/:id` (View notifications panel open)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3932, opened 2026-07-23
Move IpSpoofAttackFilter to config/initializers

sethherr:
Move `IpSpoofAttackFilter` out of `lib/` and into a self-contained `config/initializers/ip_spoof_attack_filter.rb` that both defines the class and inserts it into the middleware stack.

- Collapses the previous three coordinated pieces (class in `lib/`, `require_relative` in `application.rb`, and the `insert_after` call inside the config block) into one file.
- No behavior change — the class body and the `insert_after ActionDispatch::DebugExceptions` placement are unchanged.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -16,8 +16,6 @@ require "action_view/railtie"
 require "sprockets/railtie"
 # require "rails/test_unit/railtie"
 
-require_relative "../lib/ip_spoof_attack_filter"
-
 # Require the gems listed in Gemfile, including any gems
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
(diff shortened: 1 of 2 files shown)


sethherr:
Closing this. The goal was to move `IpSpoofAttackFilter` out of `lib/` and into `config/initializers/`, but the interesting constraint that came out of it is worth recording:

**The middleware `insert_after` can't live in `config/application.rb` if the class is defined in an initializer.** Rails 8 (`config.load_defaults 8.0`) removed string/symbol middleware arguments, so `config.middleware.insert_after` now requires the actual class **constant**, not `"IpSpoofAttackFilter"`. But `application.rb`'s config block is evaluated *before* `config/initializers/*` load, so the constant doesn't exist yet at that point — passing a string just crashes boot with `undefined method 'new' for an instance of String`.

That leaves two viable layouts:
- **Current (`main`):** class defined in `lib/`, `require_relative`'d at the top of `application.rb`, inserted from the config block with the bare constant.
- **This PR:** class *and* insert both in one `config/initializers/ip_spoof_attack_filter.rb`, using `Rails.application.config.middleware.insert_after` (the app instance exists by initializer time).

There's no clean way to split it the third way (class in initializer, insert in `application.rb`) without reintroducing a `require_relative`, which is uglier than either option above. Since this is a lateral organizational move with no behavior change and no clear win, closing rather than merging.

PR #3935, opened 2026-07-24
Hide review-app topbar in screenshots

sethherr:
Screenshots captured via the `frontend-screenshots` skill were including the dev/review-app topbar banner, which isn't part of the real page.

- Add a stable `#review-app-banner` id to the banner component so it can be targeted reliably.
- Hide `#review-app-banner` (alongside the footer) in the skill's per-navigation capture step, so captures start at the actual page content.


sethherr:
## Screenshots

Homepage captured with the skill's hide step toggled — the review-app topbar is present in the old behavior and gone with this change. (Not a page-code change, so there's no base-branch difference to compare; this shows the tooling's effect.)

### /

| Desktop | Mobile |
| --- | --- |
|  |  |
| topbar shown 👆 | topbar hidden 👇 |
|  |  |

PR #3939, opened 2026-07-24
Use UI::Button purple_outline palette for RadioButtonGroup

sethherr:
Restyles the `UI::Forms::RadioButtonGroup` segmented control to `UI::Button`'s `purple_outline` color, makes the brand purple the theme's `purple` scale, and reworks `UI::Button`'s active state so it stops fighting the cascade.

- **RadioButtonGroup**: resting/hover use the purple_outline gray-on-white look with a purple accent; the selected segment fills solid purple (white text) via `has-[:checked]:`, focus ring in the same purple. A spec asserts its purple utilities match the ones `UI::Button` emits, so the two can't drift.
- **`purple` theme scale**: the OG Bike Index purple (`#715eb2` + shades) now defines Tailwind's `purple` (`--color-purple-50…950`), replacing the default violet so `bg-purple-*` etc. resolve to the brand hue app-wide. `UI::Button` and the radio group use these utilities instead of arbitrary hex values.
- **UI::Button active state**: `active: true` renders `data-active="true"` rather than swapping class sets, and a new `is-active` custom variant (`data-active`, `aria-pressed`, `aria-current`, `:active`) carries the active colors. It's declared last so it outranks the resting, hover and dark colors with no `!important` — a plain `data-[active=true]:` variant isn't enough, since `dark:hover:*` sorts after it and wins. That collapses two parallel color maps into one delta over `COLORS` and retires `UI::Dropdown`'s hand-doubled mirror.
  - Mousedown on `primary`/`error` now darkens one step less: the shadowed `active:bg-*` shades leave `COLORS`, replacing a tie that CSS was resolving arbitrarily (the darker shade won for blue/red, the active shade for gray).

### DIFF
--- a/app/assets/tailwind/application.css
+++ b/app/assets/tailwind/application.css
@@ -17,6 +17,12 @@
  * ancestor activates tw:dark:* utilities. */
 @custom-variant dark (&:where(.dark, .dark *));
 
+/* The selected/pressed state of a button, link or menu item, however it's flagged.
+ * Keep it the last @custom-variant: tailwind sorts custom variants in declaration
+ * order, and sorting after `dark` is what lets tw:is-active:* beat the resting,
+ * hover and dark colors it overrides without `!important`. */
+@custom-variant is-active (&:is([data-active="true"], [aria-pressed="true"], [aria-current], :active));
+
 @theme {
   --font-sans: "Open Sans", sans-serif;
   --font-header: "Montserrat", sans-serif;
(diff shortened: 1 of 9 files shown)


sethherr:
## Screenshots

### /lookbook/preview/ui/forms/radio_button_group/examples

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3940, opened 2026-07-24
Add a map location picker to the org parking-notification form

sethherr:
Renames the org parking-notification form's "Use my current location" option to "Set on map" and reveals an interactive map above the message field, where the admin positions a fixed centre pin by moving the map under it — easier on a phone than dragging a marker your finger covers.

- The pin seeds from the browser's location, falling back to the organization's; moving the map (or tapping the geolocate button) sets the coordinates. A blue dot marks where the device actually is, so a moved pin can be judged against it. The pin *is* the readout — there's no coordinate or address text, and nothing reverse-geocodes while you pan.
- Pin and zoom are mirrored into the URL, so a reload or shared link restores the view, and a stored pin wins over device geolocation. The org fallback is deliberately excluded: it's somewhere to point the map, not a spot anyone picked, so it neither reaches the URL nor seeds the address fields.
- Switching to "Enter address manually" reverse-geocodes the pin into the address fields. A hand-edited address survives toggling modes, but moving the pin replaces it — otherwise the server geocodes the stale address back over the coordinates on submit.
- Both maps get an expand control: it covers the page on desktop so the site chrome stays reachable, and uses real fullscreen below `48rem`. Scroll zooms.
- Tiles come from the self-hosted MapLibre + Protomaps basemap from #3929 rather than Mapbox, so this map costs nothing per load; the ground-radius helper and the new expand control are shared with the registration map via `utils/maplibre`. Mapbox remains only for reverse-geocoding, and the stolen-bike map is untouched.

`Wrapper`'s cache key is bumped to `registrations/show-v2`. It's a `def call` component, so `cache` digests nothing, and `ActionView::Digestor` doesn't follow `render(SomeComponent.new)` into nested components either — cached fragments would otherwise keep serving the old markup.


sethherr:
## Screenshots

### /rails/view_components/registrations/show/org_top_actions/parking_notification_form/component/default?panel=parking

Captured with a mocked device fix, after choosing "First notice" (the preview bike has an earlier notification, so the panel opens on "Repeat").

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /registrations/43

The stolen-location map, which gains the expand control.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3943, opened 2026-07-24
Keep the stolen/impounded map circle over a fixed ground area

sethherr:
The translucent circle on the stolen/impounded map covered *more* ground the further you zoomed out — implying less location precision than the data actually has. `circle-radius` is in screen pixels, and the old curve (`stops: [[5, 5], [16, 240]]`, base 1.15) grew it 48× across 11 zoom levels while the map scale changed 2048×.

- The pixel radius is now derived from a real ground radius and interpolated base-2, so it doubles every zoom level and pins to a fixed footprint. The opaque `radiusBase` curve constant becomes `radiusMeters` (1000m when coordinates are rounded to `Bike::PUBLIC_COORD_LENGTH`, 250m when the exact address is public).
- Radii are picked to match the old on-screen size at the default zoom 13, so the initial view is unchanged — only zooming behaves differently.
- Extracts the MapLibre + PMTiles CDN loader into `utils/maplibre`, byte-identical to the copy on `sethherr/parking-notification-map-pin` so the two converge on merge.

Scoped to the `bike_show_redesign` component. The legacy Mapbox path (`stolen_map_controller.js:67`, still what unflagged users see) carries the same inverted scaling and is left for a follow-up.

### DIFF
--- a/app/components/registrations/show/map/component.html.erb
+++ b/app/components/registrations/show/map/component.html.erb
@@ -3,7 +3,7 @@
   data-registrations--show--map-style-url-value="<%= MAPS_STYLE_URL %>"
   data-registrations--show--map-latitude-value="<%= @latitude %>"
   data-registrations--show--map-longitude-value="<%= @longitude %>"
-  data-registrations--show--map-radius-base-value="<%= radius_base %>"
+  data-registrations--show--map-radius-meters-value="<%= radius_meters %>"
   data-registrations--show--map-point-value="<%= @point %>"
   class="
     tw:relative tw:mt-3 tw:mb-4 tw:h-80 tw:w-full tw:overflow-hidden
--- a/app/components/registrations/show/map/component.rb
+++ b/app/components/registrations/show/map/component.rb
@@ -20,9 +20,10 @@ module Registrations
 
         private
 
-        # The circle grows faster with zoom when the exact address is public
-        def radius_base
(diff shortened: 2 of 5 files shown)


sethherr:
## Screenshots

`/registrations/43` — a seeded stolen bike with non-precise coordinates (Oakland, rounded to `Bike::PUBLIC_COORD_LENGTH`).

### /registrations/43 — default zoom (13)

Unchanged, by design: the radii are picked so the initial view matches what shipped before.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /registrations/43 — zoomed out 3 levels (zoom 10)

Where the bug shows. On `main` the circle sprawls from Emeryville across Oakland; on this branch it holds its ~1km footprint.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Measured against the map's own projection, the rendered circle now covers a constant 999 m at every zoom (pixel radius doubling exactly per level):

| zoom | 8 | 10 | 13 | 15 | 16 |
| --- | --- | --- | --- | --- | --- |
| pixel radius | 4.1 | 16.6 | 132.5 | 530 | 1059.9 |
| ground radius | 999 m | 999 m | 999 m | 999 m | 999 m |

PR #3950, opened 2026-07-25
Fix the radio group's checked ring and hovered right border

sethherr:
`UI::Forms::RadioButtonGroup` restates `UI::Button`'s `purple_outline` palette under a `has-[:checked]:` variant, since its active state is driven by a checked radio rather than the `is-active` data attribute. It restated everything except the ring, so the checked segment read flatter than an active `purple_outline` button.

Fixing that surfaced a second problem: segments overlap by `-ml-px`, and every label was statically positioned, so a later sibling painted over its neighbor's right border and ring. That's why a hovered segment's right border stayed gray, and it clipped the focus ring the same way.

- The checked label now carries `ring-2 ring-purple-500/40`, matching `ACTIVE_COLORS[:purple_outline]`.
- Every label is `relative`, with a z-index ladder deciding who owns the shared pixel: hover/focus over checked over resting. Hover wins over checked deliberately — the state under the pointer should be the one drawn whole.
- The `with_html_labels` preview gets its own radio name; Lookbook's Examples page renders every scenario into one DOM, so sharing `search_status` made it and `default` a single radio group.
- The lockstep spec compares the label's `has-[:checked]:` utilities against `ACTIVE_COLORS[:purple_outline]`, so a color added to the button has to reach the group or the spec fails.

The stacking fix isn't covered by a spec — it's paint order, which only a pixel diff would catch. See the screenshots comment.

### DIFF
--- a/app/components/ui/forms/radio_button_group/component.rb
+++ b/app/components/ui/forms/radio_button_group/component.rb
@@ -43,6 +43,8 @@ module UI
             "tw:hover:border-purple-500 tw:hover:bg-purple-50 tw:dark:hover:border-purple-500 tw:dark:hover:bg-purple-950",
             "tw:has-[:checked]:bg-purple-500 tw:has-[:checked]:text-white tw:has-[:checked]:border-purple-500",
             "tw:has-[:checked]:hover:bg-purple-500 tw:has-[:checked]:hover:border-purple-500",
+            # relative/z-10 keeps the ring above the neighboring labels, which overlap it by a pixel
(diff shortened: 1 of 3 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/forms/radio_button_group/component/with_selection

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### Hovering the middle segment

Desktop only, at 4× zoom — hover isn't a touch state, and the difference is a single pixel of border. On `main` the hovered segment's right border stays gray, because the next segment overlaps it and paints over it.

| Desktop |
| --- |
|  |
| main 👆 | 
|  |

PR #3953, opened 2026-07-25
Center the OrgTopActions button icons and text on mobile

sethherr:
The org top-action buttons center their icon and text on mobile. From `sm` up they render exactly as before — flush left.

- Below `sm` the button's flex wraps instead of laying out as a grid, so the icon and its subtitle sit on one centered line with a full-width title beneath. The grid couldn't do this: the title spanning both tracks widened them, leaving the icon off-center.
- The subtitle stays left-aligned, so the notification counts line up with each other rather than centering line-by-line inside the block.


sethherr:
## Screenshots

### /lookbook/preview/registrations/show/org_top_actions/wrapper/default

Desktop is unchanged — the two desktop shots are pixel-identical apart from Lookbook's render-time badge.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3955, opened 2026-07-26
Shared UI component updates from registration flow branch

sethherr:
#3850 grew a pile of shared-component work alongside the register flow itself. This is the component half, so it can land and be reviewed without the flow — `UI::Forms::Checkbox` and `UI::Forms::Select` are new, and `Group`, `Input`, `RadioButtonGroup`, `Header`, `Button` and `Badge` each gain a variant or argument the redesign needed.

- `UI::Forms::Group` derives its label suffix from `required:` — a red `*`, or an "optional" badge — and threads `required:` through to the field. `Checkbox` now takes a `form_builder` + `attribute` as well as a bare `name`, which covers the model-scoped case `Group`'s `check_box` branch used to, so that branch is gone.
- `select` moves out of `Input` into its own `UI::Forms::Select`, labelled through `kind: :content_block`. Rails' `select` takes its option tags and options positionally, so it never fit `Input`'s `send(kind, attribute, html_options)` contract — it needed a branch plus two constructor kwargs dead for every other kind.
- `UI::Forms::RadioButtonGroup` renders as chips throughout, dropping the segmented-control pills and their z-index ladder. The chips take `UI::Button`'s `purple_outline` — the resting palette by reference, so retuning the button reaches them — and `full_width:` lays them out as equal grid columns that stay the same width when they wrap.

**The redesign blue becomes the Bike Index blue.** It turned out not to be a new color: `--redesign-blue-500` `#3397dc` and `--color-blue-500` `#3498db` are the same to within a unit per channel, and `--redesign-blue-800` `#016ec2` is exactly `blue-600`'s hue (206°) and lightness (38%) at 99% saturation instead of 77%. So `--color-blue-500` and up take the saturated values rather than carrying a parallel palette, `800`–`950` continue the ramp so the dark end doesn't fall back to the duller mix, and `kelsey_styles.css` points its blue and purple at the theme instead of restating them. `UI::Button`'s `blue` was `primary` with a different focus ring and no dark variants, so it's gone. `blue_link` was `twlink` one step lighter on hover and active with the focus outline suppressed and nothing replacing it, so `twlink` takes the lighter `blue-700` and `blue_link` goes too — a bold link is `color: :link` with `html_class: "tw:font-bold"`.

The gray badge keeps the redesign's `#f0f0f2`, but its text is `--redesign-gray-500` rather than the mockup's `#9a9aa2`, which sat at 2.45:1 against its own background — and `gray` is the default *and* the invalid-color fallback, so that reached every badge. Dark mode had the same problem (`gray-500` on `gray-800` is 3.03:1), so it's `gray-400`.

Worth a look: this repaints every blue and every default badge, moves every `twlink`'s hover and active color, adds an "optional" badge to every existing `UI::Forms::Group` field that isn't marked `required:` — the admin mail-snippet form, `AddressGroup`, the parking-notification form — and turns seven existing radio groups from segmented controls into chips.


sethherr:
## Screenshots

Lookbook previews. The four components that are new on this branch have no `main` counterpart, so those tables have no comparison row.

### /rails/view_components/ui/forms/group/component/text_field

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/radio_button_group/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/link_active

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/link_bold

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/forms/checkbox/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/forms/select/component/with_blank

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/forms/radio_button_group/component/full_width

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3957, opened 2026-07-26
Add Take picture and drag-to-drop to the file upload component

sethherr:
`UI::Forms::FileUpload` now offers three ways to attach a file — **Choose file**, **Take picture**, and dragging one onto the page — and previews whatever is already attached. All three fill the same input, so there's still one form param and one `change` event.

- **The drop target frames the controls** rather than sitting below them. Its border is transparent until a drag starts, so revealing it shifts no layout, and dropping anywhere on the component lands — including on the button, which is what the label promises.
- **Take picture** sets `capture="environment"` on the input before clicking it; the label clears it again. It shows only on coarse pointers, since desktop browsers ignore `capture` and the button would just re-open the file picker.
- **The label adapts to the pointer**: "Click or drop to choose file" for a mouse, "Choose file" for touch, which can do neither. Both ship and the media query picks one, so the hidden wording stays out of the accessibility tree too.
- **The current attachment's thumbnail** renders automatically, linked to the full size, so callers no longer guard on presence and hand-roll a preview. It goes through the record's own `<attribute>_url` — several models add a dev fallback or serve a processed copy — and covers CarrierWave and ActiveStorage.
- `camera:` defaults to on when nothing but images are accepted, so a future CSV or PDF field won't get a button that hands back a photo. `ApplicationUploader.thumbnail_version` is new: which version to preview with belongs next to where the versions are declared.

Worth a look: the coarse-pointer paths are only exercised through Chromium touch emulation (new `emulate_touch_device` helper — Playwright's `emulate_media` doesn't cover `pointer`), so whether iOS and Android actually open the camera is unverified on a real device. And `pointer: fine` hides the camera on an iPad with a trackpad attached, where `capture` would have worked; `any-pointer: coarse` would cover that but would also show it on Windows touchscreen laptops where it does nothing. I erred toward not showing a button that lies.

Also worth a look: each instance owns its own `@document` drag listeners, so a page with two upload fields would frame both at once. No page has two today, and consolidating would mean a page-level drag observer — flagging rather than building it. Separately, `label { margin-bottom: .5rem }` in bootstrap's reboot is now cancelled in two places (this component's label and `RadioButtonGroup`'s chips); killing the reboot rule would let both go, but that's a global change for another PR.


sethherr:
## Screenshots

Mobile shots run with Chromium touch emulation on, so `(pointer: coarse)` matches — that's what a phone sees: the short label, and the camera button the media query hides from a mouse. The drag state is a synthetic file drag (there's no OS drag source to use), and doesn't exist on touch.

### /rails/view_components/ui/forms/file_upload/component

The `with_existing_image` scenario is new, so the base row is its `default` scenario — the base has no thumbnail to show.

| Desktop | Mobile |
| --- | --- |
|  |  |
| sethherr/ui-component-updates-v1 👆 | this branch 👇 |
|  |  |

Mid-drag, with the file over the frame. The border is transparent until then, so revealing it shifts nothing:

| Desktop |
| --- |
|  |

### /o/brakebills/manage

The logo preview above the buttons used to be the view's own `if avatar? → link_to image_tag`; it now comes from the component.

| Desktop | Mobile |
| --- | --- |
|  |  |
| sethherr/ui-component-updates-v1 👆 | this branch 👇 |
|  |  |

PR #3964, opened 2026-07-27
Add the scanned sticker block to the registration show redesign

sethherr:
The redesigned registration show components never read `scanned_id`, so arriving from a sticker scan lost the "you scanned this sticker" block that `bikes/show` has had. `Registrations::Show::ScannedSticker::Component` ports it: the alert naming the code, plus the collapsed re-link form for viewers `authorized?` for the sticker.

- `RegistrationsController#show` resolves `?scanned_id` with `BikeSticker.lookup_with_fallback`, matching `BikesController#show`. It returns nil on blank input before touching the database, so unscanned page loads cost nothing and need no `present?` guard.
- The banner renders inside `Consumer` and `OrgAdmin` rather than above them in `Wrapper` — `Consumer`'s container carries a negative top margin that would overlap anything emitted before it. That puts it inside the fragment cache, so the sticker id joins the cache key.
- New `Atom::` namespace for the small value-rendering components: `Atom::Sticker`, `Atom::Serial` (its own translations and markup, byte-identical to `BikeHelper#render_serial_display` so the two can diverge later), and `UI::ShortId` moved across.
- `collapse#toggle` syncs `data-active` alongside `aria-expanded`, so a `UI::Button` trigger keeps its `is-active` styling while the panel is open.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -50,7 +50,7 @@ The `bin/dev` command handles building and updating Tailwind and JS.
 
 The same instinct applies beyond buttons: **check `app/components/ui/` and `app/components/atom/` before hand-rolling any UI primitive** (dropdowns → `UI::Dropdown`, tooltips → `UI::Tooltip`, form fields → `UI::Forms::*`, badges, modals, pagination, tables…). If a component exists for the pattern, use it; if it almost fits, extend it rather than forking its markup inline.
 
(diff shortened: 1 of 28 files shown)


sethherr:
## Screenshots

`?scanned_id=BR0` on a bike with that sticker claimed. The re-link form is shown expanded; it's collapsed on load.

### /registrations/71?scanned_id=BR0&view_as=owner

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /registrations/71?scanned_id=BR0&view_as=brakebills.staff

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3965, opened 2026-07-27
UI::Forms::Combobox renders no label - wrap it in UI::Forms::Group

sethherr:
The combobox styled its own label through the hotwire_combobox gem, so it was the one form field whose label skipped `twlabel` and the required / optional marker every other field gets from `UI::Forms::Group`.

- `label:`/`label_class:` are gone from `Combobox` and `ComboboxManufacturer`; every call site renders the field inside a `Group` block, the same pairing `Select` and `TextEditor` already document — previewed at `ui/forms/group/component/combobox`. The search bar is the exception: `Group` always appends a required / optional suffix, so its visually-hidden label stays a plain `label_tag`, matching the serial and price fields beside it.
- `Group` takes an optional `form_builder:` — without one it labels via `label_tag`, for the admin `form_tag` forms that have no builder. The combobox ids its input with its name in that case, so the label's `for` finds it, and it names the gem's mobile dialog the same way — that dialog covers the page below 640px, so the Group's label can't reach it.
- `Group` now infers "render the block instead of an input" from the block itself, so `kind: :content_block` is gone from all ~30 call sites (including the pre-existing `Select`/`TextEditor`/`AddressGroup` ones).
- Combobox previews moved off manufacturers, which are `ComboboxManufacturer`'s job, onto cycle types, primary activities and wheel sizes.
- Also carries `tw:justify-center` into `UI::Button::BASE_CLASSES` from sethherr/new-registration-flow (0a79c1778), with the per-call-site patches it makes redundant removed.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -54,6 +54,10 @@ The same instinct applies beyond buttons: **check `app/components/ui/` before ha
 
 **Every `UI::Tooltip` uses the default `?` button trigger** unless the user explicitly says otherwise — never pass a label as the tooltip's trigger content. See `app/components/ui/tooltip/`.
 
+## Form fields: the label comes from `UI::Forms::Group`
+
+**A `UI::Forms::*` field renders no label of its own** — wrap it in a `UI::Forms::Group` with `kind: :content_block`, passing `form_builder:` when there is one. Holds for `Combobox`, `Select`, and `TextEditor`. See `app/components/ui/forms/group/component_preview/` and `app/components/ui/forms/combobox/component_preview/`.
+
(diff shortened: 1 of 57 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/forms/combobox/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/group/component/combobox

| Desktop | Mobile |
| --- | --- |
|  |  |

### /search/registrations?stolenness=all

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The admin forms that carry most of this diff aren't captured — this workspace's dev DB has no seeded users, so `/admin/...` isn't reachable.

PR #3972, opened 2026-07-28
Restore the search button on narrow phone widths

sethherr:
The search submit button collapsed on narrow phones. Its `tw:w-20` wrapper is a flex item with the default `flex-shrink: 1`, so on `/search/marketplace` the fields column's min-content took its space back out of the button — and because `icons/searcher.svg` carries only a `viewBox`, the icon shrank with it: 80px/54px became 47px/21px at 375px and 39px/13px at 320px.

- `tw:shrink-0` pins the button on both search forms. That alone overflows the row, so the price inputs drop to `tw:min-w-16`, putting the column's min-content under what's available at 320px — they still render wider than the floor there, and desktop is untouched.
- `UI::Button` takes `spinner: true`, carried over from `sethherr/new-registration-flow`: it attaches a controller that disables the button and reveals a spinner once its form really submits. Not wired to a caller yet — it's only safe on forms that navigate, since it resets on reconnect, which rules out both search forms (they target a sibling turbo frame).
- Active state is flagged with `data-active` everywhere now, replacing `aria-pressed` on the org search settings and action-panel triggers and a `classList.add('active')` on modal triggers — that last one matched nothing in the `is-active` variant, so `UI::Button` modal triggers had no active styling at all.
- Unrelated, and needed to get CI green: `95b8f6e05` changed cached `registrations/show` markup without bumping `CACHE_VERSION`, so the checkpoint spec added in #3966 fails on `main` and the fragment cache is serving pre-centering HTML. Bumped to v4.

### DIFF
--- a/app/components/org/search/form/component.html.erb
+++ b/app/components/org/search/form/component.html.erb
@@ -55,7 +55,7 @@
         </div>
       </div>
 
-      <div class="tw:w-20">
+      <div class="tw:w-20 tw:shrink-0">
         <%= render UI::Button::Component.new(color: :primary, kind: :submit, html_class: "tw:w-full tw:h-full#{" tw:disabled:cursor-wait" if turbo?}") do %>
           <%= helpers.inline_svg_tag "icons/searcher.svg",
           id: "search-button",
--- a/app/components/org/search/form/component.html.erb
+++ b/app/components/org/search/form/component.html.erb
@@ -55,6 +55,7 @@
         </div>
       </div>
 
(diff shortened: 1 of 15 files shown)


sethherr:
## Screenshots

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/submitting

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3973, opened 2026-07-28
Autocomplete the manufacturer combobox, with free text for unknown manufacturers

sethherr:
The manufacturer combobox rendered every manufacturer into the page — 448kb of html on `admin/bikes/edit` — and offered `Manufacturer.other` as a pickable option. It now autocompletes against the Autocomplete index through `Search::ComboboxController#manufacturers`, which orders matches by manufacturer priority then alphabetically, so "trek" lands Trek above Trek Parts.

- A manufacturer that isn't indexed is entered through the "Unknown manufacturer" option — the manufacturer field's version of the search bar's "Search for" — which `BParam` (and now `Admin::BikesController`) resolves to `Manufacturer.other` plus `manufacturer_other`. Pass `no_manufacturer_other: true` where only an indexed manufacturer is acceptable, as the organization form and bulk manufacturer assignment do.
- Because the index matches on word prefixes rather than substrings, "trek" no longer matches AVANTREK at all. Manufacturers reach the index through `ScheduledAutocompleteCheckJob`, so a brand new one isn't pickable until that runs.
- Admin bike edit drops its `manufacturer_other` field for a warning under the combobox, which appears as soon as free text is entered. Also removes two dead hooks from that view: the `BASE_URL` div and `#country_select_container`, whose JS went away in b62edbf78.


sethherr:
## Screenshots

### /admin/bikes/:id/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/combobox_manufacturer/component/default

Typing "sur" into the field.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3974, opened 2026-07-28
Upload the register photo direct to R2

sethherr:
The register flow's bike photo uploads browser → R2 as soon as it's picked, so the form posts a signed blob id instead of megabytes of multipart body and submitting no longer waits on a synchronous S3 store. It lands as a `PublicImage` on the ActiveStorage backend from #3969, which strips EXIF and builds the variants off the request.

Stacked on #3969 — that's the base, so its commits are in the log here.

- Two direct upload endpoints, split by how they authorize. The registration one takes a `b_param_token`, since the photo is picked before there's an account; anyone can start a registration, so it's scoping rather than a gate and the hourly per-IP throttle is what limits it. The ActiveStorage route stays shadowed and requires a signed-in user, so its stock controller — which accepts anything from anyone — isn't reachable.
- `UI::Forms::FileUpload` does the uploading, behind `direct_upload_url:`: the file field goes nameless and a hidden signed-id field carries the blob, so the field posts nothing without JS. Off unless a caller names an endpoint, leaving the org avatar, parking notification and registration sequence fields alone.
- The signed id rides on the b_param as `params["image_signed_id"]` rather than an attachment, so the `PublicImage` built from it is the blob's only owner — and `CleanUnattachedBlobsJob` reaps whatever a registration abandons after 30 days, leaving the stolen alert images `RemoveOrphanedImagesJob` owns.
- Both endpoints and `PublicImage` hold an upload to a content type and size the app can serve. All three read what the browser declared, so they cap what can be written but don't verify the bytes.

The register integration spec uploads to a `bikeindex-test` bucket for real, which is the only way to exercise a presigned cross-origin PUT — it needs `R2_TEST_*` (repo secrets in CI, `.env.test` locally) and fails rather than skips without them.


sethherr:
## Screenshots

`/register?step=2`, anonymous — the case the registration endpoint exists for.

| | Before (#3969) | After |
| --- | --- | --- |
| Desktop | base desktop | branch desktop |
| Mobile | base mobile | branch mobile |

Before has no photo section at all — the base only renders it for a signed-in registrant, because the upload needed an account. After, it's there for everyone, uploading against the registration's token.

The field itself is `UI::Forms::FileUpload` rather than the flow's own markup, so it's the shared "Click or drop to choose file" control. Its camera button is `pointer-fine:hidden` and so doesn't appear in an emulated mobile viewport — on a real touch device it does.

PR #3976, opened 2026-07-28
Reverse-geocode the parking-notification pin server-side

sethherr:
The parking-notification map pin resolved its address through `api.mapbox.com` from the browser, while `ParkingNotification` re-geocodes with Google on save — so the address prefilled into the manual-entry fields wasn't necessarily the one that got stored, and the Mapbox token was rendered into the DOM.

- `GET /reverse_geocode` wraps `GeocodeHelper.assignable_address_hash_for` and returns `Geocodeable::ADDRESS_ATTRS` with the region resolved to a `region_record_id`, so the country and state selects are set by id instead of matched against option text. That drops `addressFromFeature`, `selectCountry` and `selectRegion` from the Stimulus controller.
- The response now only replaces a field that still holds what it did when the request went out — a server round trip is a hop slower than calling a provider directly, which widens the window where the fill lands on top of what someone has started typing.
- Deliberately outside `namespace :api`: that namespace is public+CORS or OAuth-token, and its rack-attack bucket (150/30s) is too generous for something proxying paid geocoding. It's `ReverseGeocodeController` rather than `GeocodeController` because Rails auto-includes a same-named helper, and `GeocodeHelper` is a `Functionable` module that can't be included.
- `MAPBOX_MAPPING` stays — `bikes/_stolen_map` still uses it.

### DIFF
--- a/app/components/registrations/show/org_top_actions/parking_notification_form/component.html.erb
+++ b/app/components/registrations/show/org_top_actions/parking_notification_form/component.html.erb
@@ -5,7 +5,6 @@
   data-registrations--show--parking-notification-form-notification-heading-value="<%= translation(".create_parking_notification") %>"
   data-registrations--show--parking-notification-form-impound-heading-value="<%= translation(".impound_this_bike_type", bike_type: @bike.type) %>"
   data-registrations--show--parking-notification-form-default-kind-value="<%= new_parking_notification.kind %>"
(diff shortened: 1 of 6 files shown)


sethherr:
Superseded by #3978 — the branch was renamed to `sethherr/parking-notification-latlng-test`, and a PR's head branch can't be retargeted.

PR #3979, opened 2026-07-28
Make admin bike edit + tabs view components, drop their legacy JS

sethherr:
`admin/bikes/edit` was the last big consumer of the vendored `binx_admin` bundle. The page is now `Admin::Bikes::Edit` + `Admin::Bikes::CreationData`, the `_bike_tabs` partial above it is `Admin::Bikes::Tabs`, and the jQuery they relied on is Stimulus. The form's Bootstrap fields become `UI::Forms::Group`, its two selectize selects `UI::Forms::Combobox`, and the bike_organizations table `UI::Table`.

- `admin--bike-edit-form` takes over both `BinxAdminBikesEdit` behaviors: the two "no serial" checkboxes (mutually exclusive, each overwriting the serial field with the sentinel `Bike#serial_number` expects — the original value rides a Stimulus value now rather than `window.originalSerialNumber`), and unchecking "Bike is stolen" to reveal the recovery fields. `#BParamsView` moves to the `ui--collapse` controller.
- `primary_activity_id` renders inline like the other admin comboboxes instead of through `/bikes_edit/bike_fields/_primary_activity`, so the user-facing pages keep their select. `bike_organization_ids` becomes a multiselect, which needs selection chips — hence `Admin::ComboboxController#organization_chips`, modelled on `Search::ComboboxController#chips`. It submits comma-joined ids, which `Bike#bike_organization_ids=` and the existing `permit` already take.
- Every field now carries the same required/optional marker as the comboboxes beside it, and the orgs table names its organization through `Admin::OrganizationCell` — which the association read as `nil` once the org was deleted, rendering the URL as its own link text.
- The tabs mark the active tab server side and scroll horizontally with `tw:w-max` rather than a jQuery width sum, so `initBikeTabs` matches nothing. `active_tab` comes from unvalidated `params[:active_tab]`, so instead of adding it to the fragment cache key, only the counts are cached and the markup renders uncached. The component reads `@bike` rather than the controller's ivar, and drops the never-passed `pdf` local and `admin/ownerships/edit`'s dead `target` local.
- Two N+1s the conversion surfaced: `PrimaryActivity#display_name`/`#display_name_search` read through `primary_activity_family`, making the options 49 queries — fixed at all three call sites, including `Search::Form`, and `Admin::OrganizationCell` re-queried because its template reads `organization_subject` three times.


sethherr:
## Screenshots

### /admin/bikes/26/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes/39?active_tab=ownerships

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

sethherr:
Replacing with #4183

PR #3980, opened 2026-07-28
Pull the shared UI updates out of the register flow branch

sethherr:
Lifts the parts of the registration flow (#3850, #3974) that aren't specific to it, so they land and get reviewed on their own.

- `UI::Forms::Combobox` gains `rich_display:` — `:inline` or `:stacked` — which mirrors the selected option's rich `content` onto the closed input through an overlay, since an `<input>` can't render two-tone text itself. `:stacked` puts a muted line under the title and grows the input to fit. The combobox now always renders in a wrapper, so `wrapper_class:` (for the parent's layout) applies either way and the markup doesn't shift when the rich display is toggled; only a rich display attaches the Stimulus controller.
- Enter now submits the form when the listbox is closed. hotwire_combobox cancels the key whether or not it has anything to select, so a combobox was the one input in a form that swallowed implicit submission.
- `form-persist` now covers checkboxes, radios, selects and hotwire_combobox pairs (hidden value + display text), skips file inputs, and defers its restore a tick so it can announce `form-persist:restored` to controllers whose UI hangs off restored fields. Fields that always render with a selection let the draft win — a combobox has no empty state, so "only fill what the server left blank" meant its draft never restored.
- `UI::Button spinner: true` resets on `pageshow` as well as `connect` — a bfcache restore resumes the page without reconnecting, so returning to a submitted form left the button stuck disabled.
- Also here: an `autofocus` controller that focuses the first visible field in its element, and the info icon `UI::Alert` had inlined moves to `app/assets/images/icons/info.svg`, in the outline style the rest of that directory uses.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -93,6 +93,7 @@ When deleting an `id`/`class`, grep the repo for the name before deciding what t
 This project uses the ViewComponent gem to render components.
 
 - Prefer view components to partials.
+- **If a view file only renders a single component, consider rendering it from the controller instead** (`render Foo::Component.new(...)`) and deleting the view file — the layout still wraps it.
 - Generate a new view component with `rails generate component ComponentName argument1 argument2`.
 - View components must initialize with keyword arguments. Everything the component needs must be passed in explicitly by the caller — never reach into controller state from inside a component (e.g. `controller.instance_variable_get(:@bike)`). If the component needs `@bike`, the caller renders `Component.new(bike: @bike)`.
 - In view components, use instance variables directly — don't add `attr_reader`/`attr_accessor`. Reference `@foo` everywhere, including in the template (`@current_user`, not `current_user`).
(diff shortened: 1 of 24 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/alert/component/notice

| Desktop | Mobile |
| --- | --- |
| alert notice on main, desktop | alert notice on main, mobile |
| main 👆 | this branch 👇 |
| alert notice on this branch, desktop | alert notice on this branch, mobile |

Solid glyph → outline, matching the rest of `app/assets/images/icons/`. Every `UI::Alert` that doesn't pass its own `icon:` picks this up.

The combobox previews below are new in this PR, so there's no `main` version to compare against.

### /rails/view_components/ui/forms/combobox/component/rich_display — filtered to "car"

| Desktop | Mobile |
| --- | --- |
| rich_display listbox open, desktop | rich_display listbox open, mobile |

### /rails/view_components/ui/forms/combobox/component/rich_display — after selecting

| Desktop | Mobile |
| --- | --- |
| rich_display selected, desktop | rich_display selected, mobile |

The parenthetical stays muted on the closed input — that's the overlay, since the `<input>` itself can only render one color.

### /rails/view_components/ui/forms/combobox/component/stacked — after selecting

| Desktop | Mobile |
| --- | --- |
| stacked rich display, desktop | stacked rich display, mobile |

`rich_display: :stacked` — the muted line sits under the title, and the input grows to fit both.

### /rails/view_components/ui/forms/combobox/component/in_form

| Desktop | Mobile |
| --- | --- |
| combobox in a form, desktop | combobox in a form, mobile |

Nothing to see here — this preview exists for the Enter-submits behavior, which a screenshot can't show. The system spec is the real check.

PR #3981, opened 2026-07-29
Add a UI::Forms::Email component that suggests domain typo corrections

sethherr:
Adds `UI::Forms::Email::Component`, a modern replacement for the legacy jQuery `CheckEmail`/Mailcheck: an email field whose Stimulus controller offers "Did you mean `you@gmail.com`?" when the field is left holding a near-miss of a domain we know, and swaps the address in when the suggestion is taken. Sign in and sign up use it; the remaining five legacy `CheckEmail` call sites move over in their own PR.

- The domain's name and its ending are matched apart, so an ending is corrected on a domain we've never heard of (`you@bikeshop.con`), while a name we don't know keeps its own. A part is corrected when it's a single typo — one letter dropped, added, mistyped, or swapped with its neighbor — from a known one, *and* that typo is a small part of it. Without the second half a two-letter ending is a typo of every other two-letter ending, and `protonmail.ch` gets offered `.ca`.
- Two things override that. A dropped letter is unambiguous however short it leaves the part, so `you@gmail.om` still reaches `.com`, and a name that exists as a `.com` and nothing else vouches for its own ending, which is the only way to read `you@gmail.co` as a typo rather than as Colombia. A dot with nothing on one side is normalized away before matching rather than spending the one typo the rest of the domain needs, so `you@.gmail.com` and `you@gmial..con` resolve too.
- Sign up drops the jQuery check for it — that was `UsersNew`'s only job, so the page script and its dispatcher entry go too. Sign in had no check at all before. Their name and password fields move to `twinput` alongside the email one, which otherwise differs from them in placeholder style and padding.
- Like `Combobox`, it renders no label: wrap it in a `UI::Forms::Group` block. The suggestion is a `UI::Button` inside an `aria-live` region that outlives it, since a region added along with its text has nothing to announce.
- It ships `tw:hidden!` rather than `tw:hidden` — `UI::Button`'s own `tw:inline-flex` sorts after `tw:hidden` in Tailwind's display layer and wins, which leaves an empty, unnamed button in the tab order. The `frontend-conventions` skill now says so.


sethherr:
## Screenshots

Each page below is holding a mistyped address and has just been left, so the suggestion is showing.

### /session/new

Typed `seth@yahoo.co`. There was no email check on this page before.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /users/new

Typed `seth@gmial.con`. The jQuery check answers `seth@gmx.com` — `gmail` isn't in its six-entry second-level list, so sift3 lands on `gmx` instead. Name and password move to `twinput` here, which is the placeholder and padding change.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/email/component/mistyped

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3983, opened 2026-07-29
Drop the combobox wrapper_class argument

sethherr:
Follows #3980, which gave `UI::Forms::Combobox` a `wrapper_class:` argument alongside the wrapper it now always renders. The argument never picked up a caller, and it only existed so a caller could reach through the component to the div — one it can just as easily wrap itself.

- Drops `wrapper_class:`. The register flow's flex-row combobox wraps itself instead.
- The wrapper's `tw:relative` moves behind `rich_display:` too. It exists so the overlay has something to position against, and the overlay only renders for a rich display — the other ~25 comboboxes were carrying a containing block for absolutely-positioned descendants that nothing asked for. The div itself still always renders, so the combobox sits at the same depth either way.

### DIFF
--- a/app/components/ui/forms/combobox/component.rb
+++ b/app/components/ui/forms/combobox/component.rb
@@ -18,8 +18,8 @@ module UI
       #   - :inline  one line, the muted part following the display text
       #   - :stacked a muted line below the display text, on a taller input
       #
-      # It always renders inside a wrapper div, so the markup doesn't shift when
-      # rich_display is toggled -- wrap it yourself for the parent's layout.
+      # It always renders inside a wrapper div, so the combobox sits at the same
+      # depth either way -- add your own wrapper for the parent's layout.
       #
       # Any other keyword (id:, value:, open:, free_text:, autocomplete:,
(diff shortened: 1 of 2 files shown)


sethherr:
## Screenshots

Both pairs should read as identical — the change removes a `class` attribute from a wrapper div, and nothing was positioned against it. The only difference in the preview shots is ViewComponent's render-timing pill in the corner.

### /rails/view_components/ui/forms/combobox/component/default

| Desktop | Mobile |
| --- | --- |
| combobox preview on main, desktop | combobox preview on main, mobile |
| main 👆 | this branch 👇 |
| combobox preview on this branch, desktop | combobox preview on this branch, mobile |

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
| marketplace search on main, desktop | marketplace search on main, mobile |
| main 👆 | this branch 👇 |
| marketplace search on this branch, desktop | marketplace search on this branch, mobile |

A real page rather than a second preview, since the risk this is checking is a layout shift at one of the ~25 call sites. The primary-activity combobox measures the same on both: `top: 197`, `width: 1152`, page height 487.

PR #3986, opened 2026-07-29
Add the e-vehicle acknowledgment pages to the register flow

sethherr:
Continues `/register` past step 2 into the organization's active `RegistrationSequence` — a page per `RegistrationSequencePage`, then a review with the final acknowledgment, and only then the bike. The seeded sequence is the e-vehicle acknowledgment from `kelsey_redesign/register-your-bike.html`, verbatim; builds on the flow #3850 added.

- Shows only for a **motorized** registration at an organization with an **active** sequence; every other registration still completes at step 2. Each page turns its body's `<li>`s into checkboxes and is agreed to as a whole — the server re-validates all-or-nothing, so a crafted POST can't skip it. Continue advances one page at a time, and earlier pages stay browsable and come back checked. Completion also claims the signed-in user as creator, which previously only step 2 did.
- **Activating a sequence freezes it and its pages.** Only archiving is allowed afterward; pages can't be edited, added or removed, and `reorder_page!` guards itself since `update_all` skips callbacks. A change belongs in a new draft, which activating supersedes the old one with.
- The agreement lands on its own `RegistrationSequenceAcknowledgment` — who agreed, to which sequence, and when. *What* they agreed to is read back off the frozen sequence rather than copied onto the record. Linked to the `b_param` in flight, then to the bike, since `CleanBParamsJob` sweeps the registration a day after the bike exists. Sequences are `acts_as_paranoid`, so destroying an organization can't take the rules an acknowledgment points at. It names whoever the registration is *for* (#4008's `user_name`), not the account filling the form in.
- Sequences gain `faq_url` and `acknowledgment_text`; pages gain `heading` (the design's flavour h2, falling back to `title`, which labels the rules and names the page on the review) and `organization_specific`, which badges a page with the organization's name. All editable from the org's draft editor, which gains a Settings form — and its index now lists the *active* sequence read-only, where it previously said "This is the active version your registrants see" above nothing.
- Seeds replace the template's pages with the prototype's three, grant the `registration_sequences` feature and activate a sequence for Brakebills — without an active one the pages never appear, so the flow couldn't be seen in development. Adds the E-Vehicle Acknowledgment FAQ info page the ⓘ links to.


sethherr:
## Screenshots

Branch-only — the safety-check and review steps are new routes, and this branch adds migrations, so a base-branch capture would run `main`'s code against an ahead schema.

### /register?…&step=3 — a page of safety rules

| Desktop | Mobile |
| --- | --- |
|  |  |

### /register?…&step=review — the final acknowledgment

Named for whoever the registration is *for*, and counted as the last step rather than a finished one.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /o/:slug/registration_sequences — the organization's sequences

"Current" now lists the active version, which previously showed nothing.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #3994, opened 2026-07-30
Restore the bike show overlays' token prompts on the redesign

sethherr:
Every token link in our emails points at `/bikes/:id`, which redirects redesign users to `/registrations/:id` — where none of the bike-show overlays existed, so the claim invitation, the parking/graduated notification prompt and the stolen-recovery form silently vanished for anyone on the redesign. This ports them into a `Registrations::Show::CurrentAlerts` namespace, alongside a new alert telling a sender their registration is still awaiting the new owner's claim.

- `BikeServices::ShowCurrentAlerts` resolves the three token types and the controller keeps the session protocol (spending the recovery token as it reads it, recording the claim email). Only one prompt is earned, following the legacy partial's precedence.
- **`PromptChrome` renders that prompt twice** — as the dialog that opens itself, and as an alert in the page body holding the same thing whole, since closing the dialog is otherwise the end of it. The dialog stays outside the page's fragment cache; the alert sits inside it, so the key identifies *which* prompt the request earned — two notifications on one bike are the same component, and the alert carries its token.
- **`Registrations::Show::Wrapper` has a Lookbook preview** — one scenario per overlay plus the page with none, taking the view and a bike as params. Views resolve through `BikeServices::ShowViews`, so a scenario can't show a role the real page won't. It renders a persisted registration, so it's gated out of production, where Lookbook is mounted but the bikes are real. A spec fails if an alert is ever added without a scenario.
- The scanned-sticker re-link trigger spins a chevron, `Registrations::Show::{Consumer,OrgAdmin}` became `Wrapper{Consumer,OrgAdmin}`, and each nested component namespace needs its own `i18n-tasks` routing line to keep its translation sidecar.

`_bike_show_overlays.html.haml` is untouched and still serves users who aren't on the redesign.

Because the failure this fixes is silent, `show_alerts_request_spec` walks each email link the whole way — link, redirect, rendered prompt — rather than posting to `/registrations/:id` directly.


sethherr:
## Screenshots

Every state `Registrations::Show::CurrentAlerts` can raise, from the Lookbook previews this PR adds — `/lookbook` → Registrations → Show → Wrapper, one scenario per overlay. Seeded Brakebills data.

### The token prompts

Each opens itself on arrival, and each renders again as an alert in the page body, so dismissing the dialog doesn't lose it.

#### Claim invitation — `?t=<ownership token>`

| Desktop | Mobile |
| --- | --- |
|  |  |

Dismissed, the same prompt is still there — this is what the other three look like behind their dialogs:



#### Parking notification — `?parking_notification_retrieved=`

The org's own mail snippet, and the form to mark it retrieved.

| Desktop | Mobile |
| --- | --- |
|  |  |

#### Graduated notification — `?graduated_notification_remaining=`

Same component, "Mark bike remaining" rather than "retrieved".

| Desktop | Mobile |
| --- | --- |
|  |  |

#### Stolen recovery — reached from `/bikes/:id/recovery/edit?token=`

| Desktop | Mobile |
| --- | --- |
|  |  |

### The alerts about the registration's own state

These aren't token-scoped, so they render in the page body without a dialog.

#### Scanned sticker — `?scanned_id=`, viewer authorized to relink

| Desktop | Mobile |
| --- | --- |
|  |  |

#### Sent to a new owner, not yet claimed

| Desktop | Mobile |
| --- | --- |
|  |  |

The `no_overlay` and `claim_invitation_signed_out` scenarios are in Lookbook too — the first is the page with none of these raised, the second the claim invitation over the public view, which is what the email's recipient actually sees.

PR #3995, opened 2026-07-30
Open the mobile menu below the hamburgler, not under it

sethherr:
The review-app banner pushes the navbar down, but the small screen menu and its backdrop are `position: fixed` with a hardcoded `top` — so the hamburgler bar, which is in normal flow and layered above the menu, covered the top of the open menu. A PR title long enough to wrap to three lines buried the whole first item.

- `positionMainMenu` measures the hamburgler's real bottom edge when the menu opens, and again on resize while it's open, rather than assuming the navbar starts at the top of the viewport.
- Retires the two SCSS constants (48px and 60px) that were each separately approximating the navbar's ~66px height. They collapse into one variable that now only governs the pre-JS first paint.


sethherr:
## Screenshots

### /

Viewport shots rather than full page — the menu is a fixed overlay — and the review-app banner is left visible, since it's what triggers the bug. On the local dev server it wraps to two lines (65px), which buried "Search" and half of "Marketplace" under the hamburgler bar.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #3996, opened 2026-07-30
Hold the form while the email field has something to say

sethherr:
The email field only checked itself on `focusout`, so anything it had to say could be submitted straight past. It now checks on enter too, and holds the form until the message is answered.

- **Held submit.** While a message shows, the form's submit is disabled and a "Submit form anyway" link appears beside it. Enter in the field is held too; enter on the suggestion still accepts it. Releasing puts back only what the hold took — `ui--button--submit-spinner` disables the same button to see off a second submit, and a clean focusout has no business undoing that.
- **Reserved domains.** An address at an RFC 2606 domain — `you@example.com` most often, the placeholder typed in rather than replaced — gets "That doesn't look like your real email address. Please enter an email address where you receive email". There's no correction to offer, so it's text rather than a link.
- **One definition of the rule.** That moves to `EmailDomain::RESERVED_REGEX` (from `SpamEstimator::Bike`, which already scores it 100). The component hands the pattern's source to the controller instead of keeping a second copy, so client and server can't disagree — a component spec pins the JS-readable syntax.
- `le-btn-primary` gets a `:disabled` look it had no rule for, since the landing demo modal's submit is one the field now holds.

Worth a look before merge: this holds **every** caller, including sign-in (`sessions/new`), where whether an address is real is the server's answer and the copy is wrong-footed — say so and it can take a per-caller opt-out. And `register/step1`'s placeholder is still literally `you@example.com`, which the field now warns about.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -21,7 +21,7 @@ Run `bin/lint` to automatically format the code. Always use `bin/lint`, don't us
 - Omit named arguments' values from hashes (ie prefer `{x:, y:}` instead of `{x: x, y: y}`)
 - Prefer less code, by character count (excluding whitespace and comments). Use `bin/char_count {FILE OR FOLDER}` to get the non-whitespace character count
 - prefer un-abbreviated variable names
-- Keep comments pithy — often they aren't necessary. Explain *why* for a future reader; don't narrate the change that introduced the code
+- Keep comments pithy — often they aren't necessary. Explain *why* only where a reader would otherwise get it wrong; don't narrate the change that introduced the code, and don't defend a choice against an edit nobody would make — a failing test already defends it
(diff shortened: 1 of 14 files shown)


sethherr:
## Screenshots

Both are the component's own previews. `main` only checks on `focusout`, so that's the interaction used on both sides — on this branch the same state is what enter now reaches instead of submitting. The `Register` submit is new to the previews here; there was nothing on `main` for a message to hold.

### /rails/view_components/ui/forms/email/component/mistyped

Leaving the field with `you@gmial.con`. Only the address is a button — the sentence around it is text.

| Desktop | Mobile |
| --- | --- |
| mistyped on main | mistyped on main, mobile |
| main 👆 | this branch 👇 |
| mistyped on this branch | mistyped on this branch, mobile |

### /rails/view_components/ui/forms/email/component/default

Leaving the field with `you@example.com`. `main` has nothing to say about it.

| Desktop | Mobile |
| --- | --- |
| reserved domain on main | reserved domain on main, mobile |
| main 👆 | this branch 👇 |
| reserved domain on this branch | reserved domain on this branch, mobile |

Not pictured, since it's a transition: typing anything clears the message and re-enables the submit at once.

PR #3997, opened 2026-07-30
Add an admin index for public images

sethherr:
`/admin/public_images` — a sortable, filterable table of every public image, built on `Admin::IndexSkeleton` and linked from the developer admin nav.

- Filters by imageable type, kind, and private (dropdowns for the first two, a nav toggle for private); type and kind also get a 🔎 on each row. The graph is a stacked column chart with one colored series per imageable type.
- Reads through `PublicImage.unscoped`, so admins see private images too — the model's `default_scope` hides them and pins an ordering the sortable table needs to replace.
- Rows are fragment-cached, because the humanized size column resolves through Fog with a request per image. `update-cached-sortable-links` rewrites the 🔎 links so the cached markup still respects the current filters.
- Each row names its imageable via a new `PublicImage#imageable_name`, falling back across `display_name`/`name`/`title`. `ImpoundClaim`, `MailSnippet` and `SocialPost` don't name themselves at all, so those rows show `#<id>`.
- Adds a concurrent index on `public_images.created_at`, which the page filters and sorts on, and teaches `admin_path_for_object` that a Blog is administered as `/admin/news` — the generic fallback pointed at `/admin/blogs/:id`, which isn't a route, so this also fixes Blog links on the paper-trail-versions page.


sethherr:
## Screenshots

### /admin/public_images

| Desktop | Mobile |
| --- | --- |
|  |  |

### /admin/public_images?search_size=true

| Desktop | Mobile |
| --- | --- |
|  |  |

### /admin/public_images?render_chart=true

| Desktop | Mobile |
| --- | --- |
|  |  |

Seeded data is 8 carrierwave images plus one attached row, so the AS column has a single check, the chart's other six type series are empty, and nothing is private. `?search_size=true` is included because the Size column is off by default.

PR #3998, opened 2026-07-30
Preview the pick in UI::Forms::FileUpload, use it in register step 2

sethherr:
`UI::Forms::FileUpload` now previews an image the moment it's picked, from a local object url, so a registrant sees what they chose without waiting on a submit.

- The drop frame spans the field — an `outline` rather than a `border`, so no padding is held in reserve for it and the controls line up with the field's own label instead of sitting indented under it.
- The two pointer-media-query label wordings collapse into one **Upload** button with an upload icon. The camera button is unchanged.
- Register step 2 renders the shared component instead of its own hidden input, pair of buttons and `register--photo` controller, all deleted. Its `show_photo?` gate stays: the photo still posts its bytes through `b_param.image`, so it's still withheld from an unconfirmed anonymous registrant.
- Separately, `UI::Modal` takes `open:` so the server can render a dialog already open, skipping the url param a trigger-opened modal persists.

The FileUpload and modal changes are the halves of #3974 and `sethherr/dev-server-up` that don't need direct upload, pulled forward so those branches shrink to their own subject.


sethherr:
## Screenshots

### /register?b_param_token=…&step=2

A photo picked, so the preview shows.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/file_upload/component/with_existing_image

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/modal/component/open_on_connect

New preview, so nothing on main to compare to.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4004, opened 2026-07-31
Add a new superuser ability page to admin

sethherr:
`/admin/superuser_abilities/new` — granting an ability was previously only possible from the user edit page (universal only) or the console. The index gets a button linking to it, carrying its `user_id` filter through so the form arrives prefilled.

- The user is found by email, username or id through the existing `User.friendly_find`, held on a `user_identifier` accessor the way `Membership` holds `user_email`. Leaving controller and action blank makes it universal; `kind` stays derived.
- `set_calculated_attributes` now writes blank `controller_name`/`action_name` as nil, so a form-created universal ability matches the `find_or_create_by(controller_name: nil, action_name: nil)` in `Admin::UsersController` instead of quietly becoming a second row.
- Validates `user_id` — a typo'd address re-renders the form rather than creating an unowned universal ability. On the column rather than the association, so an ability outliving its soft-deleted user stays editable.


sethherr:
## Screenshots

### /admin/superuser_abilities

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/superuser_abilities/new

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4006, opened 2026-07-31
Only show the impound update action for the org's own record

sethherr:
A staff member at an `impound_bikes` org viewing an impounded bike got the **Update Impound Record** action no matter who impounded it — the form then pointed into *their* org's impound routes, which the record isn't in. An unorganized impound record has a nil `display_id` by design, so `organization_impound_record_path` couldn't build a URL at all and the registration page 500'd ([133103079](https://app.honeybadger.io/projects/35931/faults/133103079)).

- Both the action button and its panel now require `current_impound_record.organization_id` to match the viewing org. Another org's record silently pointed at a display_id the org's `friendly_find!` would 404 on, so that case was broken too, just more quietly.
- The wrapper preview's impounded bike carries an impound record now, so the preview renders the update panel it's meant to show.

### DIFF
--- a/app/components/registrations/show/org_top_actions/impound_update/component.rb
+++ b/app/components/registrations/show/org_top_actions/impound_update/component.rb
@@ -13,7 +13,7 @@ module Registrations
           end
 
           def render?
-            @bike.status_impounded? && impound_record.present?
+            @bike.status_impounded? && impound_record&.organization_id == @organization.id
           end
 
           private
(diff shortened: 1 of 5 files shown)


sethherr:
## Screenshots

### /rails/view_components/registrations/show/org_top_actions/wrapper/component/default?panel=impound_update

The preview's impounded bike now carries an impound record, so the update panel renders. On `main` the same action button activates and opens nothing — the panel it points at never rendered.

| Desktop | Mobile |
| --- | --- |
| main - the impound update action opens an empty panel | main mobile |
| main 👆 | this branch 👇 |
| this branch - the update form renders | this branch mobile |

PR #4008, opened 2026-07-31
Ask who the registration is for when it isn't the registrant

sethherr:
Step 2 of the register flow now collects an owner name whenever the registration is going somewhere other than the registrant's own address — anonymous registrations always, and a signed-in registrant who put someone else's email in step 1.

- **Step 2, so no JS.** Step 1's email settles who the registration is for, so the server knows at render time whether to ask: the field is either there and `required` — held by the browser — or it isn't. `update` re-checks, re-rendering step 2 with what they entered rather than saving a half-finished registration.
- **`BParam#self_made?`** answers whether the registrant is the owner — the b_param side of `Ownership#self_made?`, named to match — their primary plus confirmed additional emails, normalized.
- **One gate for the org's registration fields.** `BikeServices::Displayer.include_reg_field?(field, organization, user)` replaces the five `include_field_reg_x?` helpers and `Builder.include_address_record?`; step 2 passes the user only for a `self_registration?`, so registering for someone else asks for their phone and address rather than skipping on what the registrant's own account holds.
- **No new plumbing for the name.** `user_name` is already in `BParam::REGISTRATION_INFO_ATTRS`, so it rides into `Ownership#registration_info` and `fallback_owner_name` turns it into the bike's `owner_name`.
- **Unrelated fix along for the ride.** `/register?organization_id=brakebills` — an organization's own link — redirects into `/register/new` when there's no registration to resume, and dropped every param on the way, so the registration it created went unattributed. `status` and `email` were lost the same way.


sethherr:
## Screenshots

Registering for someone else — `friend@bikeindex.org` — throughout.

### /register?step=2

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /register?step=finished

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4010, opened 2026-07-31
Alert: full-width body when a header is present

sethherr:
`UI::Alert` rendered with a `header:` now runs its body across the full alert width, on its own line below the header row, instead of indenting it into the icon's column.

- **The header shares the icon's line, the dismiss button stays with it.** `flex-1` on the header stops it wrapping below the icon at narrow widths (`flex-wrap` breaks a line before it shrinks anything), and `order-last` on the body lets it wrap past the button.
- **Vertical rhythm moved into the component** via `gap-y-2`, replacing the `mt-2` each `header:` caller had been adding for itself — `admin/dashboard/scheduled_jobs` never did, and its list sat flush against the header.
- **`Alerts::ObjectError` reads "1 error" / "2 errors prevented this User from being saved".** The count had no noun before. The key is renamed rather than restructured in place, so the four `translation.*.yml` entries — translations of the same defect, interpolating `%{errors_count}` — aren't handed a `count:` they can't use. **They need a translation.io sync**; until then non-en locales fall back to `:en` in production only (`config/application.rb`'s hash-form `i18n.fallbacks` drops the `:en` default in dev/test, so a non-en locale raises there — pre-existing, and it makes every en-only key a dev/test landmine).
- **Its bullets render again.** The `<li>`s carried `tw:block`, and `display: block` suppresses the list marker, so the `tw:list-disc` on the `<ul>` had never shown.

### DIFF
--- a/app/components/admin/bikes_table/component.rb
+++ b/app/components/admin/bikes_table/component.rb
@@ -8,7 +8,7 @@ module Admin
     # checkboxes, and skip_user to drop the owner column.
     class Component < ApplicationComponent
       # Digest of the markup inside the row cache — the cached_markup_digest spec keeps it current
-      MARKUP_DIGEST = "e88d056eec9a"
+      MARKUP_DIGEST = "55bd93d8018c"
 
       def initialize(bikes:, no_show_header: false, show_serial: false, render_sortable: false,
         skip_user: false, render_multi_check: false)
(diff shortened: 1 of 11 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/alerts/object_error/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/alert/component/dismissable_notice

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/alert/component/dismissable_with_header

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4011, opened 2026-07-31
Scroll long modal bodies instead of pushing them off-screen

sethherr:
A `UI::Modal` taller than the viewport centered itself, pushing the title bar and its close button off-screen with no way to scroll or dismiss it. The dialog is now the scroll container, capped at the viewport, with the title bar pinned by `sticky`.

- **Scrolling the dialog rather than the body carries the panel with it**, so both rounded ends only come into view when you reach them instead of sitting at the screen edge the whole way down.
- **`overflow-clip` on the panel, not `overflow-hidden`** — `hidden` would make the panel a scrollport of its own and leave the sticky title bar with nothing to stick to.
- **`Registrations::Show::Wrapper`'s `MARKUP_DIGEST` moves** — the modal renders inside its fragment cache, so the key has to change for the new markup to be served.
- A long-content modal joins the `default` preview, and a new `resize_window` spec helper forces the overflow rather than leaving it to depend on the fixture out-growing the 1920×1080 test viewport.

### DIFF
--- a/app/components/registrations/show/wrapper/component.rb
+++ b/app/components/registrations/show/wrapper/component.rb
@@ -8,7 +8,7 @@ module Registrations
       class Component < ApplicationComponent
         # Digest of the markup inside the cache block — the cached_markup_digest spec
         # keeps it current, following what this tree renders out into UI:: and elsewhere
-        MARKUP_DIGEST = "5f851681e24e"
+        MARKUP_DIGEST = "e6aac93ab137"
 
         def initialize(bike:, current_user:, view:, available_views:, bike_sticker: nil)
           @bike = bike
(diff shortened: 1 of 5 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/modal/component/default?modal_long-modal=1

New preview scenario, so there's nothing on `main` to compare against — this is the state the PR fixes. Opening scroll position: the rounded top is inset, the bottom runs off-screen.

| Desktop | Mobile |
| --- | --- |
|  |  |

Scrolled to the end — the rounded bottom and its margin come into view, and the title bar has stayed put:

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/modal/component/default?modal_settings-modal=1

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

A modal that fits is unchanged — the only difference is the new "Open Long" trigger behind the backdrop.

PR #4013, opened 2026-07-31
Open modals with the browser's own command/commandfor

sethherr:
A modal trigger's click could be swallowed: `ui--modal` wires it in `connect` and Stimulus lazy loads controllers, so a click landing before that module arrived did nothing, and there was nothing for a retry to catch. Triggers now carry the platform's own `command`/`commandfor`, so a browser with invoker commands opens the dialog whenever the click lands.

- **`commandfor` is the only name for the dialog now** — `data-open-modal` is retired, and a spec scans for it, since `UI::Button` forwards an unknown kwarg verbatim: a trigger left on the old name renders, passes a spec asserting the attribute, and opens nothing.
- **The two paths don't overlap.** Where invokers exist the controller leaves the click alone — a second `showModal` warns — and `#invoked` keeps the trigger, the URL param and the scroll lock, or adopts an already open dialog when it connects late. Where they don't, one `[commandfor]` lookup wires trigger and close button as before.
- **Closing funnels through the dialog's own `close` event**, which the close command, Escape and `#close` all end up firing, so the cleanup runs once wherever it started — and the hand-rolled Escape handler duplicating the platform's goes.
- Invoker commands are Baseline as of December 2025 (Chrome 135, Firefox 144, Safari 26.2), so the fallback carries the rest for now — held to the same behaviour by a spec that puts the detect back to false.


sethherr:
## Screenshots

Nothing to see, which is the point — the dialog is identical, it just doesn't wait for Stimulus to open.

### /rails/view_components/ui/modal/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4016, opened 2026-08-01
registrations#show: the viewer's impound claim, and visibility specs

sethherr:
The redesigned `registrations#show` had no coverage for the visibility rules it gates on, and its impound-claim card only ever offered to open a *new* claim — a viewer who already had one was shown the open-claim form again, with no way to reach the message and submit steps.

- **An existing claim replaces the open-claim form.** Unsubmitted shows the message form and a submit button, submitted shows when it went in plus the outcome. A rejected claim falls back to offering a new one, matching `not_rejected`.
- **The bike a claim was submitted *with* links to the bike it claims.** `display_impound_claim?` matches that side too, and the card had been calling `impound_record.id` on nil there.
- **The card becomes one of the current alerts** — a `UI::Alert` stacked with them in the sidebar rather than a full-width white card, taking the organization so it stays out of the staff panel, which renders the same alerts. It previews from `Wrapper::ClaimImpound::ComponentPreview`, one scenario per state.
- **`WrapperConsumer#cache_version` folds in the viewer's claim.** The card renders inside the show tree's fragment cache, keyed on user and bike — neither moves when a claim is saved, so "Save message" appeared to do nothing.
- **Specs for `user_hidden`, `likely_spam`, and short_id lookups**, which `show` gates on through `visible_by?` and `Bike.unscoped.find_id` but which only had legacy `bikes#show` coverage. `/registrations/r/z` doesn't route — the slash-bearing form reaches the redesign only through the `/r/` → `bikes#show` redirect.

Two things the diff doesn't explain. `UI::Button` takes `name`/`value` because submitting a claim is a second submit button inside the message's own form — the legacy page did the same job with JS writing a hidden status field, and anything else drops a message the claimant typed but didn't save. And `spec/requests/component_previews_request_spec.rb` is retired: what it checked was either an interaction, which belongs in a browser, or what a preview resolves, which belongs beside the component — the axe audit that came with the move caught the add-a-stolen-bike link being distinguishable by color alone.


sethherr:
## Screenshots

A found bike the viewer doesn't own, viewed by a signed-in user who already has an open claim on it. The card stacks with the other current alerts, so it moves from a full-width white card into an alert in the sidebar.

### /registrations/48

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

---

One scenario per state the card reaches, all new here, so there's nothing on `main` to compare against. Under `/rails/view_components/registrations/show/wrapper/claim_impound/component/`. The alert is yellow while the claim is still waiting on somebody and green once it's answered.

### with_stolen_registration

| Desktop | Mobile |
| --- | --- |
|  |  |

### signed_out

| Desktop | Mobile |
| --- | --- |
|  |  |

### without_stolen_registration

| Desktop | Mobile |
| --- | --- |
|  |  |

### unsubmitted

| Desktop | Mobile |
| --- | --- |
|  |  |

### submitted

| Desktop | Mobile |
| --- | --- |
|  |  |

### approved

| Desktop | Mobile |
| --- | --- |
|  |  |

### submitted_with_this_bike

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4017, opened 2026-08-01
Convert the user general alert into PageBlock::UserAlerts components

sethherr:
The general alert on every page moves out of `shared/_user_general_alert` into `PageBlock::UserAlerts`: one component per alert kind, picked by a wrapper the layout renders, each previewable in Lookbook. It also gains a fourth kind, `unfinished_registration`, which rides on the `alertable` pair #4032 added.

- **The alerts that don't interrupt now render below the navbar** rather than pinned to the bottom of the window, as `UI::Alerts::Base` notices. `Banner` and `BikeListModal` hold the chrome the kinds share, the way `CurrentAlerts::PromptChrome` does; the modals move onto `UI::Modal`, since the jQuery/bootstrap `shared/modal` partial can't render in a preview. **The two modal alerts keep a banner behind them**, so dismissing the modal isn't the end of it.
- **`unfinished_registration` alerts on a register-flow `BParam` that never made its bike** — including one that only lacks its e-vehicle acknowledgment — for as long as its token still resumes it (`TOKEN_EXPIRATION`). `RegisterController` sets `@skip_general_alert`, so it doesn't sit on top of the flow that resolves it.
- **A `BParam` `after_commit` maintains it**, rather than only `AfterUserChangeJob`: walking away from the flow is exactly when nothing else bumps the user, and enqueuing the job instead would fan out an `AfterBikeSaveJob` per bike on every step. `UserAlert.update_unfinished_registrations` resolves off the alerts rather than off the registrations still in scope, since expiry is silent and `CleanBParamsJob` deletes without callbacks — the job also backfills registrations abandoned before this.
- **`UserAlert.general_kinds` is now priority-ordered** and is what the wrapper selects on, so a fifth general kind can't render nothing silently — a spec globs the component directories and fails if one has no preview scenario and no spec.
- The four alerts' copy moves to sidecar `component.en.yml`, so **es/it/nl/nb need a translation.io sync** — until then they fall back to `:en` in production.


sethherr:
## Screenshots

`/my_account` shows the banner where it lands — below the navbar, above the page. The preview
routes are new in this PR, so there's no base-branch comparison; the phone alert is captured at
a short viewport since a standalone banner preview is otherwise an empty page.

### /my_account (unfinished_registration)

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/page_block/user_alerts/wrapper/component/phone_waiting_confirmation

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/page_block/user_alerts/wrapper/component/stolen_bike_without_location

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/page_block/user_alerts/wrapper/component/theft_alert_without_photo

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/page_block/law_enforcement_donation/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4019, opened 2026-08-01
Sign users in from emailed links instead of a password

sethherr:
New signups no longer pick a password. `users.passwordless_user` marks them, the confirmation email is a sign in link rather than a verification button, and entering their email at `/session/new` mails a fresh link and lands on a new "we sent you a magic sign in link" page.

- **They're offered a password on every sign in, not just at signup.** The flash links to the reset form, which now accepts a signed in user without a token so it doesn't have to email one. Organization passwordless users (magic link or SSO) just get "You're signed in" — their organization is what signs them in, so there's nothing to set.
- **Passwordless users get a random digest from `set_calculated_attributes`**, which is what `Users::ProcessOrganizationRoleJob` had been doing inline for org-created users, so `has_secure_password` and the password validations are untouched. That job now just passes the flag.
- **Flash keys suffixed `_html` render as markup.** The flash cookie is JSON, so it drops `html_safe` across the redirect and a link in a flash needs an explicit opt-in. Wired into all three flash render sites.
- Existing org passwordless members now skip the intermediate "Email me the link" screen, since they're `passwordless_user` too. The screen still handles a passwordless-domain email with no account yet, which is what mints it.


sethherr:
## Screenshots

### /users/new

| Desktop | Mobile |
| --- | --- |
|  |  |

### /session/new

| Desktop | Mobile |
| --- | --- |
|  |  |

### /session/magic_link_sent

| Desktop | Mobile |
| --- | --- |
|  |  |

### /my_account — the sign in flash

Signed in through the emailed link. The donation modal is hidden, it covers this on a seeded dev database.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4020, opened 2026-08-02
Render flash messages with UI::Alerts::FlashMessage

sethherr:
The four layouts that rendered flash through `/layouts/revised_messages` now render the component instead, the way admin already did. The partial stays behind for `@page_errors` and the mustache template `BikeIndexAlerts` renders client-side alerts into, so the JS alerts on the bike-edit and payments pages keep working.

- **`flash[:info]` becomes `flash[:notice]`.** `info` isn't one of `KINDS` — 24 call sites across 14 controllers set it directly, four more reach it by passing `flash_type: :info` to `authenticate_user`, and four views passed it as a kind, where it silently fell back. Renamed rather than aliased in the component, so the flash type and the kind it renders stay the same word. `KINDS` now derives from `TEXT_CLASSES`.
- **The unknown-kind guard raises in dev and test, falls back to notice elsewhere.** Rails sweeps the flash after the layout renders, so a raise from the layout leaves the offending key in the session and raises again on the next request, and the next.
- **Every flash value is a string now.** Six sites set an Array (`errors.full_messages`) or a `flash[:errors]` key; the component renders neither, and the old bootstrap partial rendered the Array's `inspect`. Five get `to_sentence`, and the theft alert plan forms get `UI::Alerts::ObjectError` above the form like the other admin forms.
- **The impound multi-update's turbo_stream replaces `#flash-messages`** alongside the results frame, rather than rendering a second copy of the flash inside the frame. The component renders its container even when empty so there's something to target, and the replace only fires when there's a flash — otherwise every filter and sort would wipe an alert the rider hadn't dismissed.
- **`UI::Alert` moves to `UI::Alerts::Base`**, so all three alert components share the namespace. The three `MARKUP_DIGEST` bumps are that rename reaching templates inside fragment caches.

Flash now sits bottom-center instead of top, and success messages no longer auto-fade after 10s — the component's behavior, already what admin gets.


sethherr:
## Screenshots

### /session/new (failed login)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4023, opened 2026-08-02
Guard the alert kind, and let long alert text run the full width

sethherr:
Follow-up to #4020. An unknown alert kind used to fall back to notice silently, so a caller that passed a typo'd or unhandled kind got a blue box and no signal. And a dismissable alert's text could only wrap inside its own column, so a long message pushed the dismiss button onto a line of its own.

- **`kind` defaults to `:notice`, and anything outside `KINDS` raises.** An omitted kind and a wrong one used to be the same input; now only the wrong one is an error. Deployed environments notify Honeybadger and fall back instead of raising — the flash is swept after the layout renders, so a raise from there would raise again on every subsequent request. That's `production? || sandbox?`, per the gotcha in `config/environments/sandbox.rb`.
- **`UI::Alerts::FlashMessage` drops its own copy of that guard** and hands the flash type straight to the alert.
- **The icon and dismiss button float rather than sitting in a flex row**, so an alert's first line renders between them and every line after runs the full width. A float displaces only the lines it overlaps, which a flex item can't do — its text is stuck in its own column. Both are kept to about a line tall and precede the text, which is what floats require; the dismiss button moving above the message in the DOM also puts it first in tab order.
- **Preview reorganization**: the warning kind variant is now the header variant, the separate header group is gone, and the dismissable group picks up a header variant and a long-text purple one that exercises the wrapping above.

The `MARKUP_DIGEST` bumps are these edits reaching templates inside fragment caches.


sethherr:
## Screenshots

### /rails/view_components/ui/alerts/base/component/dismissable_with_header

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/alerts/base/component/dismissable_purple

New preview, so there's nothing on `main` to compare against — it's the long-text case the float layout is for.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4024, opened 2026-08-03
Force SSO on sign-up, not just sign-in

eld120:
An SSO organization's IdP owns every account on its email domain, and `SessionsController` redirects an SSO-managed email to that IdP before any sign-in action runs. Sign-up had no such guard, and `identify` sends a not-yet-registered email straight to `/users/new?email=…` — so the ordinary "I don't have an account yet" path let an SSO-managed email register a Bike Index account the IdP has never heard of, with its own confirmation email and login. From then on that account signs in outside the org's control.

- **`redirect_forced_saml` now covers `UsersController#new` and `#create`.** It moves to `Sessionable` alongside `submitted_email` rather than being duplicated, since both controllers include it — and `submitted_email` picks up the sign-up form's `user[:email]` next to the sign-in form's `session[:email]`.
- **An org whose SAML configuration isn't live yet still falls through to the normal form.** `Organization.saml_email_matching` already requires `configured?`, so enabling the feature before the IdP details land doesn't lock the domain out of signing up.


spec/requests/users_request_spec.rb
```diff
@@ -10,6 +10,24 @@
...
+      end
+      let!(:saml_configuration) { FactoryBot.create(:organization_saml_configuration, :enabled, organization:) }
+
+      it "hands the prefilled email off to the IdP rather than the signup form" do
+        get "#{base_url}/new", params: {email: "student@sso.edu"}
+        expect(response).to redirect_to(saml_init_path(org_slug: organization.to_param))
+      end
+
+      it "renders for an email the org doesn't claim" do
+        get "#{base_url}/new", params: {email: "student@example.edu"}
```
sethherr, line 25: combine or wrap in context blocks with let(:email)

PR #4035, opened 2026-08-04
Cut Honeybadger CSP and frontend noise, and fix what it was hiding

sethherr:
Cuts Honeybadger noise and fixes the real breakage it was hiding. The [CSP project](https://app.honeybadger.io/projects/138070) logs ~30k distinct faults a week, which is why these sat unnoticed for months.

- **CSP reports are normalized and filtered before forwarding.** `blocked-uri` drops its query string (Honeybadger fingerprints on the whole uri, so a rotating ad `label` minted a new fault per request); a cross-origin block our own policy would have permitted is dropped, since we can't act on it either way; and a `font-src` violation from a third-party host is dropped, because every font we load is on our origin or in `font_src` — `images.simplycodes.com` alone was ~1,100 notices in 7 days from an extension injecting page-level `<link>`s. The browser's user agent now rides along as context, since Honeybadger otherwise records our forwarder's.

- **A search results fetch that fails outright now says so.** A dropped connection rejects the frame's eager fetch instead of answering with a status, so the 429 path never saw it and the spinner ran forever. Covers all five pages that render the frame component.

- **Two frontend noise fixes.** The bike-show scroll handler unbound itself *after* calling `loadPhotos`, so a raise left it bound and re-raised on every later scroll — two Android WebViews turned one bad page load into ~250 notices. And fetches abandoned by navigating away no longer report (~230/week); that filter is scoped to a `pagehide` flag, because the phrasings browsers use for an abandoned fetch are the same ones they use for a genuine network failure — which is exactly what the retry above exists for.

- **The `/donate` `connect-src` violation is still open**, now tracked in #4039. It's a real fetch to `/payments` following its cross-origin redirect to Stripe, but it isn't Turbo — `submissionIsNavigatable` is false for all three checkout forms with or without `data-turbo="false"`, and submitting records no fetch at all — so those opt-outs come out as dead code. Forwarding the browser's user agent should identify the actual caller.

`CspReport` owns the classification, and reads the global policy to decide what our own CSP would have permitted — no controller narrows it, so a per-controller policy can only make that answer too strict, never too permissive.


sethherr:
## Screenshots

The new search-results fetch-failure state — the spinner is dropped and replaced with the error and a retry, instead of spinning forever. Reached by rejecting the results frame's eager fetch with `TypeError: Failed to fetch`, the way Chrome does when a connection drops.

Search results page showing a red alert reading &quot;Couldn't load these results — check your connection and try again.&quot; with a Try again button

Clicking **Try again** reloads the frame and the results render normally.

PR #4045, opened 2026-08-06
Grant org roles from a feature, not from passwordless sign in

eld120:
Claiming a domain for passwordless sign in also silently made every account on it a member — one feature deciding two unrelated things. The role now comes from its own organization feature, `user_role_for_user_email_domain` ("Automatic User Role"), and passwordless goes back to only signing people in.

- **`CallbackJobs::AfterUserCreateJob` grants the role off the new feature**, via `Organization.user_role_email_matching`. An org keeps today's behavior by enabling both.
- **`SessionsController#create_magic_link` and `Saml::AssertionProcessor#provision_user` no longer create org roles at all** — both were calling `OrganizationRole.create_passwordless` purely for the user it produced as a side effect, which is `UserServices::PasswordlessCreator`'s job. That method is down to one caller and is now named `create_for_user_email_domain`.
- **An auto-granted role sends no invitation email.** Nobody invited these users, and the old suppression (`created_by_magic_link`) no longer applies now that the role comes from confirmation rather than from the request — without this they'd get "Join {Org}" on top of the sign in link.
- **Not included: the backfill.** An org relying on `passwordless_users` for auto-roles loses them on deploy unless the new slug is added to it first.

Seeded as skipped for Brakebills, alongside the other two features that change how auth works.

### DIFF
--- a/app/models/organization.rb
+++ b/app/models/organization.rb
@@ -253,7 +253,17 @@ class Organization < ApplicationRecord
       domain = email_domain(str)
       return nil if domain.blank?
 
-      permitted_domain_signin("passwordless_users").find_by(user_email_domain: domain)
+      permitted_domain_orgs("passwordless_users").find_by(user_email_domain: domain)
+    end
+
+    # The org that grants a role to anyone signing up on its domain, however they authenticate.
+    # Separate from the sign-in features on purpose: claiming a domain for login says nothing
+    # about whether those accounts should also become members.
+    def user_role_email_matching(str)
+      domain = email_domain(str)
+      return nil if domain.blank?
+
+      permitted_domain_orgs("user_role_for_user_email_domain").find_by(user_email_domain: domain)
     end
 
     # The org that forces SSO for an email's domain: feature enabled + a live IdP config.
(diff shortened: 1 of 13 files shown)


app/models/organization.rb
```diff
@@ -250,10 +250,14 @@ def with_any_enabled_feature_slugs(slugs)
...
-      domain = email_domain(str)
-      return nil if domain.blank?
+      feature_email_matching("passwordless_users", str)
+    end
 
-      permitted_domain_signin("passwordless_users").find_by(user_email_domain: domain)
+    # The org that grants a role to anyone signing up on its domain, however they authenticate.
+    # Separate from the sign-in features on purpose: claiming a domain for login says nothing
+    # about whether those accounts should also become members.
+    def user_role_email_matching(str)
```
sethherr: nit: This only has one call site, I'd prefer to inline the logic there than add this method to `Organization` (which is already too big of a class).

app/models/organization.rb
```diff
@@ -276,7 +280,14 @@ def example
 
     private
 
-    def permitted_domain_signin(feature_slug)
+    def feature_email_matching(feature_slug, str)
```
sethherr: With `user_role_email_matching` moved out of the model, this should be moved back into its one call site in `passwordless_email_matching`

app/services/saml/assertion_processor.rb
```diff
@@ -61,11 +61,11 @@ def asserted_email(response, saml_configuration)
 
     # Only mint an account when the email's domain is the one this org claims for SSO, so an
     # assertion can never create an account on a domain that routes logins somewhere else.
+    # Provisioning grants no membership - that is user_role_for_user_email_domain's job.
```
sethherr: nit: membership > user_role

Sorry - but not referring to the user role as a membership keeps it distinct from our membership model

sethherr (approved):

PR #4049, opened 2026-08-07
Add a target param to the organization QR endpoint

sethherr:
Renames the `shop_display_qr` endpoint to `qr` and lets the caller pick what the code points at, so one endpoint covers the shop printout and the links a shop hands a rider.

- `target=shop_display` keeps the existing embed URL. `target=landing` points at `/{slug}` — only a routed page for organizations in `ORGANIZATIONS_WITH_LANDING_PAGES`, but the right target regardless. Anything else, blank or unrecognized, falls through to `/register/new?organization_id={slug}`.
- `/organizations/{slug}/qr` sends the png directly instead of redirecting to `.png` first, and a missing organization returns rather than falling into a double render.
- Adds a heading on `/o/{slug}/bikes/new` pointing at the new registration flow.


sethherr:
## Screenshots

### /o/brakebills/bikes/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4050, opened 2026-08-07
Add PKCE to the OAuth authorization code flow

sethherr:
`oauth_access_grants` had no `code_challenge` columns, so Doorkeeper's `pkce_supported?` was false and PKCE was silently inert — and since `oauth_applications.confidential` is `DEFAULT false` and exposed nowhere, every app is a public client whose code exchange nothing was protecting.

- **Adds the two grant columns and restricts `code_challenge_method` to `S256`.** The columns are the whole switch — `pkce_supported?` is `column_names.include?("code_challenge")`. `plain` sends the verifier in the clear, and nothing uses PKCE today, so there's no compatibility cost to excluding it. PKCE stays optional: `force_pkce` would reject every existing authorization-code client at once, since they're all public.
- **Exposes `confidential` on the application form**, so an app that *can* hold a secret can require one. The help text says "when requesting tokens" rather than naming the code exchange, and a warning sits under the checkbox: turning it on makes Doorkeeper reject *every* token request without a `client_secret`, including the refresh-token call the API docs document without one.
- **Drops `"password"` from `grant_flows`.** `resource_owner_from_credentials` was never defined, and its default logs a warning and returns `nil` rather than raising — so valid credentials came back `invalid_grant`, which reads as "your password is wrong" instead of "this flow was never wired up". Now `unsupported_grant_type`. Nothing in the app, docs, or swagger definitions advertised it.
- **Fixes `.markdown` templates truncating at their first ERB tag** (first commit, separable). `MarkdownHandler` bound `output` to the first of ERB's several compiled statements and dropped the rest, with no exception and no log line. This was already live — the refresh-token example in the API docs had been rendering as a bare `POST` with no URL, and the new PKCE section made it swallow three more sections.

The application form moves from HAML to ERB and the UI components along the way: `UI::Forms::Group` for both fields, `UI::Alerts::ObjectError` for errors, `UI::Button`/`ButtonLink` for the actions. That retires the Bootstrap `form-horizontal` grid it was built on, so per-field `doorkeeper_errors_for` becomes one error list at the top.


sethherr:
## Screenshots

The OAuth application form (`/oauth/applications/:id/edit`), shown against the seeded `Localhost Dev iOS` app. `Confidential` is off, which is correct for an iOS app — it can't keep a `client_secret`, so it stays public and should use PKCE instead.

The form moved to ERB and the UI components in this PR, so the field styling, error display and buttons all change with it.

### /oauth/applications/:id/edit

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4055, opened 2026-08-07
Add four footer links; fix news list-marker leak and data repo URLs

sethherr:
The footer was missing the navbar's main destinations and had no link to [status.bikeindex.org](https://status.bikeindex.org).

- **Search**, **Marketplace** and **Register a bike** join the Bike Index column, **Status page** goes in Resources. Register points at `/register/new` rather than the navbar's `choose_registration_path`.
- `#news_show ul, ol` in `news/show.scss` was unscoped. `page_id` is the body id and `InfoController#show`/`#membership` set it to `news_show`, so that id selector outranked `list-unstyled` on the header, footer and content-menu navs — every list in the site chrome got markers and an indent on those pages. Now scoped to `.primary-content-block`.
- The manufacturers index and `Spreadsheets::ImporterJob` both pointed at `bikeindex/resources` (the index at a path that never held the CSV); they now use `bikeindex/bike_data/data`. The manufacturers fragment-cache key is bumped so the new link isn't held back by the existing cache.

The four `translation.*.yml` files carry the new keys inline. Those files are sync-generated, so the next translation.io sync overwrites them — without them the footer component spec, which renders in `:nl` under `raise_on_missing_translations`, fails.


sethherr:
## Screenshots

### /rails/view_components/page_block/footer/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4056, opened 2026-08-08
Confirm additional emails through a POST, not the emailed GET

sethherr:
`UserEmailsController#confirm` was the last emailed-token flow that changed state on a bare GET — the link in `customer_mailer/additional_email_confirmation` hit `get "confirm"`, which called `@user_email.confirm` and merged the address in. Every other emailed-token flow already lands on an interstitial that posts.

- **`confirm` is now `match via: %i[get post]`**, mirroring `users#confirm` — the GET renders `user_emails/confirm`, which auto-posts the token back. The route helper is unchanged, so links already in inboxes still resolve. `ensure_user_email_ownership` means only the signed-in owner could ever have spent the token, so this covers a prefetch in their own browser rather than a scanner.
- **`users/unsubscribe` drops its hand-rolled form and inline `document.forms[...].submit()`** for the same component and the `auto-submit` Stimulus controller, so both auto-posting pages share one mechanism. `Sessions::SignInInterstitial::Component` takes `heading:` and `submit_text:` to make that work, defaulting to the sign-in copy; its body sentence is now subject-neutral and owned by the component, retiring the byte-identical `users.unsubscribe.if_not_click_button`.
- **Resetting a password told you "Logged in!" instead of that your password was reset.** `update_password_with_reset_token` set `password_reset_successfully`, then `sign_in_and_redirect` assigned `flash[:success]` again, so that translation had never rendered. The generic message is now a fallback (`||=`) rather than an assignment — no other caller sets a success flash before signing in, and leaving the caller's assignment in place keeps it on the paths that return before `set_sign_in_flash`, notably a stored `return_to`.
- **Nothing held the GET/POST pattern in place before**, which is how the GET-only route survived. `emailed_token_routes_request_spec` pins all five flows to a GET that renders and a POST that acts, and fails when a template that auto-posts isn't one of them. The two flows whose interstitial wasn't checked now assert its `form[action]` and token field, matching what `users#confirm` and `magic_link` already did. A component system spec drives the auto-POST in a browser — request specs post directly, so the Stimulus controller had no coverage at all.
- **`passwordless_signup_spec` becomes `signup_spec`** and carries both ways back in: setting a password from the nudge then signing in with it, and signing in by magic link then confirming an additional email, which walks the emailed link through the new interstitial. Driving the real UI is what surfaced the flash bug above, plus two labels that were never associated with their input (`for="user_Password"` against a `user_password` field, and the additional-email template with no `for` at all).

The component's name and its `components.sessions.sign_in_interstitial` locale scope now cover four flows, only two of which sign anyone in. Renaming orphans 12 translated strings across es/it/nl/nb, so it's left for its own commit alongside a translation.io sync.


sethherr:
## Screenshots

### /rails/view_components/sessions/sign_in_interstitial/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The body sentence is the only visual change — it moves off "If you are not…" so it reads correctly under the unsubscribe and email-confirmation headings too.

`/users/:id/unsubscribe` and `/user_emails/:id/confirm` aren't captured: both auto-post and redirect within ~100ms, on this branch and on `main`, so there's no stable frame on either side.

PR #4057, opened 2026-08-08
Tighten UI::Alerts::Base icon spacing, error icon and header color

sethherr:
Small cleanups to `UI::Alerts::Base`, plus a preview reorganization.

- **The icon moves closer to the text and centers on its line** — the gutter halves (`tw:mr-4` → `tw:mr-2`), and the floated icon now takes the line-height of the line it shares rather than being nudged onto that line's baseline. It still occupies only the first line, so second lines wrap back under it.
- **The `:error` kind gets an exclamation-triangle default icon** instead of the info circle; the other kinds are unchanged. New `icons/exclamation-triangle.svg` is drawn to match `info.svg` (20×20, `stroke-width: 1.5`).
- **New `default_header_color:` option** renders the header in `twtext-color!` rather than the kind's color. The banned-user alert on `users#show` uses it, and both alerts on that page move to `tw:mb-8` so they sit further off the profile below.
- **Previews now walk every kind twice** — plain, then with a header — so the header spacing is comparable across colors, and the dismissable group covers all five kinds instead of three. The old standalone link and custom-icon scenarios fold into the success and purple pairs.

Three `MARKUP_DIGEST` constants change because the alert markup is inside those components' fragment caches — `bin/update_component_digests` regenerated them.


sethherr:
## Screenshots

### /lookbook/preview/ui/alerts/base/kind_variants

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /lookbook/preview/ui/alerts/base/dismissable_variants

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4062, opened 2026-08-08
Convert the admin navbar into Admin::Navbar

sethherr:
The admin navbar was inline in `layouts/admin`, with its page picker a selectize `<select>` navigated by a jQuery handler buried in the vendored `application_standalone.js` bundle. It's now `Admin::Navbar::Component`, and the layout renders it in one line.

- **The nav link tables move into the component.** `admin_nav_select_links`, `dev_nav_select_links`, `admin_nav_select_link_active` and `admin_nav_display_view_all` all leave `AdminHelper`. The tail of `admin_nav_display_view_all` stays behind as `admin_search_filtered?`, since `admin/bikes/index` also uses it to decide whether to show sitewide counts.
- **The picker is `UI::Forms::Combobox`**, navigated by a Stimulus controller listening for `hw-combobox:selection`. The combobox also fires that event when closed untouched, with a blank value — which is what the old `!= window.initialValue` check was doing. Options carry explicit ids; without one the gem gives each a uuid, ~1.5 KB gzipped of incompressible hex per page.
- **The Mailers menu is `UI::Dropdown`** (own commit), so it toggles through `ui--dropdown` rather than Bootstrap's `data-toggle`, and its trigger is a real button rather than an href-less anchor.
- **The admin layout renders the review app topbar**, which only `layouts/application` had — so admin pages on a review app or dev looked like production. Its links needed `tw:text-inherit` to stay black against admin.css's `a { color: #007bff }`.
- **Dead markup dropped**: the `more`, `nav-link users`, `nav-link bikes` and `admin-nav-dropdown` classes and a `role="button"` on real links, none of which anything selected on; plus the `.admin-navbar` `fancy-select`/`dropdown-toggle` rules, now that the picker's width is `tw:min-w-75` / `tw:max-sm:min-w-60`.

The five shortcut links keep their own literal list rather than deriving from the shared table — their labels are shorter ("Stolen", not "Stolen Bikes") and their order is deliberate rather than alphabetical, so only the paths overlap.

The view-all cases move from `spec/helpers` to the component spec, where `with_request_url` drives them off a real path instead of a stubbed `request` double.

`admin_search_filtered?` overlaps `Binxtils::SortableHelper#sortable_search_params?`, already this app's idiom in four views. Swapping to it would delete the helper and its spec, but the gem excludes `per_page` where the moved code doesn't — so on `?per_page=100` the "All" link would disappear and the counts would appear. Probably the better behavior, but it's a behavior change, so it's left for its own PR.


sethherr:
## Screenshots

Captured on the local dev server, so the topbar renders in its purple development flavor. `/admin/maintenance` stands in for "an admin page" — it's short enough that the navbar is legible.

### /admin/maintenance

The topbar is new on admin, and its links are black rather than Bootstrap's blue.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes?period=week

The "All Bikes" link and the `Viewing Bikes` placeholder. Topbar hidden here so the comparison is only the navbar.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/maintenance — Mailers dropdown open

Bootstrap's caret becomes `UI::Dropdown`'s chevron, and the menu gets its rounded/shadowed surface. The menu takes `menu_class: "tw:w-max"` so it sizes to its entries rather than shrink-to-fitting against the button. Desktop only — the shortcuts and this dropdown are `d-none d-lg-flex`, so neither renders at mobile width, on either branch.

| Desktop | Mobile |
| --- | --- |
|  | _n/a — hidden below `lg`_ |
| main 👆 | this branch 👇 |
|  | _n/a — hidden below `lg`_ |

PR #4063, opened 2026-08-08
Add a helper_text slot to UI::Forms::Group

sethherr:
Helper text under a form field was hand-rolled at every call site — a sibling `<p>` carrying its own copy of the type classes, plus a `wrapper_class` override to tighten the gap the Group's default margin left behind.

- **`renders_one :helper_text`** renders that paragraph under the field, with the input pointing at it via `aria-describedby`; a content block renders its own field, which the Group can't reach, so it takes `helper_text_id` and describes that field itself.
- **Migrates the three existing instances** — the admin superuser ability form's three fields, and the registration photo's privacy note, which is now announced to screen readers.
- The fallback to `UI::Forms::Input` branches on `content.present?` rather than `content?`, which ViewComponent returns true for *any* block, including one that only sets a slot.


sethherr:
## Screenshots

### /admin/superuser_abilities/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /register?step=2

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/group/component/helper_text

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4064, opened 2026-08-08
Top 10 segments for Strava activities; search and display

sethherr:
Strava reports a "top 10" whenever an effort places on a segment's all-time leaderboard. Those placements are now stored, searchable, and shown on the activity card next to PRs.

- **`top_10_ranks` holds each qualifying `kom_rank` ascending** — a KOM plus a 2nd place is `[1, 2]` — collected from `segment_efforts` on the detail response, the same fetch that already builds `segment_locations`.
- **It's a column rather than a `strava_data` key.** `strava_data` is replaced wholesale on every activity-list sync, and summaries carry no `segment_efforts`, so a key there would be cleared by the next re-list. `pr_count` can live in the bucket because summaries include it; this can't. `spec/models/strava_activity_spec.rb` covers the re-list case.
- **The SPA gets a "Top 10" checkbox** beside the Kudos range (URL param `top10=1`), matching how `noEquipment` is wired, plus a `🏅 top 10: #1, #2` badge after the PR badge that lists the first five placements and counts any remainder.

Two notes for the reviewer:

Rows enriched before this ships stay `NULL` until they're fetched again — we don't store `segment_efforts`, so there's nothing to backfill from. `RE_ENRICH_AFTER` is an hour, so ordinary use fills it in.

A local-legend filter was scoped alongside this and dropped: it isn't in the activity payload at all, only on `GET /segments/{id}`, so it would cost one extra API call per segment per activity.


sethherr:
## Screenshots

Storybook (`Components/SearchFilters`, `Components/ActivityCard`) rather than `/strava_search` — the page only renders for a user with a connected Strava integration, and there's no seed data for that, so these use the repo's mock athlete.

### Components/SearchFilters — Default

The new checkbox, to the right of the Kudos range.

| Desktop | Mobile |
| --- | --- |
|  |  |

### Components/ActivityCard — Nordic Ski

`top_10_ranks: [2, 7]` — badge sits after the PR badge, in the row's text color; only PRs stay orange.

| Desktop | Mobile |
| --- | --- |
|  |  |

### Components/ActivityCard — Long Distance Ride

`top_10_ranks: [1, 1, 2, 3, 4, 6, 8, 9, 10]` — the five-rank cap with the remainder spelled out.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4066, opened 2026-08-08
Add UI::ButtonGroup, a link/button take on the radio chips

sethherr:
`UI::Forms::RadioButtonGroup` renders a row of purple-outline chips backed by sr-only radios. `UI::ButtonGroup` renders that same row for the cases that aren't a form field — each chip either navigates (an `<a>`) or acts (a `<button>`).

- **`entries: [{label:, href:, active:, disabled:}, …]`**, mirroring the radio group's API. No `href` renders a `<button type="button">`; anything else in the hash (`data:`, `title:`, `target:`) becomes an attribute. A disabled entry is a `<button disabled>` even when it has an `href`, since an `<a>` takes no disabled attribute.
- **The chip look now has one home.** `RESTING_CHIP_CLASSES` and `layout_classes` move here, and the radio group builds its `<label>` chips from them — its rendered classes are unchanged, which is why the two `MARKUP_DIGEST`s move.
- The active chip carries `data-active` (driving the `is-active` variant, same as `UI::Button`) plus `aria-current` on links / `aria-pressed` on buttons, so the state isn't color-only.

No call sites yet — the existing Bootstrap `btn-group` filter rows (`organized/stickers`, `bikes_edit` frame size) are the obvious candidates for a follow-up.

### DIFF
--- a/app/components/org/search_results/bikes_table/component.rb
+++ b/app/components/org/search_results/bikes_table/component.rb
@@ -9,7 +9,7 @@ module Org
       # registrations on the show page). Pass render_sortable to enable sort links.
       class Component < ApplicationComponent
         # Digest of the markup inside the row cache — the cached_markup_digest spec keeps it current
-        MARKUP_DIGEST = "3e3766884307"
+        MARKUP_DIGEST = "bafe963e2a87"
 
         delegate :additional_registration_fields, :column_renames, to: :settings_component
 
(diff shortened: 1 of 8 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/button_group/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/button_group/component/with_disabled

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/forms/radio_button_group/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4067, opened 2026-08-08
Make secondary the purple outline and error the danger outline

sethherr:
Two colors take over the outline styles that sat beside them, and the originals are gone: `:secondary` becomes what `:purple_outline` was, and `:error` becomes what `:danger_outline` was. Secondary is the default color, so this is what most buttons in the app now look like.

- **Every default button changes**, since `UI::Button` and `UI::Dropdown`'s trigger both default to `:secondary`. The most repeated instance is the period selector in `Admin::IndexSkeleton` (~68 admin index views) and the org index pages, where the *selected* period goes from gray-200 to filled purple. The five `:error` call sites — revoke, delete page, start over, undo, "I'm 100% sure" — go from filled red to a soft outline, and "Mark stolen" (the one `:danger_outline` site) now asks for `:error`, which is what it always meant.
- **What sits beside a button follows it.** `UI::Dropdown`'s active entry and `UI::Pagination`'s current page hardcoded the grays secondary used to carry, and `.twinput` was a shade darker with a blue focus ring — all three take the new palette now. The dropdown and paginator specs compare against `UI::Button`'s active fill rather than a literal, so the next retune can't leave them behind.
- **Disabled is honest now.** `DISABLED_CLASSES` no longer drops pointer events, so a disabled button shows `cursor-not-allowed` rather than the arrow the browser takes from underneath a `pointer-events: none` element. `UI::ButtonLink` gained a real disabled state: passing `disabled:` used to add an attribute an `<a>` ignores, leaving the link followable at full opacity, so it now drops the href and carries `aria-disabled` + `tabindex="-1"`.
- **Each state answers differently now.** `primary` and `purple` pressed to the same fill they hovered to, so pressing one looked like nothing happened — their `:active` fills are a step darker. `link` color had no focus ring at all, since `build_classes` skipped every extra for it; focus and disabled reach it now, so a keyboard can see where it is and a disabled one dims.
- **What holds all of that up is cascade order**, and it's worth a look: every hover is guarded `not-disabled:not-aria-disabled:`, `.twlink`'s hover/pressed rules are guarded too (link color's came from there rather than the utilities), and the `is-active` variant carries the same two `:not()`s — which keeps a disabled element from looking active *and* ties the specificity it has to beat. Without that last part a pressed button rendered its hover color.

Previews pair every color with its active and disabled state on both components, and ButtonLink's States and `button_to` groups became one Extra types group. The legacy `.select2-*` focus styles still ring blue — they're for markup the search UI has moved off, so they're left alone.


sethherr:
## Screenshots

Component previews rather than real pages — this workspace's dev database is empty, and these are component-level color changes the previews show directly. Happy to seed and capture a bike show page and an admin index if that's more useful.

### /rails/view_components/ui/button/component/secondary

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/secondary_active

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/error

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button/component/primary_active and purple_active

The pressed and active fills, a step darker so pressing no longer repeats the hover.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/button_link/component/secondary

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/dropdown/component/custom_button

Captured with the menu open, since the active entry is what changed.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/pagination/component/middle_page

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/input/component/text_field

The `.twinput` border change is a single shade (gray-300 → gray-200), so the pair looks nearly identical — included because it's the one CSS change that reaches every field in the app.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Hover, pressed, focus and cursor changes don't show in a static capture, so `component_system_spec` asserts them in a browser instead, for every color and on both components: a disabled one holds its colors through hover and press, an enabled one looks different resting, hovered and pressed, and an active one ignores hover but rings identically for focus and press.

PR #4068, opened 2026-08-08
Add the register flow's report step for stolen and found registrations

sethherr:
A stolen or found registration now has a step of its own for what it's reporting, so the stolen or impound record arrives with the bike rather than being chased down afterwards.

- **`?step=report` asks everything `/bikes/:id/edit/theft_details` does about a theft** — when and where, what happened, locking and lock defeat, estimated value, police report, and who can see the phone number step 2 already collected — and the impound record's own fields for a vehicle that was found. It writes them onto the `BParam` as `stolen_record` / `impound_record` attrs, which `BikeServices::Builder` already knows how to build a record from.
- **Where it lands depends on the confirmation email**: right after step 2 when the registration has a creator, and after the acknowledgment pages when it doesn't — the emailed link is what proves the address the report belongs to, and it's clicked once those pages are done. Either way no bike is created until the report is in.
- **A theft has to say when and where.** The date opens blank rather than on the current time, since a field that arrives pre-answered is one nobody corrects and "now" is the least likely answer for a theft; the location is required alongside it. Both are checked server side, and a failed step re-renders with what was entered.
- **The flow is an ordered list of steps now.** `Register.steps` is what places the report, each step opens the next through `step_completed?`, and the progress bar, the back links and step 2's own submit label all read their answer out of that one list — rather than the acknowledgment pages counting off a fixed offset. It's built once a request, since placing the report asks whether the registration has a creator yet.
- **The statuses that report something are told apart from the rest throughout**: `/register/new?status=stolen` takes the shorthand every other entry point does, the confirmation email says what's waiting rather than asking for an email confirmation, step 2 asks for a phone rather than offering one, and the completion card doesn't tell someone who just reported a theft that we'll watch in case it's ever stolen.

Two notes for the reviewer:

`status_abandoned` and `unregistered_parking_notification` don't get the step, since `BikeServices::Builder` builds a record for neither — step 2's "Found/Abandoned" option is `status_impounded`, which does.

The status is picked in step 2's own form, so three things it decides are rechecked in the browser by `register--status-fields`: which fields show, whether the phone is required, and whether the button finishes the registration. Each carries a `status → copy` map so the server render and the client recheck read the same answer. `UI::Forms::Group` grows `required_toggleable` for the phone — a label can't be rebuilt from JS, so both markers render and the controller flips which is hidden.


sethherr:
## Screenshots

### /register?step=report — stolen

When and where are required and open blank; everything else is optional.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /register?step=report — found

| Desktop | Mobile |
| --- | --- |
|  |  |

### /register?step=2 — stolen

Three changes: the progress bar gains a segment for the report, the button says `Next` rather than claiming to complete the registration, and the phone is required and captioned rather than badged optional.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /register?step=finished — after a theft report

The generic card told someone who had just reported a theft that we'd keep watch "if it's ever reported stolen", and gave them nothing to do next. It heads with what happened now, and carries the same checklist the theft details page ends on — with what the flow already did ticked off.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4070, opened 2026-08-09
Permit region_string in BikeServices::StolenRecordUpdator

sethherr:
`BikeServices::StolenRecordUpdator` never permitted `region_string`, so a region with no `State` record behind it — most addresses outside the US — was dropped on the way into the stolen record, even though the column exists and `Bike#build_new_stolen_record` slices for it.

- **`region_string` joins `OLD_ATTR_ACCESSIBLE` and `permitted_attributes`**, the two lists the updator's write paths slice against. A region that does match a state still resolves to `region_record_id` and clears itself, which is `Geocodeable#assign_region_record` doing its usual job.
- **The start over modal drops the vehicle type from its copy** — "This begins a new registration — nothing you've entered here carries over."

### DIFF
--- a/app/components/register/step1/component.en.yml
+++ b/app/components/register/step1/component.en.yml
@@ -16,9 +16,8 @@ en:
         register_your_bike_with_org_html: Register your %{cycle_type} with %{org_name}!
         start_over: Start over
         start_over_certain: Start over?
-        start_over_discards: >-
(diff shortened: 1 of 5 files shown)


sethherr:
## Screenshots

### /rails/view_components/register/step1/component/returned_from_step_2

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4071, opened 2026-08-09
Convert the search kind and result view selectors to RadioButtonGroup

sethherr:
`Search::KindSelectFields` built its segmented control by hand — a `<ul>` of `Search::KindOption` radios plus ~20 lines of conditional border and rounding classes to fake the joined-chip look. Both it and `Search::ResultViewSelect` now render `UI::Forms::RadioButtonGroup`, the first call sites for the chips added in #4066.

- **`for_sale` on registration search stays a link**, since it navigates to marketplace search rather than scoping in place. It renders as a one-entry `UI::ButtonGroup` beside the radios, so it still looks identical to its neighbours. The count spans ride along inside the chip labels, so the `data-count-target` JS is untouched.
- **`Search::KindOption` had no other callers**, so it goes, along with `li_classes`, `final_radio_option`, `opt_button_url` and `opt_selected?`. On mobile the chips now wrap rather than stacking full-width.
- **A radio group entry's extra keys become attributes on its chip label**, the way `UI::ButtonGroup` entries already do — that's what carries the result-view tooltips.

Also fixes a flake in the `EverythingCombobox` specs (the group is already tagged `:flaky`): an earlier example's color stayed in the autocomplete Redis index after its row rolled back, so the combobox matched the stale one ahead of the current example's.


sethherr:
## Screenshots

### /search/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

sethherr:
Not making this change right now

PR #4072, opened 2026-08-09
Save which of our addresses a bug report was sent to

sethherr:
Everything Postmark forwards in becomes a `BugReport`, but the report only recorded the sender — not whether it arrived at contact@, support@ or bugs@.

- **`receiver` is set from the inbound mail** by `BugReport.receiver_from_mail`, preferring an address of ours: `X-Original-To` (the envelope recipient Postmark's ingress prepends, the only source when we're bcc'd), then `To`, then `Cc`. With none of ours among the recipients it falls back to the first one, so a forward still shows where it came from.
- **The admin index gets a sortable column, an "All receivers" filter dropdown and a 🔎 link per row.** The column shows just the mailbox name — `contact`, not `contact@bikeindex.org` — since `BugReport.display_receiver` only drops our own domain, so an address anywhere else still reads in full. The show page gets a row, and `receiver` joins the index JSON.
- **`Backfills::BugReportReceiverJob` covers the reports that predate the column** — only the ones whose inbound email is still around, since Action Mailbox incinerates them after 30 days. Run it from the console like the other backfills.
- **The show page picks tags with the combobox**, free text and multiselect, replacing the comma-separated text field. `tag_chips` renders each chip verbatim: the combobox removes a selection by the text as typed, so a normalized chip strands the tag it came from — dismissing it would leave the tag in the field and save it anyway. `tags=` still normalizes on save.


sethherr:
## Screenshots

### /admin/bug_reports

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bug_reports/1

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4073, opened 2026-08-09
Stop email scanners spending magic-link tokens; fix the CSRF dead end

sethherr:
From a support report that every sign-in "gives me an error code". Production logs for that account show 17 of ~20 POSTs failing CSRF, and across two days 122 of 380 magic-link sign-ins failing on an invalid token — 64 of them on the user's *first* click.

- **The interstitial auto-submitted, which is exactly what it exists to prevent.** Emailed links land on a form that POSTs rather than spending the token on the GET — but the Stimulus `auto-submit` fired for anything running JS, and email security scanners run headless browsers. Logs show AWS and Zscaler IPs doing `GET /session/magic_link` → POST → `/my_account` and nothing else ever, 13 of them followed within 15 minutes by a failed sign-in from a different IP. Every emailed-link form now waits for a click, and `auto_submit` is gone rather than left unused.
- **`ApplicationController#handle_unverified_request` asked for `:csrf_invalid`, a key in no locale file.** `raise_on_missing_translations` is on in dev/test and off in production, so the flash shipped as the literal `Translation missing: en.controllers.application.handle_unverified_request.csrf_invalid` — the "error code" in the report. It now shares the `invalid_authenticity_token` copy #3875 added, from one place rather than a scope literal per controller, and the `ControllerHelpers#translation` doc comment that taught the dead key is fixed too.
- **A CSRF failure in the sign-in flow redirected to the homepage**, discarding the form and the token. `SessionsController` hands back the step they were on with the email or the magic-link token preserved — the same thing `UsersController` already did for signup.

Unsubscribe was the interesting call: its signed id lasts a year and re-posting is idempotent, so nothing is *spent* by a scanner following it. But the reader still ends up unsubscribed without asking, and never finds out — so it waits for a click like the rest.

`emailed_token_routes_request_spec` now asserts the flat rule, scoped to the templates behind emailed links (search boxes and comboboxes call `requestSubmit` for good reasons). The feedbacks spec covers `ApplicationController`'s own handler, which nothing exercised before — that's why the missing key survived years of green CI.

Two follow-ups this doesn't take on: `spec/i18n_spec.rb` still can't catch a dead key reached through `translation(:key, scope: [...])`, because `config/i18n-tasks.yml` ignores `controllers.*`/`components.*` and i18n-tasks only scans `t`/`translate` — a static check over the ~36 explicit-`scope:` call sites would close it. And `users.unsubscribe.confirm_unsubscribe` replaces a key that had translations in all four locales, so it wants a translation.io sync.

Not fixed here: why that user's browser fails CSRF at all. She's the worst of 18 IPs over two days and the rest are mostly mobile carriers, so it reads as client-side cookie loss — but the app no longer makes it unreadable or unrecoverable.


sethherr:
## Screenshots

### /rails/view_components/sessions/sign_in_interstitial/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/register/step_confirm/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/sessions/sign_in_interstitial/component/unsubscribing

New preview, so there's no `main` version to compare against — on `main` this page auto-submitted, so a scanner following the emailed unsubscribe link unsubscribed the reader.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4074, opened 2026-08-09
Give the registration sequence template a draft lifecycle

sethherr:
The template that organizations clone their registration sequence from was a permanently editable singleton — the one sequence in the system that didn't follow its own rules. It now drafts, activates and archives exactly like an organization's: one live template, frozen once activated, and changing it means drafting a replacement that supersedes it on activation.

- **`nil` is the template's owner throughout**, so `active_for` / `draft_for` / `make_active!` / `discard_draft!` work on it with no special-casing. The `editable` scope collapses into `draft`, and `"template"` stops being a status — a template is Draft, Current or Previous like anything else. It stays an admin index *filter*, since it's which sequence rather than which stage, so that dropdown no longer partitions the table.
- **Schema**: the single-template unique index becomes one-draft and one-live template guards, and the existing template row is activated so organizations keep cloning something live. The per-organization indexes are untouched — keyed on `COALESCE` they'd have stopped serving `active_for`, which runs on every motorized register step.
- **Admin gains `create` and `destroy`, and an `activate` param on `update`** (the way the pages controller's `update` takes a `position` to reorder), generic across owners rather than template-only — which also fills the gap the organization screens already promise ("contact Bike Index when you're ready to make it live"). Activate sits with the View | Preview | Edit switcher, and Create draft takes its place where the frozen Edit chip is inert; the index links the template so it isn't hunted for in the table.
- **Discard draft moved to the foot of the sequence screens**, in its own component — it's the one action that doesn't keep the draft, and it shares its warning with the organization screen rather than repeating it.

`POST /admin/registration_sequences` takes a blank `organization_id` to mean the template, so it uses `friendly_find!` — a typo'd slug 404s instead of quietly opening the draft of the template every organization clones.


sethherr:
## Screenshots

No `main` comparison: this branch adds a migration, so `main`'s code against the migrated schema isn't a fair before. The template screens also don't exist on `main` in this form — it has one permanently-editable template, no draft above it.

### /admin/registration_sequences

The template is linked in the nav rather than hunted for in the table — its draft when there is one, else the live one.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /admin/registration_sequences/:id — the live template

Frozen like any live sequence, so Edit is inert and explains itself; Create draft is the way in, right-aligned on the switcher's line.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /admin/registration_sequences/:id/edit — the template's draft

Reached by pressing Create draft above, which cloned all three pages off the live template. Activate sits with the switcher; Discard draft is at the foot of the page, away from the actions that keep the draft.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4076, opened 2026-08-09
Say why a magic link failed, not just that it did

sethherr:
Follow-up to #4073. That PR stopped email scanners spending magic-link tokens, but scanners were only one of the reasons sign-in was failing — and every reason showed the same "Unable to authenticate that token" page.

Across the two days of logs behind #4073, 122 of 380 magic-link sign-ins failed: **33 were re-clicks of a link that had already worked**, 11 were superseded by a newer email, and the rest had timed out. All three read to the user as "this site is broken", which is what the original report said.

- **The reason comes from the token's own timestamp.** A dead token matches no user whatever killed it, so there's no record left to consult — but every token is `<epoch>-<hex>`. Older than the window and it timed out; recent and it was either spent signing in or replaced by a newer link. Unparseable keeps the generic message, so a mangled URL never gets told it "expired".
- **`SecurityTokenizer.recognized_token?`** — `token_time` already decides parseable-or-not and then throws the answer away, floored at `EARLIEST_TOKEN_TIME`. `BParam` carries a comment explaining that convention precisely because it's invisible at the call site; this stops it needing a third one.
- **`User::AUTH_TOKEN_EXPIRY`** replaces the bare `2.hours` in `auth_token_expired?`, and the copy interpolates it, so the window we tell people about can't drift from the one we enforce.

The three messages:

| Reason | Copy |
| --- | --- |
| Expired | That sign in link has expired — Sign in links work for 2 hours after they're sent. |
| Already used / superseded | That sign in link has already been used — You may already be signed in somewhere else. If you asked for more than one email, only the link in the newest one works. |
| Unparseable | Unable to authenticate that token |

"Already used" and "superseded" are deliberately one message: both leave a token matching no user, so the copy has to cover both honestly. Collapsing them for real would mean not spending the token on use — the other follow-up, and the one that removes those 33 failures outright rather than explaining them.

Signed-in users never reach this page; `skip_if_signed_in` already bounces them to their account with "you're already signed in".

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -42,7 +42,7 @@ class SessionsController < ApplicationController
 
   def magic_link
     @token = params[:token]
-    @incorrect_token = params[:incorrect_token].presence
+    @failure = magic_link_failure(params[:incorrect_token]) if params[:incorrect_token].present?
   end
 
   def sign_in_with_magic_link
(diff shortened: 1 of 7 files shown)


sethherr:
## Screenshots

Same page, two different dead tokens. On `main` both render the identical generic message — which is the point. The alert itself also moves from Bootstrap's `.alert-warning` to `UI::Alerts::Base` (`kind: :warning`), hence the icon and the Tailwind palette.

### /session/magic_link?incorrect_token=… (token older than the window)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /session/magic_link?incorrect_token=… (token inside the window)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

A token that can't be parsed at all still gets the generic message, unchanged from `main` — so no screenshot.

PR #4078, opened 2026-08-09
Cut the emailed token window to 10 minutes, index the lookup

sethherr:
Follow-up to #4073 and #4076. Two days of production logs put 99% of magic-link clicks within 5 minutes of the email being sent, and 99.7% within 15 — so a two-hour token was exposure with almost nothing to show for it.

- **`User::AUTH_TOKEN_EXPIRY` drops from 2 hours to 10 minutes**, password reset tokens included since they share it. The window is stated to users via `distance_of_time_in_words`, so the copy tracks the constant instead of carrying its own hardcoded unit — which is how "2 hours" got baked into the string in #4076.
- **`users.magic_link_token` gets an index.** It had none, so every sign-in attempt seq-scanned `users`. Partial on `IS NOT NULL`, which is only the handful with a link outstanding, so it stays tiny.
- **Specs stop hardcoding the window.** Eight sites carried `121.minutes` / `3.hours` / `1.hour`, all meaning "past the expiry"; they now read `(User::AUTH_TOKEN_EXPIRY + 1.minute).ago`.

### DIFF
--- a/app/components/register/step_confirm/component.rb
+++ b/app/components/register/step_confirm/component.rb
@@ -2,9 +2,8 @@
 
 module Register
   module StepConfirm
-    # Where the emailed confirmation link lands. Nothing is confirmed by rendering it -
-    # confirming is single use, and a link scanner runs this page's JS, so the form waits
-    # for a click rather than posting itself
+    # Where the emailed confirmation link lands. Confirming is single use and scanners run the
+    # page's JS, so the form waits for a click rather than submitting on render
     class Component < ApplicationComponent
       def initialize(b_param:, token:)
         @b_param = b_param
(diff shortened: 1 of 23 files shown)


sethherr:
## Screenshots

The only user-visible change is the stated window. It reads through `distance_of_time_in_words(User::AUTH_TOKEN_EXPIRY)` now, so it follows the constant rather than restating a unit.

### /session/magic_link?incorrect_token=… (expired link)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The already-used and unreadable-token states are untouched by this PR.

PR #4079, opened 2026-08-09
Stop bootstrap defaults leaking into alerts

sethherr:
Bootstrap's defaults were reaching into `UI::Alerts::Base::Component`: the header rendered in the legacy heading color instead of the alert's, spacing and line-height came from `.container`, and the last paragraph's margin left a band of empty space at the bottom of the box.

- **The `!` variants were never needed, and one of them never existed.** `revised.scss` wraps legacy in `@layer legacy`, declared before `components`/`utilities`, so plain utilities already beat it. `header_color_classes` built its class with `gsub("00", "00!")` — a string Tailwind's source scanner never sees — so no rule was generated at all and the header fell through to `.header-font`'s color. That's the miscolored header.
- **The header owns its font and case** (`tw:font-sans tw:normal-case` rather than the legacy `uncap`), instead of depending on a legacy counter-rule that only wins because it's `!important` in an earlier layer.
- **`.twalert-body` owns what layers can't fix on their own** — `.container`'s `font-size: 15px; line-height: 2em` reach the alert by inheritance, and its `p, ul, ol, dl` margins come from a layer that outranks preflight.

Paragraphs inside an alert now sit at Tailwind's base margin of 0, so a body that wants a gap between two of them has to ask for one.

### DIFF
--- a/app/assets/tailwind/application.css
+++ b/app/assets/tailwind/application.css
@@ -106,6 +106,12 @@
     @apply tw:text-blue-700 tw:dark:text-blue-300 tw:font-bold tw:underline;
   }
 
+  /* Bootstrap's `.container p, ul, ol, dl` margins are in the legacy layer, which
+   * beats Tailwind's preflight -- restore the base spacing inside an alert */
+  .twalert :is(p, ul, ol, dl) {
+    @apply tw:m-0;
+  }
+
   /* Links inherit the alert's own color, so the underline is what marks them as links */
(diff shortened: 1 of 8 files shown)


sethherr:
## Screenshots

### /session/magic_link?incorrect_token=…

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4083, opened 2026-08-09
Layer the rest of the legacy stylesheets, drop the !s they replace

sethherr:
Follow-up to #4079, which established `@layer legacy` as the reason a plain tailwind utility beats bootstrap without `!important`. Two halves: make that true everywhere, then drop the `!`s it makes redundant.

- **Four more stylesheets are layered.** `og_application`, `registrations`, `documentation_v2` and `embed_styles` were unlayered, so their rules beat *every* tailwind utility — the rule only actually held on layouts loading `revised.scss`. Only `embed_styles` changes anything today; the other three load on layouts with no tailwind (`documentation` is `layout false`, `reg_embed` loads a single stylesheet), so they're pre-emptive. `email.scss` is deliberately left out — `premailer-rails` inlines it into outgoing mail and css_parser doesn't understand `@layer`.
- **One layer-order declaration, imported.** It only works if every entry point agrees, and six of them restated it in three different wordings. Now `legacy_includes/_css_layers.scss`, imported by all six; `application.css` isn't Sass so it keeps its own copy, which the partial says.
- **`embed_layout` loads tailwind** instead of hand-defining `.tw\:hidden` in an inline `<style>` — it renders `UI::Forms::LegacyFormWell::AddressRecord`, which is tailwind-styled and was getting exactly two rules. That brings preflight, which lands on everything `embed_styles` never restated, so each of those is restored with utilities: heading sizing on the error alert's `<h4>` and the success `<h1>`, list markers and indent on the validation-error `<ul>`, paragraph margins, link underlines, the file picker's chrome, and the base line-height. The form is back to main's exact height — 1096px desktop, 1220px mobile. Two caveats: the file picker is an *approximation* of native chrome (a class can't restore a UA-drawn control), and the widget now ships ~93K of tailwind on top of 39K of `embed_styles` into an iframe embedded on shops' own sites.
- **Nine `!`s dropped under `app/components/ui/`**, each one beating legacy: the definition-list `dd` and radio chip bottom margins, `ui-table`'s `border-collapse`, the form-well static-field padding, and one inside a `static_field_class` that nothing called. Kept where the target is a utility in the same layer — `period_select` beats `UI::Button`'s `:sm` size, and `tw:hidden!` is `collapse_utils.js`'s runtime vocabulary, which has to beat whatever display utility an element carries.

The embed's font, colour and line-height come from an **unlayered** `#bi_embed` rule rather
than a class: `body { @apply tw:base-font }` in `application.css` is itself unlayered, so no
layered rule — utility or otherwise — can outrank it, and between two unlayered rules an id
beats a type selector. Moving the tailwind rule into a layer would fix the trap generally, but
`revised.css` and `og_application.scss` both have `body { font-family: … }` in `legacy`, which
would then win and flip those pages' font — so not here.

Each `!` removal was checked against its rendered page rather than by reading the cascade, including on admin — where `admin.scss` imports a 251KB legacy file *outside* any layer, so "it's in `@layer legacy`, the plain utility wins" still can't be assumed there. The four `MARKUP_DIGEST` bumps are `bin/update_component_digests` output. 73 `!`s remain outside `app/components/ui/`.


sethherr:
## Screenshots

All four are parity checks. The embed form gains tailwind, so preflight resets are
restated with utilities to hold it where it was; the other three are the `!` removals,
where the plain utility emits the same declaration.

### /organizations/:slug/embed_extended

Restored to main's exact height (1096px desktop / 1220px mobile) after restating what
preflight resets. The file picker is an approximation of the native chrome — a class
can't restore a UA-drawn control — so it is the one visible difference.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /registrations/57

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/exports

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4084, opened 2026-08-09
Retry a register step the server 429s or 500s, and degrade without JS

sethherr:
A step that came back 429 or 500 used to be the end of the registration — the browser posted the form and the rider landed on an error page with everything they'd just typed behind it. The steps now submit through Turbo, which makes a failed step a response the page can take over: `register--retry` re-submits it behind the submit button's spinner, holding the button so a second click can't submit the step twice. A 422 is left alone, since that's the step's own validation answering.

- **When a retry won't help, the page says so.** rack_attack answers a throttle with `text/plain` and a `retry-after` counted in tens of seconds — Turbo renders a non-HTML failure as nothing at all, so retrying into it would leave a spinner and no explanation. The controller declines to wait out anything longer than a few seconds, and scrolls a notice into view while handing the button back with the form still filled in.
- **`request.format = :html` for every register action**, and the pages opt out of Turbo's snapshot cache the way they already opt out of the browser's. A Turbo submission asks for `text/vnd.turbo-stream.html` first, so the next step came back as a stream message grafted onto the step it should have replaced; and a back/forward restored from a snapshot showed a step the registration had moved past. Relatedly, `data-turbo` on a form opts in every link inside it, so the serial-guide link (which leaves for a legacy jQuery page) opts back out.

**A rider without JavaScript can now finish a registration** — before, step 1 was a dead end. A combobox submits through a hidden field only its Stimulus controller writes, so nothing picked in one reached the server, and step 1 requires a manufacturer.

- **`UI::Forms::Combobox` takes a `no_js:`**, rendering the plain control it falls back to inside a `<noscript>` along with a rule hiding the combobox it stands in for. A **select** by default — it posts the option's own value, so it needs nothing the combobox doesn't already hold and can't be mistyped. A **textbox** (options as a datalist) only where a list can't serve: free text, options fetched from an endpoint, or more than 99 of them. Of the five the flow uses, only the manufacturer is one. Deliberately not `search--pagination-fallback`'s render-both-and-remove-one shape — a duplicate *form field* live in the DOM before Stimulus connects can post over the real one, and this swaps nothing on load.
- **Two attributes that only mean anything with JavaScript are now applied by JavaScript.** A combobox with a fallback ships without `required`, because a required control the stylesheet has hidden is one no browser will submit past (`An invalid form control with name='' is not focusable`) — a rider would fill the fallback in and go nowhere. `ui--forms--js-required` marks it back on connect. Same inversion for the acknowledgment pages' submit, which shipped `disabled` and so was unpressable without JS; the server already turns away a submission that agreed to nothing.

- **The no-JS specs run in a real browser with scripting off** (`:playwright_no_js` — the same chromium, `javaScriptEnabled: false`). rack_test can't answer what this PR is about: it applies no stylesheets and runs no HTML5 constraint validation, so the form a browser refused to submit passed there four examples deep. The new driver renders `<noscript>` as DOM, applies the rule that hides each combobox, and validates — putting `required` back on the combobox now fails every example in the file rather than none. Costs 12s against rack_test's 1s.

Also rides along: `sandbox-test-setup` now covers starting `bin/dev` in the web sandbox, plus notes in `frontend-conventions` and `integration-testing` for what this work turned up. Unrelated to the rest of the diff; happy to split it out.


sethherr:
## Screenshots

The JavaScript-on shots below are unchanged from the previous capture: every change in this PR is inside a `<noscript>` or an attribute JS re-adds on connect, so the scripted rendering is identical by construction.

Both register steps, with JavaScript on — the no-JS fallbacks live in a `<noscript>`, so nothing about the JS rendering changes. Included as the parity check for that.

### /register?step=1 — with JavaScript disabled

The rendering nothing else in the PR shows, and the one the change is *for*. On `main` a scriptless rider gets a combobox whose input carries no `name` — measured: **zero** visible fields post `b_param[manufacturer_id]`, so submitting stays on step 1 forever. On this branch it's a plain textbox and a native select.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 (dead end) | this branch 👇 |
|  |  |

### /register?step=2 — with JavaScript disabled

Reached by filling in the step 1 fallbacks and pressing Next, which is not a page `main` can get a scriptless rider to at all — so there's nothing to compare against.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /register?step=1

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /register?step=2

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4085, opened 2026-08-09
Add PromotedAlert, a duplicate of TheftAlert

sethherr:
First of three stacked PRs renaming TheftAlert to PromotedAlert — the name the frontend has used all along. The three are shaped for a zero-downtime rollout: expand here, migrate and switch in #4086, contract in #4087. **Each one needs to be deployed and fully rolled out before the next merges.**

Stands up `promoted_alerts` and `PromotedAlert` alongside the existing table and model. Nothing in the app reads or writes them yet, and nothing outside the new files changes — `PromotedAlert` is a straight duplicate of `TheftAlert` with the two raw-SQL scopes repointed at the new table.

- **The migration reserves an id range.** #4086's backfill copies rows *keeping their ids*, but the app starts creating promoted alerts the moment #4086 deploys, which is before anyone runs the backfill. Those new alerts would take ids 1, 2, 3 — ids the copy still needs. The copy's `ON CONFLICT (id) DO NOTHING` would then silently skip the original alerts, and the notifications repointed at those ids would attach to whichever alert took them.
- **`ALTER SEQUENCE … START`, not a `setval`.** pg_dump writes `START` into `db/structure.sql`, so databases built by `db:schema:load` — CI, test, dev, review apps — get the reservation too. A `setval` would only ever exist in a database the migration itself ran against.
- **Reserving here rather than in the backfill is deliberate.** By the time #4086's code runs, this migration is long since applied, so there is no window where the reservation isn't in place.

## Why this has to ship on its own first

During #4086's phased restart (`restart_signal: usr1` in `.cloud66/manifest.yml` — workers recycle under a live master, so old and new serve side by side) the new code writes `notifiable_type`/`alertable_type` of `"PromotedAlert"`. Old workers dereference those polymorphic types with `constantize`. Shipping the class here, one deploy earlier, is what stops that being a `NameError`.

`StolenRecord` is untouched — the `has_many`, the scopes and the `theft_alert_missing_photo?` union all belong with the switch-over, so they land in #4086.

## Stack

1. **this PR** — add the table and model, reserve the id range
2. #4086 — backfill, and switch the app over
3. #4087 — drop TheftAlert


sethherr:
This would be a good change, but it isn't worth doing until we're actively working on this part of the codebase

PR #4086, opened 2026-08-09
Backfill promoted_alerts and read alerts through PromotedAlert

sethherr:
Second of three. Stacked on #4085 — **that has to be deployed and fully rolled out before this merges**, both because it reserves the id range this depends on and because it puts the `PromotedAlert` constant on every process before this one starts writing polymorphic types that name it.

## The backfill

`Backfills::PromotedAlertJob` copies `theft_alerts` into `promoted_alerts` verbatim: one `INSERT … SELECT` over `TheftAlert.column_names`, so every column comes across including `id`, `created_at` and `updated_at`. `ON CONFLICT (id) DO NOTHING` makes re-runs free. It then nudges the sequence forward — never back, so it can't undo #4085's reservation — and repoints `notifications.notifiable_type` / `user_alerts.alertable_type`, after the rows exist so the polymorphic associations never point at nothing.

**Run it once the deploy has finished rolling out**, not during it: until every process is on this code, an old worker meeting a repointed row is the one thing here that would actually break.

## Taking either alert

The app now creates and reads `PromotedAlert`s, and between this deploy and the backfill an alert may only exist as a `TheftAlert`. Two fallbacks cover that:

- `PromotedAlert.find_alert(id)` — ids are shared, so it falls back to `TheftAlert.find`. Used by the activate / facebook-update / notification jobs and the admin controller.
- `Payment#promoted_alert` — falls back to `payment.theft_alert`, gated on the payment's kind.

Both are scoped so they only fire when a specific alert is genuinely missing from `promoted_alerts`, which stops happening the moment the copy finishes. That is what lets #4087 drop the table in a single deploy: no code still running has a reason to query it. Keeping that property cost two small changes — the kind gate above, and `TheftAlert#missing_photo?` asking the stolen record directly instead of going through a predicate that now counts `promoted_alerts`.

The relation reads went the other way and dropped their `theft_alerts` half — the expiry sweep, the email merge, and `StolenRecord#theft_alert_missing_photo?`. In the window before the backfill they see an alert late (an expired one is swept on the next daily run; the missing-photo nudge is a day late), which isn't worth a query against the old table on every stolen record.

## Deliberately unchanged

Routes, controller classes and view paths (`/admin/theft_alerts`, `/bikes/:id/theft_alert`), and the `theft_alert` enum values stored on `payments.kind`, `notifications.kind` and `user_alerts.kind`. Those are URLs and persisted data, not the model. `admin_path_for_object` gets an explicit `PromotedAlert` case so it keeps producing the real admin path instead of `/admin/promoted_alerts/:id`.

## Rough edges in the restart window

Two things are worth knowing about rather than engineering around, both bounded by the phased restart:

- A job enqueued by new code for a brand-new alert, picked up by an old worker, raises `RecordNotFound` and retries — the jobs are configured for 3–4 attempts, which outlasts the restart.
- If someone completes a Stripe checkout mid-restart and both their return request and the follow-up `AfterUserChangeJob` land on old workers, their alert stays pending. After running the backfill, check `PromotedAlert.paid.pending.where(start_at: nil)` around the deploy time and use the admin Activate button on anything stuck.
- An admin form loaded before the deploy and submitted after posts `theft_alert[…]` and gets a 400. Reload and resubmit.

## Testing

The 571 examples across every spec touching alerts pass, including new coverage for both fallbacks and for the backfill copying every column, keeping the timestamps, not duplicating on a second run, and not winding the sequence back. The VCR cassette directory moved with the renamed context (contents untouched). Frontend diffs are variable renames inside the admin and bike alert views with no visual change, so there are no screenshots.

Not run here: `spec/components/page_block/user_alerts/wrapper/component_system_spec.rb` — this sandbox has no Playwright browser.


sethherr:
This would be a good change, but it isn't worth doing until we're actively working on this part of the codebase.

[Duplicate of comment on first PR in the stack](https://github.com/bikeindex/bike_index/pull/4085#issuecomment-5330384978)

PR #4087, opened 2026-08-09
Remove TheftAlert

sethherr:
Third of three, the contract step. Stacked on #4086.

**Merge gate: #4086 deployed and fully rolled out, and `Backfills::PromotedAlertJob` finished.** The whole safety argument below rests on the copy being complete.

## What goes

- `app/models/theft_alert.rb`, its spec and its factory.
- The `theft_alerts` table (`DropTheftAlerts`, reversible).
- `Backfills::PromotedAlertJob` and its spec — it selects from `theft_alerts`, so it can't outlive the table.
- The transitional fallbacks from #4086: `PromotedAlert.find_alert` (call sites go back to `PromotedAlert.find`), the `Payment#promoted_alert` override and its `has_one :theft_alert`.
- `has_many :theft_alerts` on `StolenRecord`, `User` and `Bike`, and the `with_theft_alerts` scopes.

`/admin/theft_alerts` and `/bikes/:id/theft_alert` keep their URLs, and the `theft_alert` enum values on payments, notifications and user_alerts are untouched — including `StolenRecord#theft_alert_missing_photo?`, which is named for the `theft_alert_without_photo` UserAlert kind it feeds rather than for the model.

## Why the drop is safe mid-restart

Deploys here are a Puma phased restart, so #4086's code is still serving requests while this migration runs. That only works because #4086 left exactly two paths that touch `theft_alerts`, and both are guarded by "this alert is missing from `promoted_alerts`" — `PromotedAlert.find_alert`'s fallback and `Payment#promoted_alert`'s kind-gated fallback. Once the backfill has run, neither condition can be true, so no live process has a reason to query the table. Nothing else in #4086 reads it: the relation reads were already narrowed to `promoted_alerts` there, precisely so this step could be one deploy instead of two.

`drop_table` takes an ACCESS EXCLUSIVE lock, which for the same reason has nothing to wait behind.

If the backfill has *not* finished, this is not safe — those fallbacks would start raising, and the alerts that were never copied would be gone. Hence the merge gate.

## Testing

`spec/models`, `spec/jobs`, `spec/helpers`, `spec/mailers`, `spec/services`, `spec/requests` and `spec/components` pass, except four image specs that fail because this sandbox has no ImageMagick (`executable not found: "mogrify"`) — nothing to do with these changes.

`:js, type: :system` specs weren't run: this sandbox has no Playwright browser.


sethherr:
This would be a good change, but it isn't worth doing until we're actively working on this part of the codebase.

[Duplicate of comment on first PR in the stack](https://github.com/bikeindex/bike_index/pull/4085#issuecomment-5330384978)

PR #4093, opened 2026-08-10
Upload the embed form's photo straight to storage

sethherr:
The embed forms' photo now uses `UI::Forms::FileUpload::Component` with a direct upload url — the same field the `/register` flow's step 2 uses, so the bytes go browser → R2 and the form posts the blob's signed id. Without JS the field still posts its own bytes.

- **They upload through `/register/direct_uploads`.** An embed registration's b_param belongs to the organization handing out the form, so `find_token` — which resumes a registration — won't hand it to the anonymous person filling it in; the fallback is `BParam.find_from_token`, the lookup the embed page itself built the form from. A second endpoint would only have been a second budget for the same anonymous b_params.
- **That budget goes from 10/hr to 60/hr per IP.** It was sized for a registration needing one upload; a shop's embed form registers bike after bike from the one address.
- **`BikesController#create` lifts `image_signed_id` out of the bike attributes** and carries the stored one through a resubmission — the embed branch replaces the b_param's params rather than merging them, so a validation error would otherwise drop the photo.
- **The embed form submits itself** from its `strip-inputs:valid` handler, and a native `form.submit()` fires no event for the upload to hold. Holding now runs in the document's capture phase, and `strip-inputs` bows out of a submit something else already cancelled.
- **Riding along, unrelated to the embed change:** CI cancels a superseded run (except on `main`, where `deploy_production` runs); two search specs that were failing on loaded CI shards are fixed at their causes — a click landing mid frame-swap, and a probe reading `defaultPrevented` inside its own listener, where it reports registration order rather than the app's verdict; and a `fixing-flaky-failures` skill collects the diagnosis.


sethherr:
## Screenshots

### /organizations/:slug/embed

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /organizations/:slug/embed_extended

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4102, opened 2026-08-12
Fix typing over a combobox selection, hiding a missing serial, sticker gating

sethherr:
Three fixes, all found walking the register flow.

- **Typing over a combobox selection replaces it now.** Clicking a combobox that held a selection left the caret wherever the click landed, so typing inserted into the middle of the display text — the result matched no option, and the gem cleared the hidden field along with the selection the rider already had. The display is selected on the click rather than the focus, which is what covers an autofocused field like step 1's manufacturer, where the entering click brings no focus event of its own; the gem's own refocus while filtering still never selects. It's bound through the target-connected callbacks the gem declares but leaves empty — the seam that also reaches the small-viewport dialog's combobox, which takes no action from the caller. That does mean select-on-entering-click for every combobox in the app, deliberately: it's what the browser's own URL bar does, and the alternative is opting in at 29 call sites.
- **A serial marked missing hides its input instead of greying it out.** `readonly` left "unknown" sitting in a field the rider could neither edit nor clear; the input collapses away instead, still holding the value that submits. Unchecking has to undo an *animated* hide, which `CollapseUtils.show` couldn't do at duration 0 — it removed the `display: none` and returned with the element still pinned to `tw:scale-y-0` and `height: 0`, so the serial never came back, and being required and empty it blocked submission with nothing on screen to fix. Every duration-0 caller was one animated hide away from the same thing.
- **Step 2 only asks for a bike sticker when there's one to give**: an organization that lets ordinary users assign its stickers (`bike_stickers_user_editable`, the way `bikes/new` gates it), or a registration already carrying a scanned code. Everyone else was being asked for a code they had no way to have — and with no organization on the registration, `BikeServices::Creator` never even looks a code up.

`register_spec`'s e-vehicle example is flaky, and not because of this branch: at the failed submit `bike[user_name]` is empty, the `fill_in` right after step 1 having gone missing — it's the one example there that doesn't wait for the page before filling. Run back to back with identical instrumentation it fails at the same rate with and without these commits (1 in 8 either way), and adding any `execute_script` to the flow hides it entirely. Worth its own look rather than blocking here.

**Riding along, unrelated to the fixes:** `fixing-flaky-failures` gains the difference between "is this spec flaky" and "did my change make it flaky" — the second needs both arms run back to back, which is what settled the question above — and `integration-testing` gains why a component preview can pass while the real page is broken.


sethherr:
## Screenshots

### /rails/view_components/register/step2/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4104, opened 2026-08-12
Add UI::PhoneDisplay, replacing the phone helpers and hand-rolled tel: links

sethherr:
`phone_link` and `phone_display` were view helpers, so the only way to get a formatted phone number was from a template — which is how three hand-rolled `tel:` links ended up beside them, rendering the same values unformatted. `UI::PhoneDisplay::Component` replaces all of it: a `tel:` link by default, plain text with `skip_link: true`.

- **The three hand-rolled links are converted too**, not just the helper call sites. Both were in `info/where`, formatting one shop's `location.phone` two ways on one page, and `Registrations::Show::StolenDetails` carried the same class string as the two panels next to it. They're the failure mode a greppable helper name doesn't catch, so the frontend-conventions skill now states the rule the way it does for `UI::Time`.
- **The component owns its classes** — `twlink tw:font-mono` linked, mono alone for `skip_link`, overridable via `html_class:`. The call sites passing that string by hand drop it, and the pdf's `phone-number-link` goes with them since nothing styled it.
- **Formatting moves to `Phonifyer.display`** for the two callers that need a String rather than markup — a `text_field` value and a translation interpolation. It returns an unescaped string, where the old helper handed those two an `html_safe` one from ActionView's `number_to_phone`; escaping now happens at the view boundary.

Unrelated to the above, the last commit bumps brakeman to 8.0.6 — `bin/brakeman` runs `--ensure-latest`, so the release broke `lint_and_scan` on every branch.


sethherr:
## Screenshots

Seeded dev data carries no phone numbers, so these use a stolen record set to `3025551234 x204` (secondary `+44 7802 730000`) and a user phone of `+91 8041505583 x2929` — chosen to exercise the extension and country-code paths.

### /registrations/43

`main` renders the owner phone two ways on one page: `ContactOwner` went through `phone_link` and is formatted, while `StolenDetails` hand-rolled its `tel:` and shows the raw `3025551234 x204` with no dial pause in the href. Both are the component now.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes/43/edit

`skip_link` mode — the phone and secondary phone now render mono.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /my_account/edit

The phone list entry, plus the `Phonifyer.display` form-value path.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4105, opened 2026-08-13
Fix the blank search buttons and the page the combobox locks

sethherr:
An EPS user reported the blue search button rendering as an empty blue box, and the page refusing to scroll after searching by make, model or color. Both are real, and a third search button turned out to be in the same state.

- **`icons/searcher.svg` carried no width or height**, and it is the button's entire content — so a browser that resolves a viewBox-only SVG to zero draws a blank button. It was the one inline icon in the app in that state. The two components size it explicitly on top of the intrinsic dimensions, keeping what they render today; `shared/_bike_search_form` is left alone, since `.searchit svg` already sizes it there.

- **The combobox locks body scroll below 640px and only unlocks along its own collapse path**, which a keypress or a click has to start. Android's back gesture closes the dialog through neither, leaving the page unscrollable until reload. The patch hooks `_openInDialog` rather than `connect`/`disconnect`: the registered controller is a subclass that redefines both, so an override of those never runs.

- **`/stolen`'s search button has been drawing a broken image since packs went away.** Its vendored webpack bundle emits the icon's URL from webpacker's publicPath, so it asks for `/packs/media/…` and nothing serves it. That one is worth showing the reporter — Bootstrap's white button color means a browser that renders alt text draws the word "search" there and one that doesn't draws a plain blue box, which matches the report more closely than anything on the search page.

Carries one unrelated fix: `search_registrations_spec`'s back/forward example failed through all four retries on [#4104](https://github.com/bikeindex/bike_index/pull/4104), and the Capybara screenshot in that run's artifacts is a blank page — so `current_path` was nil for about:blank, not either mechanism the `flaky: 4` comment names. Only the `go_back` half pinned the address bar, and it's the `go_forward` half that then navigates away; both assert it now.


sethherr:
## Screenshots

### /stolen

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The search-page icon is unchanged on purpose — it renders 54×54 in Chromium either way. The fix is for browsers that resolve a viewBox-only SVG to zero.

PR #4106, opened 2026-08-13
Gate registration sequence editing behind its own organization feature

sethherr:
`registration_sequences` granted viewing and editing together. It splits in two, the way `bike_stickers_user_editable` splits from `bike_stickers`: the existing slug now grants read-only access — the current sequence, the archive, and the registrant preview — and the new `registration_sequences_edit` is required to change anything.

- **Without the edit feature the draft is invisible, not merely locked.** No draft section on the index at all — not an empty state, not a read-only listing — and previewing one by URL is refused. An organization that can't act on a draft has no reason to know it exists.
- **`RegistrationSequencePagesController` is gated outright**, since every action there mutates the draft. The sequences controller gates everything except `index` and `show`, and `show` refuses drafts per-record.
- **The edit slug implies the view slug** in `calculated_enabled_feature_slugs`. Granted alone it would lock the organization out of the controller entirely — the opposite of what it grants — and the parent invoice's "features passed on to children" field is free text, so that combination is reachable.

The seed feature record grants both slugs, with a "view only" variant beside it mirroring "Bike Stickers: NOT user editable". Production's `OrganizationFeature` rows need the same split before this changes anything for a real organization.


sethherr:
## Screenshots

An organization with `registration_sequences` but **not** `registration_sequences_edit`, signed in as an org admin. On `main` there's no such distinction, so the draft section renders for anyone with the feature.

### /o/:organization_id/registration_sequences

| Desktop | Mobile |
| --- | --- |
| registration sequences index on main, draft section visible | registration sequences index on main, mobile |
| main 👆 | this branch 👇 |
| registration sequences index on this branch, draft section absent | registration sequences index on this branch, mobile |

The Draft heading, the "You don't have a draft sequence" empty state and the "Create a sequence" button are all gone — not disabled, absent. An organization that also has the edit feature sees the page unchanged.

PR #4107, opened 2026-08-13
Show permitted orgs the owner's phone and other registrations

sethherr:
The registration show page withheld two things from law enforcement that the legacy bike show page has always given them: the owner's phone number, and the "Registrations by user" section.

- **The org action panel offers the phone**, keyed on `Bike#phoneable_by?` like everything else — the legacy access panel's unstolen-notification box has that line, and the redesign's counterpart didn't. It reads `Bike#phone`, so phone registrations and the unstolen fallback to the user's own number are covered, not just the stolen record's.
- **Other registrations are gated on the `additional_registrations_information` feature and nothing else**, as the legacy panel gates them. Requiring `show_contact?` as well (full staff, on the org's own registration) meant a law enforcement member searching someone else's bike — the case the feature exists for — lost the section. It renders "None found" rather than disappearing when the owner has no others, so the answer is visible either way.
- **`Registrations::Show::OwnerPhone`** holds the "Or call" line the contact card and the message panel now share.
- **Dev seeds could not reach any of this**: Bike Recovery Team carried the Law Enforcement features but `kind: "other"`, and the police check reads `Organization.law_enforcement`, not the feature slugs. The team is law enforcement with a member (`recovery@bikeindex.org`) now, and the seeded stolen records carry a phone — the first police-only, the second public, the rest signed-in users.


sethherr:
## Screenshots

Signed in as `recovery@bikeindex.org`, the seeded Bike Recovery Team member. Bike 35's phone is police-only; bike 26's is public.

### /registrations/35?panel=message

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /registrations/26

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4108, opened 2026-08-13
Place the user alert clear of the page shells it collided with

sethherr:
The general alert renders below the navbar on every page, which put it underneath three different things: the homepage's viewport-fixed bike tile grid, the landing pages' absolutely positioned one, and the organization menu.

- **The layout and `organized_skeleton` each render `Wrapper::Component`**, the skeleton passing `wrapper_class: nil` so it sits in the content column the menu doesn't cover; the wrapper's `render?` defers to the kind's own component, so the layout can ask whether an alert will really show before making room for one.
- **The navbar's gap is published as `--nav-gap`**, replacing the hardcoded `-mt-9`/`-mt-15` pairs the two full-bleed shells carried and the four `pt-9`/`pt-15` cancelers spread across org, admin and Lookbook — the layout zeroes it on `<main>` when an alert stands in that gap, so the registration page's photos get pushed down rather than sliding underneath it.
- **The landing pages carry their hero color on the body**, since their hero sits flush under the navbar and an alert standing in that row was otherwise a white band above it.
- **An unfinished registration made for someone else no longer alerts its creator**, which is what an organization user registering a customer's bike, or anyone registering a friend's, was getting.

`show_general_alert` now owns giving way to the law enforcement donation modal, which gets the first spec to render a page with that session flag set — and caught that the modal had stopped showing at all. `PageBlock::UserAlerts::Banner` folds into the wrapper, which is where the container it existed for now lives.


sethherr:
## Screenshots

### /

| Desktop | Mobile |
| --- | --- |
|  |  |

### /resources

| Desktop | Mobile |
| --- | --- |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |

### /registrations/42?view_as=public

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4109, opened 2026-08-13
Color the framed step 1's button from the frame's ?button=

sethherr:
The framed step 1 on an organization's landing page takes a hex color for its button, the same `?button=` the legacy embed form has always taken — `/register/embed?organization_id=brakebills&button=c9a227`. Builds on the frame from [#4112](https://github.com/bikeindex/bike_index/pull/4112).

- **Only the framed variety is colored.** The color belongs to the page the frame sits on, so it rides the iframe's `src` — one request, nothing held in the session, and the flow's own full-width pages ignore the param.
- **Hover darkens the color it was given**, via `HexColor.darken` in hsla. An inline background outranks the palette's hover class, so the shade comes back as an `!important` utility off a CSS variable — Tailwind only generates classes it can read literally, which is what the variable is for. Guarded with `not-disabled:not-aria-disabled:` like every `UI::Button` hover, since an `!important` would outrank that guard too.
- **`HexColor` replaces `RegistrationsController#valid_hex`**, and validates rather than truncates — `?button=@user + 1233` was rendering `background: #user12`, which browsers ignore anyway, so nothing that isn't a 3- or 6-digit hex reaches the attribute at all. Both entry points sanitize at the controller.

Brakebills' seeded landing page names its gold once and frames the form in it, so the button and the page's own borders can't drift apart.


sethherr:
## Screenshots

### /register/embed?organization_id=brakebills&button=c9a227

The frame the landing page embeds, in Brakebills' gold.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4110, opened 2026-08-13
Add a non-e-vehicle kind to registration sequences

sethherr:
`RegistrationSequence` gains a `kind`: the existing sequence becomes `e_vehicle` and `non_e_vehicle` sits beside it. An organization runs both independently — each with its own draft → activate → archive lifecycle, its own template to clone from, and its own section on the index. Existing rows are all `e_vehicle`, so nothing changes for an organization until it drafts the new kind.

- **Every lookup takes a required `kind:`**, so no call site can silently resolve the wrong sequence, and the four partial unique guards widen to `(owner, kind)` — one draft and one live sequence per kind. A link naming a kind that isn't one 404s rather than quietly opening the e-vehicle draft.
- **The register flow picks the kind from what's being registered.** A bike at an organization with an active non-e-vehicle sequence now walks its acknowledgment pages, where before only e-vehicles had any. The "⚡️ Electric (motorized) detected" badge stays e-vehicle-only, and the acknowledgment eyebrow and FAQ link become per-kind translations rather than an English noun interpolated into a translated sentence.
- **Seeds gain a non-e-vehicle template** — riding and parking rules, no battery page — and give Brakebills a live sequence of each.

Two unrelated changes ride along. The seeded `OrganizationFeature` names shorten to "Registration Sequences: Edit" and ": View Only". And `db:seed` now builds the dartsass stylesheets itself when `app/assets/builds` is empty — a bare `bin/rails db:seed` on a fresh workspace died on `stylesheet_link_tag "email"` — which lets `bin/setup` drop the unconditional build it ran for that same reason.


sethherr:
## Screenshots

The dev DB is shared across the checkout, so `main` is rendering the same four sequences this branch seeded — the before shots are the old code showing today's data, which is what makes the third pair below the interesting one.

### /o/brakebills/registration_sequences

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/registration_sequences

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/registration_sequences/4

The same non-e-vehicle sequence on both sides. `main` has no concept of a kind, so it labels it "E-Vehicle Acknowledgment" and titles it plain "Current"; this branch reads "Current Non-e-vehicle" and "Registration Acknowledgment", and drops the "⚡️ Electric (motorized) detected" badge.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4111, opened 2026-08-13
Start the new registration flow from an organization's page

sethherr:
`/o/{slug}/registrations/new` starts the `/register` flow as a registration belonging to the organization — its opening step rendered there, its submission and every step after continuing on `/register`. It cross-links both ways with the embed form it replaces, which #4185 will remove along with everything marked for it here.

- **Registrations started there take a new `Ownership::ORIGIN_ENUM` origin, `register_flow_organized`**, which the created bike's ownership carries. The places keyed off the literal `"register_flow"` move to `BParam#register_flow?`, and `reusable?` now requires a matching origin, so arriving from the org page doesn't take over a blank shell started on `/register` and register the bike as though it came in there.
- **The organization's menu follows the flow.** The sidebar's "Add a bike" row points at it, and going back to the embed form sets a preference that flips the row back — cleared by taking the flow's link again, passed into `UserServices::MenuItemsOrg` and folded into its cache key, since the menu is cached per organization and user. `api/v3/me` serves the same menu, so its "Add a bike" path moves with it.
- **Step 1 doesn't seed the owner email here**, unlike `/register` — the member is registering someone else's vehicle, so the field starts empty and shows the organization's configured placeholder.
- **The embed form's frame is sized to what it renders**, rather than to a `min-height` that was wrong both ways: the registration form is ~1430px and scrolled inside it, the parking notification one ~850px and left dead space below. Only the page this links back to, which is same-origin, so `org--embed-iframe` can measure it.
- **`current_user_or_unconfirmed_user` drops out of the sidebar it was threaded through.** The sidebar only renders for a confirmed user, so it was always `current_user` — `UserServices::MenuItemsAccount` takes one `user:` now, and `PrimaryMenu`/`UserSettingsMenu` stop forwarding what they only held to pass down.


sethherr:
## Screenshots

### /o/{slug}/registrations/new

New route — nothing on `main` to compare against.

| Desktop | Mobile |
| --- | --- |
| org registrations new, desktop | org registrations new, mobile |

### /o/{slug}/bikes/new

| Desktop | Mobile |
| --- | --- |
| org bikes new, base, desktop | org bikes new, base, mobile |
| main 👆 | this branch 👇 |
| org bikes new, desktop | org bikes new, mobile |

PR #4112, opened 2026-08-13
Frame the register flow's step 1 on organization landing pages

sethherr:
Brakebills' landing page framed the legacy `/registrations/embed` widget. It now frames the register flow's step 1, and pressing Next navigates the landing page itself onto step 2 rather than continuing inside the box. The app-wide `X-Frame-Options: SAMEORIGIN` already allows exactly that and refuses everything else, so nothing new is needed to permit the frame.

- **`register#embed` renders `Register::Embed::Component` with `layout: false`** — the component is the whole document, since the embedding page is the chrome: no nav, no footer, no analytics (the page around it counts the pageview), plus `noindex`. It renders rather than redirecting into a tokenized step, so the frame is one request and nothing past step 1 is embeddable.
- **Framed, the flow drops its gray full-bleed shell and the organization's name** — the frame is the shell, and the page around it already says whose registration this is. The form takes `target="_top"` and `data-turbo="false"` (Turbo ignores a target that doesn't name an iframe, and would render step 2 back inside the box), and loses autofocus, which scrolls the embedding page down to the frame on load.
- **Step 1 asks to "Register your vehicle", everywhere** — the heading and the page title used to name the chosen cycle type, which is answering the question the step is still asking. The section label ("Bike info") keeps the live swap.
- **The email placeholder comes from the organization**, the way the legacy embed form reads it: the label a paid organization sets in admin wins, a school's own name stands in otherwise, and everyone else keeps `you@example.com`. Without this, replacing the widget would have dropped the campus-address prompt universities ask for.
- **The combobox picker measures the screen, not the frame.** The gem checks its own window against the mobile breakpoint, so a 420px iframe opened the full-screen dialog on a desktop — inside the frame, where it can't be full-screen. It reads the top window now, so a phone still gets the dialog and a desktop drops down inline. Unframed, `window.top` is `window`; Lookbook previews are exempt, since dragging its viewport handle is how they're reviewed.

The legacy widget stays — it's the only one that works off-site, and it carries per-org styling this has no equivalent for. One caveat: `/register/embed` creates a registration row per new visitor where the legacy form created none.

Carries one unrelated fix: `register_spec`'s e-vehicle example types into step 2's color combobox straight after step 1's Turbo navigation, and failed ~2 runs in 3 on an unmodified tree. Waiting for step 2's content took it to ~1 in 3; it's tagged `flaky: 4` for the rest, since nothing on the page proves the combobox has hydrated.


sethherr:
## Screenshots

### /brakebills

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4113, opened 2026-08-13
Seed a landing page for Brakebills

sethherr:
`landing_html` had no seed data, so nothing in dev or a review app showed what an organization landing page actually looks like. Brakebills gets one now, adapted from Princeton's production page — the same layout, icons and embedded registration iframe.

- **The template is ERB** (`db/seeds/brakebills_landing_page.html.erb`), rendered by `seed_organizations.rb`, so the header logo can point at the seeded org's own avatar rather than a hardcoded upload path. Copy and campus links are Brakebills; the background and bottom thumbnails keep Princeton's photos, since there are no Brakebills ones.
- **`ORGANIZATIONS_WITH_LANDING_PAGES` defaults to `brakebills`** rather than `ikes university`, so `/brakebills` routes. Neither old slug had anything behind it: Ikes Bike's has no seeded `landing_html`, and there is no university organization at all.
- **`.herb.yml` excludes `db/seeds/**` from `erb-prefer-image-tag-helper`** — seed templates render through `ERB.new`, which has no view helpers.

Viewable at `/brakebills` and `/o/brakebills/landing`. Production sets `ORGANIZATIONS_WITH_LANDING_PAGES` — `/princeton` routes there and `/ikes` doesn't — so changing the fallback only affects dev, test and review apps.


sethherr:
## Screenshots

`/o/brakebills/landing`

Brakebills landing page, desktop

Brakebills landing page, mobile

PR #4114, opened 2026-08-13
Register with the registrant's own organization, unless they say otherwise

sethherr:
The register flow only ever attributed a registration to the organization a link named. Without one, it now takes the single organization the signed-in registrant belongs to — or failing that, the single one their other bikes are registered with. Two of either says nothing about this bike, so it stays unattributed.

- **Step 2 leads with a checked "Register with {org}"**, and only for an automatic assignment — one a link named isn't up for debate. Unchecking drops the organization and everything it asks for: its own fields, the address it requires, and the heading naming it, since a required field nobody can reach would block the submission outright.
- **The assignment is recorded either way**, "nothing to assign" included, so it's made once rather than re-derived on every request through the flow — and so unchecking survives the next one, with the checkbox still there to change their mind.
- **The filter covers #4112's landing-page embed too**, where the link always names the organization, so that one still wins.

`register--organization` and `register--status-fields` both have a say in the address, so the new controller sets a flag the existing one reads and announces the change rather than applying it — neither depends on which connects first. Their shared "collapse and disable" moves to `collapseField` in `collapse_utils`.


sethherr:
## Screenshots

### /rails/view_components/register/step2/component/with_auto_organization

The organization assigned automatically — its checkbox heads the section whose contents it decides. New preview, so nothing on main to compare it to.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/register/step2/component/with_organization

An organization a link named — unchanged, as it should be: no checkbox, and the fields the new targets wrap render exactly as before. The pair differs by 122 pixels, all of them the dev render-timing badge.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4118, opened 2026-08-13
Read and write landing pages through OrganizationLandingPage

sethherr:
Follow-up to [#4116](https://github.com/bikeindex/bike_index/pull/4116), which added `OrganizationLandingPage` and backfilled it from `organizations.landing_html` without reading it. This swaps the readers and writers over. **Run the backfill immediately before this deploys** — a landing page with no row renders empty, and the backfill skips organizations that already have one, so a `landing_html` edit made after it ran won't be picked up.

- **`landing_html` stays put**, it just has no readers left. Dropping the column is the next PR.
- **The admin form edits the landing page directly**, rather than reaching it through `Organization` — which gains nothing here, not a method and not `accepts_nested_attributes_for`. Opening the editor builds an unsaved record, so browsing to it no longer leaves a blank page behind the way the mail snippet beside it does, and the first save is one insert.
- **The editor links to the page's version history, and saving reports drift.** `enabled` is a copy of `ORGANIZATIONS_WITH_LANDING_PAGES` that only the backfill job reconciles, so a save flashes an error naming both sides when they disagree — nothing else would say so.
- **The landing page touches its organization**, so the fragment stays keyed on the organization and a cache hit costs no extra queries. `organizations.updated_at` doesn't move on its own when the body changes, which is why an edit used to stay invisible until something else touched the organization.

`enabled` still isn't a gate: `LandingPages::ORGANIZATIONS` draws the routes at boot, so the column can't be authoritative until routing reads the table. That, and the "Enabled" admin cells still keyed on the constant — which means the one screen showing enabled state can't display the drift the new flash reports — belong with the routing change.

### DIFF
--- a/.claude/skills/rspec-testing/SKILL.md
+++ b/.claude/skills/rspec-testing/SKILL.md
@@ -22,7 +22,7 @@ When something fails outside the files you changed, re-run that spec file on its
 ## What to test (and what not to)
 
 - Tests should either: help make the code correct now, or prevent bugs in the future. Don't add tests that don't do one of those things.
-- Use **request specs**, not controller specs — request specs go through the full middleware/routing stack, so they catch breakage controller specs can't see. Everything making the same request should be in a single test.
+- Use **request specs**, not controller specs — request specs go through the full middleware/routing stack, so they catch breakage controller specs can't see. Everything making the same request should be in a single test. For markup a component owns, see [A component's own markup](#a-components-own-markup-is-tested-in-its-component-spec).
 - Avoid testing private methods.
 - Avoid mocking objects.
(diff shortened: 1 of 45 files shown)


sethherr:
## Screenshots

### /brakebills

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Identical, which is the point — `main` renders `landing_html`, this branch renders the same content out of `OrganizationLandingPage` (both captured at 1440×1718 after running the backfill locally).

The admin editor and the two admin checkmarks also changed, but no seeded user can reach them: `admin@bikeindex.org` is a superuser without `developer?`, and `dev@bikeindex.org` is a developer without superuser.

PR #4119, opened 2026-08-13
Label register step 1's email field with the organization's custom label

sethherr:
Register step 1 read the organization's `owner_email` label into the email field's *placeholder* (#4112). It labels the field with it now — which is what `bikes/new` and the org embed fields have always done with the same setting — and a new `email_placeholder` registration label takes over the placeholder. A school's short name still stands in when nothing is set ("Brakebills email"), now as the label.

- **`UpdateOrganizationAssociationsJob` fills a blank `email_placeholder`** with `you@user_email_domain`, or the auto user's domain when that isn't set — skipping the Bike Index account that stands in for an organization without members of its own. An admin's own value is never overwritten.
- **Both keys are editable in the admin organization form**, gated on `paid?` rather than a feature slug. The three near-identical `reg_label-` inputs collapse into one skeleton that varies only its copy.
- **Brakebills seeds an `owner_email` label**, so dev and review apps show a customized field.

Worth a look before this merges: an organization that set `owner_email` to an example address (`you@uiowa.edu`) — because the legacy `/registrations/embed` renders that key as a placeholder — will now see it as the field's label, and wants moving to the new key.

### DIFF
--- a/app/components/register/step1/component.en.yml
+++ b/app/components/register/step1/component.en.yml
@@ -8,7 +8,7 @@ en:
         electric_motorized: "⚡️ Electric (motorized)"
         email: Email
         email_placeholder: you@example.com
-        email_placeholder_school: "%{org_name} email"
+        email_school: "%{org_name} email"
         just_the_essentials: Just the essentials to start.
         next: Next
         no_need_to_wait: We'll email a confirmation link — you don't have to wait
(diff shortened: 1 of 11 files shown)


sethherr:
## Screenshots

### /register/new?organization_id=brakebills

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4120, opened 2026-08-13
Rebuild the admin organization form as UI::Forms view components

sethherr:
The admin organization form becomes a set of view components, and every field in it goes through the `UI::Forms` component library instead of hand-rolled Bootstrap markup. Both partials it was built from — `_form.html.haml` and `_location_fields.html.haml` — are gone, and `edit.html.haml` is down to a `form_for` and a submit button.

- **Three components under `Admin::OrganizationForm`**: `Wrapper` holds the core fields and renders `FeatureSettings` (the "Settings for organization features" card, which gates itself with `render?`) and `LocationFields` alongside them. Each derives what it needs from the form builder rather than reading controller ivars, so `edit` has no body left, and the locations section moves in from `edit.html.haml` gated on `persisted?`.
- **`UI::Forms::NestedFields` renders an association's whole editable collection** — the saved records, a blank set in a `<template>`, and the add link — building each record's wrapper itself and taking the per-record component as an argument, so an added record can't drift from a saved one. Its Stimulus controller follows the contract [`@stimulus-components/rails-nested-form`](https://www.stimulus-components.com/docs/stimulus-rails-nested-form/) established, which retires the form's last dependency on the vendored admin bundle's jQuery handlers.
- **Every note qualifying a field moves out of its label into `Group`'s `helper_text` slot**, which is what gets it `aria-describedby` — passed explicitly for the fields rendering their own input through a content block, including the comboboxes. Checkboxes keep theirs inline: `UI::Forms::Checkbox` has no slot, and for a checkbox the note reads as label text anyway.
- **Removing a location an admin had just added no longer blocks submission.** It used to stay hidden in the form, where its required address fields rejected a control nobody could see; a never-saved record is now detached, and only a saved one is hidden with `_destroy` set. Two smaller changes: the ambassador toggle's label targets move to the column wrapper, because `Group` renders the label itself, and the stolen-message radius fields drop the form builder they were never scoped to, so their labels point at the field they belong to instead of an id that never existed.

`link_to_add_fields` keeps its two other callers and picks up a note pointing at the replacement — porting `organized/manages/locations` means porting its page-specific CoffeeScript too, so it's left out. The address inputs inside a location card still come from `Org::LocationAddressFields::Component`, which the organized pages share.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4124, opened 2026-08-14
Drop organizations.landing_html, add a landing pages admin index

sethherr:
Follow-up to [#4118](https://github.com/bikeindex/bike_index/pull/4118), which left `organizations.landing_html` in place with no readers. The column goes here, so this depends on that PR's backfill having run — there's no copying it across afterwards.

- **`Backfills::OrganizationLandingPageJob` is deleted.** Its `landing_html` copy went with the column, and reconciling `enabled` is the editor's checkbox now.
- **A `Dev:` admin index for landing pages**, whose Enabled column flags drift from `ORGANIZATIONS_WITH_LANDING_PAGES` — #4118 noted the one screen showing enabled state couldn't. Rows link to the editor, the organization and the version history.
- **The editor sets `enabled`, and alerts only when it disagrees with the constant** rather than flashing the disagreement after a save. Its header button links to the public page for the organizations that route one.
- **Seeds**: the landing page template is generic now (`organization_landing_page.html.erb`, rendered for an `organization`) and its register frame names the hover shade step 1 was deriving; `dev@bikeindex.org` holds superuser and developer, so one sign-in reaches the pages needing both. The spec factory renders the template too, for a default body the size of a real one.

`enabled` still isn't a gate: `LandingPages::ORGANIZATIONS` draws the routes at boot, so the checkbox keeps the copy honest rather than making it authoritative. Making it authoritative is the routing change's job, and it's what would retire `enabled_mismatch_error` and its alert.


sethherr:
## Screenshots

### /admin/organization_landing_pages

| Desktop | Mobile |
| --- | --- |
|  |  |

### /admin/organizations/brakebills/custom_layouts/landing_page/edit

| Desktop | Mobile |
| --- | --- |
|  |  |

No `main` comparison for either: the index is a new route, and the branch's migration drops a column `main`'s code still selects, so the base can't render against this database.

The editor still shows the `button_hover` notice because this database was seeded before the template named its hover shade; a fresh `db:seed` renders it without.

PR #4125, opened 2026-08-15
Rework the pr skill into named sections and fix what it turned up

sethherr:
The `/pr` skill's steps ran `0 → 0.5 → 0.7 → 1 → 2 → 3`, and that order was wrong in a way the numbering hid: step 0 linted and audited against `origin/$BASE` before step 0.5 chose the base and before step 1 fetched it. `$BASE` never survived anyway — each command runs in its own shell, so every `origin/$BASE` ran as `origin/` and the diff commands aborted. It's five named sections now — Orient, Prepare the branch, Publish, Screenshots, What this run taught you — with the base branch and PR number substituted as literals rather than threaded through the environment. The two modes went with the numbering: a description-only ask is one line on the section that rewrites code, rather than a skip-list carried down the file.

- **The last section asks what the run should change** — did a skill mislead you, did you work around something undocumented, did the branch establish a convention. Small answers get committed onto the branch; larger ones go back to the author. Most of the commits here are that section run against this skill and the ones it calls.
- **What's left is a workflow rather than a manual.** The pre-push cleanup moved to `references/pre-push-cleanup.md` — it was half the skill and none of it was about writing a PR — and the two rules in it that govern any change moved to AGENTS.md (`class << self` past 5 class methods; never hardcoding "bike" in a translation value that means the registration's cycle type). The steps stay written for `gh`, with the sandbox that lacks it getting an appendix instead of notes interrupting the middle. The rest are corrections to its own commands: `date -v-2d` is macOS-only, a merged or closed PR was treated as editable, the frontend classifier fired on mailer and API views, and a scoped `bin/lint` silently becomes a whole-repo run when its file list comes back empty.
- **The workflow posts one kind of comment, at most once, and one skill owns it.** `github-upload-image-to-pr` finds, creates, edits and verifies the `## Screenshots` comment; the phase composes a body and hands it over, replacing a lookup implemented twice and a "run it to step 7 and stop" protocol the two files coordinated with. Everything else a run has to say goes in the body or in chat. #4126 is the worked example of both halves going wrong — a substitute artifact posted under its own heading, then a second comment reporting that the first still held.
- **The skills it leans on took the fixes exercising them turned up.** `frontend-screenshots` captures full page at both viewports, names the seeded member login that exists (`member@brakebills.edu`), declares the Playwright tools it drives, marks the base-ref checkout as the one sanctioned exception to never-change-branch, and now owns the `tmp/` filename rule that `frontend-conventions` was being triggered for. `github-upload-image-to-pr` uploads in one call and verifies against the rewritten `private-user-images` src, and the two hand the browser between them rather than each closing it. `integration-testing` and `sandbox-test-setup` stop charging a Selenium boot and chasing a chromedriver this repo hasn't used since it moved to `capybara-playwright-driver`.

`skill-creator`'s validator then found four descriptions over the 1024-character limit, all for the same reason — body content in the trigger field. `sandbox-test-setup` was the worst because it was describing three skills, so it's now a selector plus one reference per environment, 487 lines down to 60.

The screenshot phase is exercised, against this branch's own non-frontend diff: the `## Screenshots` comment below was captured, uploaded and posted through it, and `-F body=@file` was confirmed by editing that comment in place.


sethherr:
## Screenshots

This branch changes only skill docs, so there is no visual change to compare — these captures exercise the screenshot phase itself (a full page and a component preview, signed out).

### /

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/page_block/review_app_banner/component/superadmin_signed_in

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4126, opened 2026-08-15
Move the page skeletons into PageBlock::MainContent components

sethherr:
The layout wrapped every page in one of four `views/shared/_*_skeleton` partials, picked by `ApplicationHelper#current_page_skeleton` — which returned a partial *name* for the layout to interpolate into a render path. Those partials are now components under `PageBlock::MainContent`, named for the `<main id="main-content">` they fill, and a wrapper component picks between them.

- **`Wrapper.kind` holds the route → wrapper mapping**, and `Wrapper#call` renders that wrapper around the page — or the page bare, for the routes that lay out the full width themselves. `body_class` and both layouts' organized checks call `kind` directly rather than comparing against a partial name: they want the route decision, not a component, and the doorkeeper layout never renders one.
- **Every subject is a keyword argument**, down through each wrapper — the layout renders `Wrapper` before knowing which one it gets, so it passes all of them and only the matched wrapper's are read. That makes all four wrapping branches renderable on their own, which is how the spec covers them. `content_page_type` moves into `MainContent::Content`, its only caller.
- **The markup carries over byte-identical**, bar whitespace that doesn't render — six `<li>` reflows and three blank lines. Translations move to component sidecars under `components.page_block.main_content` (with an i18n-tasks write rule for the new sub-namespace). Of the two brakeman ignores whose `render_path` ran through `_content_skeleton`, one is re-fingerprinted; the other is dropped, because brakeman doesn't scan component sidecars and no longer resolves the model behind `_recovery_display`'s remaining render path.

Two fixes the move surfaced, both pre-existing:

- **`referral_source` returns the source** rather than `params[:source].present?` — an incoming `?source=` was making every donate link `source=true` instead of carrying the referrer through.
- **`_html` translations escape what they interpolate**, which is the other half of what Rails' `t` does. A component passing user-entered text — an organization's name, an Ascend name taken off an uploaded filename — was emitting it raw. Both `translation` wrappers hand the lookup to `ActiveSupport::HtmlSafeTranslation`, which is where Rails keeps that half. Values that are already html_safe pass through untouched, so every existing `link_to` / `tag` / `render` interpolation is unaffected.


sethherr:
## Rendered-markup parity instead of screenshots

The `/pr` screenshot phase needs the GitHub attachment uploader, which this sandbox has no browser session for. Since this diff is a move, I compared the rendered HTML directly against `main` instead — stronger evidence than a screenshot pair for a refactor, because it catches whitespace-level drift a screenshot wouldn't.

Same dev server, same empty database, fetched on each branch:

| Path | Skeleton | Result |
| --- | --- | --- |
| `/about` | content (info) | identical |
| `/where` | content (info) | identical |
| `/news` | content (news) | identical |
| `/manufacturers` | content | identical |
| `/terms` | none — renders bare | identical |

Identical after normalizing three per-request values (`csrf-token`, the `UI::Tooltip` id, mini-profiler ids) and the development-only `<!-- BEGIN app/views/... -->` template annotation, which now names the component path.

Not covered here: the organized, edit-bike and OAuth-applications skeletons all need auth and seed data. Those are exercised by `spec/requests/organized`, `spec/requests/bikes/edits_request_spec.rb` and `spec/requests/oauth`, which pass on this branch.


---
_Generated by [Claude Code](https://claude.ai/code)_

sethherr:
Since the parity check above, the branch picked up the namespace rename and the two fixes. The table still holds for those five paths — none of the three changes reaches them:

- the rename touches no markup,
- `referral_source` only runs on the why-donate post,
- the `_html` escaping only changes output where a **non**-`html_safe` value is interpolated into an `_html` key, and the content wrapper has none.


---
_Generated by [Claude Code](https://claude.ai/code)_

sethherr:
## Screenshots

One page per wrapper this PR extracts, captured on `main` and on this branch — same dev server, same seeded DB, same signed-in user (`member@brakebills.edu`).

The pairs are pixel-identical apart from the Rack MiniProfiler badge, which reports each request's timing and so differs on every load. Measured with `magick compare -metric AE`: 0–272 differing pixels per pair, every one of them inside that badge's bounding box (`/oauth/applications` desktop is a literal 0). Two captures of the *same* code differ by 250 pixels the same way, so that's the noise floor, not a change.

### /about

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/dashboard

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/65/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /oauth/applications

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4127, opened 2026-08-15
Convert the navbar into PageBlock::Navbar

sethherr:
The navbar was 200 lines inline in `layouts/application`, wrapped in a `cache` whose version string (`"main_navbar5"`) had to be hand-bumped for every markup edit. It's now three components under `PageBlock::Navbar`, and both layouts that render a navbar render them.

- **`Wrapper` owns the nav shell, the logo and the fragment cache**, and renders the menus. `logo_only: true` renders just the logo — that's the doorkeeper layout's whole call, and it skips the cache, since that markup is one link and was uncached before. Anything else must pass `page_id`, which keys the cache.
- **`PrimaryMenu` renders from a manifest of items** — `{label:, path:, active:, item_class:, html_options:}` plus `{type: :divider}` and `{type: :settings}` — rather than repeated `<li>` blocks. `MenuLink` turns one item into an anchor for both menus: `active:` absent or `:match_controller` defers to `active_link`, `true`/`false` decides it there (which is what keeps `#getStolenBackLink` inactive for `welcome/index.coffee` to set).
- **`OrganizationMenu` and `SettingsMenu` are the two dropdowns.** The former's `render?` carries the condition the layout's `if/elsif` used to encode, so the signed-out signup link is a plain `if current_user.blank?` beside it.
- **Everything is passed in**, including `controller_namespace`/`controller_name`/`action_name` — the same three `PageBlock::HeaderTags` already takes — so active state compares arguments rather than reaching into `controller`, and the specs set them instead of driving a URL. `Org::MenuItems` gets the same treatment: its `routed_controller`/`routed_action` read `controller.request.path_parameters`, and `MainContent::Organized` now threads the namespace down to it (it already took the other two).
- **`MARKUP_DIGEST` replaces the hand-bumped cache version**, the way `PageBlock::Footer` does, so editing any of the three — or `Org::MenuItems`/`UI::IconChevron` inside them — invalidates the fragment. `unregistered_parking_notification` joins the key: the org dropdown's active item depends on it and `page_id` doesn't vary on it, so the old key could serve the wrong one on `organized/bikes#new`.

Translations are per component, which needs a `write` rule in `i18n-tasks.yml` — without one `normalize` consolidates the sidecars back into the parent namespace's file.

Deliberately left alone as legacy carry-over, worth their own pass: both dropdowns still use Bootstrap's `data-toggle` rather than `UI::Dropdown`, and the settings gear is a 38-line inline SVG whose `<defs><style>.cls-1{...}</style></defs>` leaks a global class name (`payments_layout` defines a conflicting one). Also left: `page_id` in the cache key is what makes it one entry per page per user, and only ~6 `active` classes vary by page — dropping it (the way `#getStolenBackLink` already does with JS) would collapse that, but it's a behavior change, not a move.


sethherr:
## Screenshots

The navbar is pixel-identical to `main` in every state below — diffing each pair, the only differences are the homepage's randomized hero cards and the mini-profiler's timing badge. Captured signed in as the seeded `member@brakebills.edu` (Brakebills member) for the menu states.

Re-verified after merging `main` (which brought in #4126): the merged branch still matches `main` on `/news` to within the profiler badge. These images predate that merge, but nothing in it moved the navbar.

### /

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### / — settings menu open

The `settings_items` manifest: the org link, registrations, register, the user-settings link with its `data-email`, a divider, logout. On mobile the gear is hidden and these render inline in the hamburger menu, so that's the mobile shot.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### / — organization dropdown open

`Org::MenuItems` inside the navbar's cached fragment — the part `unregistered_parking_notification` now keys. The pill is desktop-only; its "View Brakebills" link appears in the mobile menu above.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /news

Blog is the active link here — the item that moved from `match_controller: true` to `active: :match_controller`.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/page_block/navbar/component/logo_only

The doorkeeper layout's nav, which the OAuth authorization prompt renders. New preview, so there's nothing on `main` to compare against.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4128, opened 2026-08-15
Pass controller state into view components as arguments

sethherr:
`Admin::IndexSkeleton` read fourteen controller instance variables through `controller.instance_variable_get`, and eleven components included `Binxtils::SortableHelper` to reach `sortable_search_params` off the controller's `params`. Everything a component needs is now a constructor argument, so a component renders from what it's given and nothing gathers assigns on its behalf. Builds on #4177, which split out the pieces that stood alone.

- **Two value objects carry the state that's identical on every page.** `ComponentStates::IndexState` is the request an admin index knows about itself; `ComponentStates::SortState` is how a table is sorted and what its links have to preserve. Both are built once in `ControllerHelpers` and handed over whole, so each index view passes one of them plus what's actually its own — the collection, what it's called, its table.
- **The cell components resolve search params once per request rather than once per row.** `Binxtils::SortableHelper` memoizes them per *instance*, and `Admin::{Users,Bikes,Organizations}::Cell` are built once per row — a 100-row admin index was running `params.permit` a couple of hundred times per render.
- **Four helpers go.** `render_admin_current_header` and `render_admin_pagination_with_count` wrapped a component in a helper only to read ivars on its behalf, so they're inlined at their call sites; `humanized_time_range` and `humanized_time_range_column` now take a `period:` rather than reading `@period` off whatever they're mixed into, which is what lets `Admin::PaginationWithCount` format its own range.

One behavior change worth a look: `?show_serial=` and `?search_manufacturer=` were read from `params` by every `Admin::Bikes::Table` render, and are now passed from `admin/bikes/index` — the only page those params exist on.

`frontend-conventions` gets the rule this follows, and the distinction that matters: what's banned is a component reaching out, not the number of arguments it takes, so cohesive controller-owned state can be named and passed whole.


sethherr:
## Screenshots

Three admin indexes at `?period=week` and two org pages, full page at both viewports. This is a refactor with no intended visual change, so the pair is the point: `magick compare -metric AE` reports **0 differing pixels** on all ten.

### /admin/bikes?period=week

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/users?period=week

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bug_reports?period=week

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4130, opened 2026-08-16
Drive the navbar from stimulus, with UI::Button toggles

sethherr:
The pass #4127 deferred: the navbar's two dropdowns were still `<a href="#">` driven by Bootstrap's `data-toggle`, and the menu was still `nav_header.coffee`. The navbar renders pixel-identical either side of this, bar the focus ring, which is `UI::Button`'s now.

- **Both toggles are `UI::Button::Component`.** `revised.scss` sits in `@layer legacy`, so a component's utilities outrank every nav rule it lands in — the toggles are styled on the component, and the rules that used to reach them are deleted rather than left reading as live. `$grid-breakpoint-lg` gets a tailwind name (`--breakpoint-navbar`) rather than a literal media query, since `lg:` is 1024px and the navbar switches at 992px.
- **One Stimulus controller replaces `nav_header.coffee` and Bootstrap's dropdown data-api**, toggling the same classes the stylesheet already keys off and handing it the open menu's offset as `--navbar-bottom`. That fixes the navbar going inert after a Turbo visit: `initializeBikeIndex` runs the full legacy init once per JS context, so a Turbo navigation to a new page id left the hamburgler unbound and unlabelled until a reload.
- **The navbar's browser coverage is one spec now.** `mobile_nav_menu_spec` picks up the component system spec's cases and opens each dropdown for real, which the Lookbook preview can't do. `registrations_search_spec`'s access-panel assertion is scoped to the panel — unscoped it was matching the navbar's organization name against the panel's heading, across the gap the unbound hamburgler left.

The homepage's category buttons carried their own label in the image `alt` (`alt="Bike Shops"` inside a link reading "Bike Shops"), which the navbar spec's axe audits had to skip past — they're `alt=""` now, so the landing page is audited whole.

`signup_spec` and `registration/edit_spec` were failing intermittently in the parallel suite, for a race unrelated to the navbar: `init.coffee` constructs its per-page class in `$(document).ready`, but `click_link` returns with the new document still parsing, so an interaction landing in between is swallowed silently — a dropped click reads as "the reveal never rendered". `wait_for_page_script` waits on `window.pageScript`, the legacy counterpart to `wait_for_stimulus`.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Both pairs are pixel-identical (`magick compare -metric AE` → 0) — this page carries both dropdown toggles at desktop width and the hamburgler at mobile. The homepage was captured too and its navbar strip is identical as well; it isn't posted because the hero content below the nav rotates per page load.

PR #4131, opened 2026-08-16
Rebuild the organization menu as a sidebar, replacing the navbar

sethherr:
Rebuilds the organization menu as the design's left sidebar, rendered in place of the navbar on every `layouts/application` page for a reader with a passive organization. Everyone else keeps the navbar unchanged.

- **`PageBlock::Navbar::OrgSidebar` is the whole nav for those readers**, so it carries their account links too. Every collapsed and mobile style is a tailwind variant on `data-collapsed` / `data-mobile-open`; its Stimulus controller sets those two and the width the content column is offset by, leaving the account menu to `UI::Dropdown` and the groups to `ui--collapse`. Which row is current is `UI::ActiveLink`'s, answered in the browser since #4140 — the controller opens the group holding it, flags that group, and scrolls it into view when the menu is taller than the column.
- **`OrganizedServices::UserMenuItems` owns every path, feature gate and match**, now grouped the way the design lays them out rather than flat. `api/v3/me` serves the same payload, so its shape changes with it.
- **Both old renderings go** — the navbar dropdown and the organized-only left menu — along with `Org::MenuItems`, the styles and Stimulus behind them, and the per-page arguments the navbar no longer varies by. The content column is the window's height, so a short page doesn't end partway down the menu.
- **Riding along, unrelated**: a location created through the org locations form saved without coordinates, because `AddressRecord` geocodes in its own `after_validation` — after `Location#set_calculated_attributes` has already copied them — so the organization never got `search_coordinates_set?`. Happy to split it out.

The design's rows map onto what the old menu already reached: Messaging is the custom emails index, falling back to the stolen message's edit page for an org that has only that one; Reports is the overview dashboard. Add an Impounded Vehicle is the one row left greyed, since `organized` routes no `impound_records#new`.

Worth a reviewer's call: the sidebar's account dropdown carries four rows, so a reader with a passive organization no longer gets the two `Navbar::SettingsMenu` also builds — the per-organization switcher and marketplace messages — anywhere on the site. Restoring them means sharing that menu's manifest rather than the sidebar declaring its own.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /my_account

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4132, opened 2026-08-16
Add a placeholder to register step 2's owner name field

sethherr:
Step 2's owner name input had no placeholder, so nothing signalled it wants a full name rather than a first name or a nickname.

- **`First & Last name` on the `user_name` field**, which only renders when `user_name_required?` — a registration made on someone else's behalf, or one whose owner email isn't the signed-in user's. The `register/step2` `for_someone_else` preview is where it shows.
- **`frontend-screenshots` stops cropping tall previews.** Capturing that preview is what turned it up: its blanket `fullPage: false` for preview URLs cut off the field this PR changes, which sits at ~1120px.


sethherr:
## Screenshots

### /rails/view_components/register/step2/component/for_someone_else

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4133, opened 2026-08-16
Replace the active_link helper with UI::ActiveLink

sethherr:
`ApplicationHelper#active_link` appended `active` to a link's class when it pointed at the page being rendered. It's `UI::ActiveLink::Component` now, at all 39 call sites, with a preview per variety.

- **`match:` says what counts as "the page it points at"** — `:path`, `:controller`, or `:controller_action` — replacing the `match_controller:` boolean. `active:` is the other half: a caller that already knows the answer passes it rather than being routed around the component. Together they let `PageBlock::Navbar::MenuLink` and `Org::MenuItems` drop the "does the component decide, or do I?" branch each had grown, and let `PrimaryMenu` delete `search_active?` and its three constructor arguments — its search links carry query params, so they match on route.
- **The cache digests move because they finally cover these links.** `calculated_markup_digest` follows `render X::Component` and can't see into a helper, so editing a footer link's markup used to serve stale cached footers. `bin/update_component_digests` regenerates the two constants.
- **Link text is escaped rather than passed through `raw`.** 

No link changes when it's active: rendering eleven URLs on this branch and on `main` — including `/search/registrations` at each stolenness, with a query, and on page 2 — produces the same active links on both.

Also removed `admin_nav_link`, a dead 18-line sibling of the deleted helper, and renamed `Admin::Navbar`'s private `active_link` (it returns a nav manifest entry, not a link) to `current_nav_link`.

### An unrelated test fix, found while verifying this

CarrierWave's cache was namespaced per parallel worker but its `root` wasn't, so `store_dir`'s `uploads/Pu/<id>` was shared by all 8 workers *and* across runs — `public/uploads/Pu` had accumulated 3,438 stale `small_*` files. A worker whose new record landed on an occupied id saw another run's file, which is what failed `carrier_wave_process_job_spec`'s "the small version doesn't exist yet" assertion. Each worker now gets its own root, with a matching `base_path` so rendered URLs still point at the file, and drops it after the suite. The full suite goes from 2 failures to 0.

Left alone: `layouts/application_bikehub` still hand-rolls an active class off `controller_name` for its signup/login pair. It's the same idea but it wasn't an `active_link` caller, and both links carry a `partner:` param, so it wants its own look.


sethherr:
## Screenshots

### /about

Renders the navbar, the content nav and the footer — 36 of the 39 converted links. `magick compare` reports 0 differing pixels at both viewports.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4136, opened 2026-08-17
Resolve the navbar's active link in the browser, not per page

sethherr:
#4127 left `page_id` in the navbar's fragment cache key, making it one entry per page per user for the sake of a handful of `active` classes. Those resolve in the browser now, so the cache is one entry per user and organization.

- **Links carry the rule, the body carries the page.** A link ships `data-active-path` or `data-active-routes="controller#action"`, the layout stamps `<body data-current-route>`, and `page-block--navbar` marks the match on connect, on `turbo:frame-load` and on `popstate`. `NavRoute` holds both halves, and the uncached org sidebar evaluates the same rule server-side rather than keeping a second one.
- **`route_override_key` replaces `page_id` in the key.** `Org::MenuItems` injects a link on three pages when the org lacks that feature — markup JS can't recover — so the key carries just those, and is nil everywhere else.
- **`welcome/index.coffee`'s `updateActiveLink` is gone**, along with the `active: false` + `#getStolenBackLink` pair it existed to patch; that link is plain `:auto` now, and the two search links swap hand-computed booleans for `:match_controller`.

Worth knowing before it spreads: the active link is unmarked until Stimulus connects, and the matcher now exists in both Ruby and JS. `PageBlock::Footer::Component` has the identical shape and is deliberately untouched, because whether these fragment caches earn their key is worth settling first — the org dropdown is where the render cost sits, and it wants measuring on a real request against production-shaped data rather than a component benchmark.


sethherr:
## Screenshots

Pixel-identical to the base at both viewports (`magick compare -metric AE` = 0 on all four pairs) — the active link lands in the same place, it's just resolved in the browser now.

### /news

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

sethherr:
Doing differently

PR #4139, opened 2026-08-18
Inherit the surrounding color in .twlink-underlined

sethherr:
Visiting a hot sheet without a base location flashed "Organization must have a location set to enable Hot Sheets" and then rendered the edit page's own alert saying the same thing. Fixing that ran into the other half: a link inside a colored alert came out legacy blue rather than the alert's color.

- **Admins get the alert alone.** The show guard still redirects them to edit, which renders it; members, redirected to the org root where there's no alert, keep the flash. That alert is `UI::Alerts::Base` now rather than hand-rolled `.alert.alert-warning` markup.
- **`.twlink-underlined` inherits the color it sits in**, which takes `text-inherit` — this layer beats legacy only where it declares the property, so `a { color: #3498db }` was supplying the color otherwise. Every link inside an alert uses it, replacing the `.twalert-body .twlink` override, as do the review-app banner's three. Two admin tables move their color off the anchor onto the wrapping element, since a `tw:text-*` there outranks the components layer at every state.
- **Its hover shifts that inherited color and thickens the underline.** The shift is clamped in oklch, so text already at the floor — the banner pins black on a bright accent — lifts rather than hovering to itself, and no surface needs a hover class of its own.
- **The `tw*` classes move to `bike_index_components.css`**, imported where the `@layer components` block sat, which is what keeps the compiled `tailwind.css` byte-identical. Rules that style one component's markup (`.ui-table*`, `.search-results-frame-wrapper`) stay behind in `application.css`.


sethherr:
## Screenshots

### /o/bike-recovery-team/hot_sheet

Redirected to the edit page by the show guard, as an org admin of an org with no location set.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/alerts/base/component/error_with_header

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/page_block/review_app_banner/component/with_pr_link

Unchanged at rest — the banner's links swapped `tw:text-inherit tw:underline` for the shared class, and what differs is the hover.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4140, opened 2026-08-18
Resolve UI::ActiveLink's current page in the browser

sethherr:
The navbar and footer render from fragment caches keyed on `page_id`, so the link they marked current was whichever page filled the cache — `PrimaryMenu` pinned "Stolen bike?" inactive and `welcome/index.coffee` re-marked it by hand on that one info page. `UI::ActiveLink` now renders what to compare and lets `ui/active_link_controller.js` set `aria-current` against the page, off a `<body data-page-route>` that `body_tag` gives every layout.

- **`match:` is the only control — there's no way to pass the answer in.** `:full_path` counts the link's query string, which tells the org menu's two `organized/bikes#new` links apart; `matching_controllers:` extends a match across a second controller, keeping "Manage Registration sequences" current while a page of one is edited.
- **Neither cache key carries `page_id`.** The footer's is the user, organization and pixel flag — its language switcher submits an action-less form the browser aims at the current URL, and `default_url_options` stops forwarding a `locale` that wouldn't change what renders. The navbar's one page-dependent part is the org menu, which renders outside the cache, leaving its key the user alone.
- **The org menu manifest carries `match:`** in place of the `:on_bikes_new`-style markers the component resolved per request, dropping `unregistered_parking_notification` from five components. Its cached payload — and the API v3 `me` menu it's serialized into — moves to v2.
- **`Admin::Navbar` keeps a server-side check**, since its picker names the current page in prose rather than marking a link. It calls `current_page_active?` directly, so the component offers no answer at all.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -73,7 +73,7 @@ The same instinct applies beyond buttons: **check `app/components/ui/` and `app/
 
 ## Current-page links: always `UI::ActiveLink`
 
-**Every link that goes `aria-current` on the page it points at goes through `UI::ActiveLink::Component`** — never `current_page?` in a template, and never a Stimulus controller of your own comparing `window.location`. It always derives the state in the browser (`app/javascript/controllers/ui/active_link_controller.js`), so there's no way to pass the answer in: `match:` is the only control over what counts as the page it points at. Any layout rendering one needs `<%= tag.attributes(body_attributes) %>` on its `<body>`. See `app/components/ui/active_link/`.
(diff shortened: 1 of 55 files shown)


sethherr:
## Screenshots

Every pair is pixel-identical — the marker moves from the server to the browser without changing what renders.

### /help

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/registrations?query=trek

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4141, opened 2026-08-18
Flag UI::ActiveLink's current page with aria-current

sethherr:
`UI::ActiveLink` (#4133) marked the current page with an `active` class, but the `is-active` variant that every other selected button, menu item and toggle styles through already matches `[aria-current]` — which is also what a screen reader wants on a nav link. The component sets that instead.

- **The legacy nav, org menu and footer rules move to `[aria-current]`**, and `welcome/index.coffee` sets the attribute on the one link the cached navbar leaves to JS.
- **`.twlink` and `.twlink-underlined` state their active state as the variant**, rather than repeating a marker list that was already drifting from it — it covers the `:active` pseudo and the disabled guards those rules carried separately. `UI::TableColumn`'s sort link flags itself `data-active` to keep reaching them, and the underlined hover's oklch shift moves to a custom property so both states share one formula.
- **The admin navbar drops `.nav-link` for utilities.** Its display, padding and color came from the vendored bootstrap dump, which isn't ours to edit and keys the active color off the class; two constants carry the same declarations.
- **The news sidebar's hand-rolled active link becomes an `ActiveLink`**, passing `active:` for the `sortable_search_params?` condition the component can't derive.

### DIFF
--- a/app/assets/javascripts/revised/pages/welcome/index.coffee
+++ b/app/assets/javascripts/revised/pages/welcome/index.coffee
@@ -24,8 +24,8 @@ class BikeIndex.WelcomeIndex extends BikeIndex
 
   updateActiveLink: ->
     if window.location.pathname.match("info/how-to-get-your-stolen-bike-back")
-      $(".primary-nav-item .active").removeClass("active")
-      $("#getStolenBackLink").addClass("active")
+      $(".primary-nav-item [aria-current]").removeAttr("aria-current")
+      $("#getStolenBackLink").attr("aria-current", "page")
 
   # Return the DOM node for the slide with index `index` in the slides
   # container element.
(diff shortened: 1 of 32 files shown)


sethherr:
## Screenshots

Every pair should look the same — the markers change, not the styling. The navbar's active item, the footer's, the info sidebar's underline, the org menu's arrow, and the admin navbar's darker shortcut are all pixel-identical to `main` (the admin table's `Added` header is the sortable link now flagged `data-active`).

### /help

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/dashboard

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4142, opened 2026-08-18
Move .twtext-color to bike_index_components.css as a utility

sethherr:
`.twtext-color` and `.twtext-color\!` were the last `tw*` classes left behind in `application.css`, sitting unlayered under a TODO asking for the body to use them.

- **It's one `@utility` now rather than two classes.** Tailwind v4's `@apply` only takes utilities, and `base-font` applies this one, so the default text color is stated once; the `!` variant comes from Tailwind rather than a hand-escaped `.twtext-color\!` rule. Markup reads `tw:twtext-color`. As a utility instead of an unlayered rule it now loses to a `tw:text-*` on the same element rather than beating it — none of the four call sites carry one.
- **The `body.dark` block goes with it.** Its text color duplicated what `base-font` already emits, and "'dark:' syntax isn't working for the body element" was a v3 artifact: that mode generated `.dark body`, which can't match a `<body class="dark">`. v4's `&:where(.dark, .dark *)` does, so the background folds into the body rule.
- **The stolen checklist's eight links take `.twlink-underlined`**, the inherit-the-surrounding-color class from #4139. Nothing styled them before — the pages that render it load only `tailwind` and `revised`, neither of which colors a bare anchor — so "add a photo of your bike" read as plain text.

Compiled `tailwind.css` is the same but for those two folds, 17 bytes larger. Five `MARKUP_DIGEST` constants move because the renamed class sits in markup they cache.

Also: `frontend-conventions` gains the `tw:twtext-color` bullet and the rule behind the prefix — a class that anything `@apply`s has to be an `@utility`. And AGENTS.md's testing paragraph points verification at scoped `bundle exec rspec` instead of `bin/ci`, which is what the `rspec-testing`, `fixing-flaky-failures` and `pr` skills already said.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -33,7 +33,7 @@ Scope it rather than running bare `bin/lint`: a whole-repo run reformats files o
 - Labels should use the `twlabel` class.
 - Basic links should use the `twlink` class.
 - A link that should take the color of the text around it — inside an alert, a colored `<small>`, the review-app banner — uses `twlink-underlined` instead. Never hand-roll `tw:text-inherit tw:underline`. Its color must come from an ancestor, not a `tw:text-*` on the link itself, which would outrank the class at every state.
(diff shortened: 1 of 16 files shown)


sethherr:
## Screenshots

### /rails/view_components/bikes/stolen_checklist/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/page_block/marketplace_listing_panel/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4144, opened 2026-08-18
Merge the footer's FAQ and Help links, and fix the pagination hover

sethherr:
The footer's Bike Index column listed **FAQ** and **Help** separately, but `/support` redirects to `/help` — both links went to the same page.

- **They're one "FAQ & Help" link now.** The four sync-generated `translation.*.yml` files carry the new key inline, since the footer component spec renders in `:nl` under `raise_on_missing_translations` — a rule `AGENTS.md` now states.
- **"Register a bike" matches on controller**, so the footer flags it active across the registration flow rather than only on `/register/new`.
- **`UI::Pagination`'s hover stops underlining and keeps its border.** Bootstrap's reboot underlines every `a:hover` inside a `.container`, and `-space-x-px` overlaps each item's border with its neighbor's, so the hovered link's edge stayed gray under the item after it. The current page takes the same `z-10` — its purple border was covered the same way. The gap's `px-2` is prefixed too; unprefixed it was reaching bootstrap's, which only exists on a page that loads `revised`.
- **CI's translation sync has been skipping `main` since `f71c7a127`** — the container refuses the workspace as "dubious ownership", so `git branch --list` printed nothing and `bin/check_translations` read that as "some other branch" and exited 0. Last automated sync PR was #3783, on June 30. It sets `safe.directory` under `$CI`, and the branch now comes from an assignment, so a git that can't answer aborts instead of skipping green.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,8 @@ A registration is as often an e-scooter, a stroller or a wheelchair, so **never 
 
 Run `bundle exec rails prepare_translations` after hand-editing a `component.en.yml`; `bin/lint` doesn't normalize YAML.
 
+A spec that renders in another locale — `spec/components/page_block/footer/component_spec.rb` does, in `:nl` — fails on a key the sync-generated `config/locales/translation.*.yml` don't carry yet, since `raise_on_missing_translations` is on in test. Add the key inline to all four; the next sync overwrites them.
+
 ## Subagents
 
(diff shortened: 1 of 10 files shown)


sethherr:
## Screenshots

### Footer, desktop

| `origin/main` | this branch |
| --- | --- |
| footer on main, desktop | footer on this branch, desktop |

### Footer, mobile

| `origin/main` | this branch |
| --- | --- |
| footer on main, mobile | footer on this branch, mobile |

### `UI::Pagination` on `/o/brakebills/registrations`, page 2 hovered

`origin/main` — underlined, and the hovered border's right edge sits under the next item:

pagination hover on main

this branch:

pagination hover on this branch

PR #4145, opened 2026-08-18
Make the marketplace result view switcher work

sethherr:
The marketplace's **View as List** / **View as thumbnails** switcher did nothing. Its radios are fields in the search form, so picking one only checked the radio — the value sat there until the rider happened to press search.

- **Changing the result view submits the form**, which re-runs the search through the results frame, so the layout switch keeps the current filters and lands in the URL. The counts fetch skips `search_result_view`, since the kind counts can't vary by layout.
- **The radios are `sr-only` rather than `hidden`.** `display: none` kept them out of the tab order entirely, so the switcher — and the change handler this now depends on — never reached a keyboard.
- **The CI image bakes `gh`.** Now that #4144 has the translation sync running on `main` again, it pushed a branch and then died on `gh pr create` — the `test` job moved into a container in `f71c7a127`, and that image had no `gh`, which is otherwise what this repo's CI speaks throughout.


sethherr:
## Screenshots

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/marketplace?search_result_view=bike_box

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4149, opened 2026-08-19
Prefix support for UI::Forms::Input, and two admin fixes

sethherr:
The admin organization form's permitted-domain field wrapped its input in Bootstrap `input-group` markup, but `.twinput` is `w-full block` — so the `@` sat on a line of its own above the field. Two other admin bugs found alongside it.

- **`UI::Forms::Input` takes a `prefix:`**, rendered against the field's left edge. The org form passes it a prefixed input as its `UI::Forms::Group` content block, in place of the input-group markup.
- **A blank `short_name` falls back to the org name.** The new-organization form posts `short_name: ""`, so the `|| name` fallback never ran — the slug came out empty and `OrganizationNameValidator` rejected it as "Invalid name", pointing at the name the admin *had* filled in.
- **Admin's delete-user button is a `UI::ButtonLink`.** `method: :delete` was passed inside the URL helper, so it became a query param: the button was a plain GET to `/admin/users/:id`, which redirects to edit. Nothing was deleted, and the user kept a working session.

### DIFF
--- a/.claude/skills/rspec-testing/SKILL.md
+++ b/.claude/skills/rspec-testing/SKILL.md
@@ -51,6 +51,10 @@ What a component decides about its markup — a label, a placeholder, a class, w
 
 A component having no spec yet isn't a reason to put it in the request spec instead. `spec/components/register/step2/component_spec.rb` is the pattern, including the `render_x` helper that reloads the record so an object updated mid-example isn't answered from the copy the previous render left behind.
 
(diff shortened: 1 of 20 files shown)


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/users/3/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4150, opened 2026-08-19
Render every serial through Atom::Serial

sethherr:
`BikeHelper#render_serial_display` and `Atom::Serial::Component` rendered the same markup, so the helper, its `helpers.bike_helper` translations and all nine of its call sites are gone.

- **The component takes a raw `serial:` string and an `html_class`**, the way `Atom::ShortId` and `Atom::Sticker` take a record or a string, so the org multi-search heading and its result chips render through it rather than writing `.serial-span` themselves. The chip's color and underline move onto that span — `.serial-span` sets its own color and weight, so they can't reach the text from the `<a>` — while its padding stays on the link, which is the badge's flex item.
- **A sticker search echoes its codes through `Atom::Sticker`**, which is the only rendered change here: a code searched in multi search used to read as a serial, and now looks like it does everywhere else. The atom takes an `html_class` for the chip link.
- **A placeholder word only stands in for a bike's serial.** `unknown` typed into multi-search is the number being searched for, not the absence of one, so the raw path echoes what was typed.
- **`.serial-span` stays in `application.css`**: `multi_search_controller.js` builds the placeholder chips shown before results arrive. Nothing else writes the class now.

The two `MARKUP_DIGEST` bumps are the spec-enforced kind.


sethherr:
## Screenshots

Component previews, since the changed markup is the serial atom and the multi-search result chip.

### /rails/view_components/atom/serial/component/hidden

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/org/search_results/multi_result_chip/component/one_result

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |


### /rails/view_components/org/search_results/multi_result_chip/component/sticker_code

New preview scenario, so branch only — this is the sticker-code chip the search now renders.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4151, opened 2026-08-19
Render the organization bike list with UI::Table

sethherr:
The organization registrations page has two tables behind it. Orgs with `bike_search` get `Org::SearchResults::BikesTable`, which is already on `UI::Table`; orgs without it — `cannondale`, say — fell through to `organized/bikes/_list`, a hand-rolled Bootstrap table and the last one on the org side.

- **The list renders through `UI::Table::Component`**, with the partial converted from haml to ERB. Same columns and the same feature gating; the cell blocks take their caller state through locals, since `UI::Table` `instance_exec`s them against itself.
- **The claimedness dropdown moves to `UI::Dropdown`**, sticker codes to `Atom::Sticker` and the sticker-assign link to `UI::ButtonLink`, which retires the page's last Bootstrap `data-toggle` and its last rails-ujs `method:`. `Stickers` was the one untranslated column header, so it has a key now.
- **The claimedness links carry `sortable_search_params`**, so picking a filter no longer resets the period select sitting above it.


sethherr:
## Screenshots

### /o/cannondale/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

`claimed_ownerships` and `bike_stickers` were enabled on the seeded `cannondale` org for both captures, so the dropdown and sticker columns are visible — the org has neither by default.

PR #4154, opened 2026-08-19
Render the organization alerts with UI::Alerts

sethherr:
Every Bootstrap `.alert` in the organization interface was hand-rolled markup rather than the alert component — seven of them, across the graduated notifications, parking notifications, custom emails and user invite pages.

- **They render through `UI::Alerts::Base::Component` now.** The three that carried a heading pass it as `header:`; links inside take `twlink-underlined`, and the bodies drop the Bootstrap `text-warning`/`text-danger`/`text-muted`/inline-grey that would otherwise override the colour the alert had just set.
- **The hot sheet's recovery banner is a `UI::Badge` instead.** `.recovered-alert` was an alert only in name — a green strip inside each row of the bike list, wearing an `<h3>`. Its SCSS is deleted; the card already carries a "Recovered" row below it.
- **`UI::Header` takes block content and h4–h6.** Groundwork for the rest of the organization headings, so no caller in this diff — but a heading that wraps a link or a conditional `<small>` can't pass its markup as a `text:` string, and `TAG_CLASSES` stopped at h3, so an h4 rendered at h1 size. Passing neither text nor content now raises instead of emitting an empty heading.
- **None of the eight branches had coverage.** Every one sits behind a conditional, so the request specs passed without ever rendering them. Two took real work to reach: the repeat-failure alert's ids round-trip through the session, and the not-normally-viewable email alert only renders for a superuser, since `find_mail_snippets` swaps the kind out for anyone else.

`.hot-sheet-notification-panel*` keeps its own SCSS — those panels also size the form they wrap, so converting them is more than a container swap.


sethherr:
## Screenshots

### /o/brakebills/graduated_notifications

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

`brakebills` had its `graduated_notification_interval` cleared for both captures, since that's what this alert is gated on.

The other six converted alerts each sit behind a conditional rather than at a plain URL — session state set by a POST, a notification sent before its org's email snippet was edited, an org with a permitted email domain, a superuser viewing an email kind the org can't. They're covered by request specs instead of screenshots.

PR #4161, opened 2026-08-19
Give review apps a SAML SP keypair, and fix the metadata endpoint

eld120:
SSO can't be exercised on a review app or on sandbox today: `Saml::SettingsBuilder` reads `SAML_SP_CERTIFICATE` / `SAML_SP_PRIVATE_KEY` straight from `ENV`, and `config/deploy.review.yml` never declared either. Both are `nil` there, so AuthnRequests go out unsigned and the IdP rejects every login with `SigAlg was null`. This wires the keypair through, and fixes the two things an IdP admin hits first.

- **The keypair travels base64-encoded.** Kamal writes each secret to the host as one `KEY=value` line, which a multi-line PEM breaks. `pem_from_env` decodes it and still accepts a raw PEM, so `.env.local` is unchanged; `saml:generate_sp_keypair` now prints the one-line form alongside the PEM. Decoding is strict — a mangled certificate otherwise drops the `<KeyDescriptor>` from our metadata with no error anywhere.
- **Metadata is served as `application/xml`, at one url.** The registered `application/samlmetadata+xml` is an unknown type, so browsers downloaded it instead of showing it — an IdP admin handed the url got a file. The path takes no extension either: it *is* the entityID, so `/metadata.xml` 404s rather than serving the same document under a second name, and the admin card links exactly the string the document declares. The other SSO routes stop accepting an extension too.
- **`/sso/<slug>/sp.crt` serves the certificate on its own**, as `application/pem-certificate-chain`, for IdP tooling that registers a bare certificate rather than reading one out of metadata. Same content-type-and-url-names-the-file approach the university IdPs use. It's the app-wide keypair, so every organization serves the same bytes.

Each environment gets its own throwaway keypair — an IdP trusts whichever certificate our published metadata carries, so a review app's needn't match production's, and production's private key has no business on the shared review host. The two 1Password fields are read apart from the mandatory `kamal secrets fetch`: kamal runs its own secrets commands in-process, so a field missing from the item would raise out of `.kamal/secrets` and take down every kamal command — `console`, `logs`, `dbc`, `destroy` — rather than just SSO. An item without them now loads fine, and a review app deploys normally with only the SSO endpoints 404ing.

**Before this deploys anywhere**, two secrets have to be added to the `review-app` GitHub Environment: `REVIEW_APP_SAML_SP_CERTIFICATE` and `REVIEW_APP_SAML_SP_PRIVATE_KEY`. Without them the app boots fine and SSO fails at the first login.


.kamal/secrets
```diff
@@ -9,7 +9,7 @@
 
 KAMAL_REGISTRY_PASSWORD=$(gh auth token)
 
-SECRETS=$(kamal secrets fetch --adapter 1password --account bike-index --from "Kamal/BikeIndex Review" POSTGRES_PASSWORD SECRET_KEY_BASE SESSION_SECRET VERIFICATION_SECRET GOOGLE_MAPS GOOGLE_MAPS_STATIC GOOGLE_GEOCODER MAPBOX_GEOCODER MAPBOX_MAPPING R2_DEV_ENDPOINT R2_DEV_ACCESS_KEY R2_DEV_ACCESS_KEY_SECRET HONEYBADGER_API_KEY HONEYBADGER_FRONTEND_API_KEY)
+SECRETS=$(kamal secrets fetch --adapter 1password --account bike-index --from "Kamal/BikeIndex Review" POSTGRES_PASSWORD SECRET_KEY_BASE SESSION_SECRET VERIFICATION_SECRET GOOGLE_MAPS GOOGLE_MAPS_STATIC GOOGLE_GEOCODER MAPBOX_GEOCODER MAPBOX_MAPPING R2_DEV_ENDPOINT R2_DEV_ACCESS_KEY R2_DEV_ACCESS_KEY_SECRET HONEYBADGER_API_KEY HONEYBADGER_FRONTEND_API_KEY SAML_SP_CERTIFICATE SAML_SP_PRIVATE_KEY)
```
sethherr: `SAML_SP_CERTIFICATE` / `SAML_SP_PRIVATE_KEY` are in the mandatory fetch, so if the `Kamal/BikeIndex Review` item doesn't have those two fields, `kamal secrets fetch` exits non-zero, kamal raises on the failed inline substitution, and the whole secrets file fails to load — taking out `bin/kamal_review destroy`, `console`, `logs` and `dbc` along with `deploy`.

Splitting them into their own fetch that's allowed to fail keeps the blast radius to SSO:

```suggestion
SECRETS=$(kamal secrets fetch --adapter 1password --account bike-index --from "Kamal/BikeIndex Review" POSTGRES_PASSWORD SECRET_KEY_BASE SESSION_SECRET VERIFICATION_SECRET GOOGLE_MAPS GOOGLE_MAPS_STATIC GOOGLE_GEOCODER MAPBOX_GEOCODER MAPBOX_MAPPING R2_DEV_ENDPOINT R2_DEV_ACCESS_KEY R2_DEV_ACCESS_KEY_SECRET HONEYBADGER_API_KEY HONEYBADGER_FRONTEND_API_KEY)

# Optional — a review app without an SP keypair deploys fine, it just 404s the SSO
# metadata and sp.crt endpoints. Fetched apart from SECRETS so a missing 1Password
# field can't take down every kamal command.
SAML_SECRETS=$(kamal secrets fetch --adapter 1password --account bike-index --from "Kamal/BikeIndex Review" SAML_SP_CERTIFICATE SAML_SP_PRIVATE_KEY || echo "{}")
```

sethherr (commented):
Following up on the review — one thing worth changing before this merges.

These two are the only secrets here whose absence is fatal to **everything** rather than just to SSO. Nothing else in the change needs them to exist:

- `Saml::SettingsBuilder` reads them through `.presence`, so nil → the metadata and `sp.crt` endpoints 404
- `.kamal/secrets-ci` substitutes an unset `$SAML_SP_CERTIFICATE` to empty, so CI without the `REVIEW_APP_SAML_SP_*` environment secrets is fine
- `env.secret` in `config/deploy.review.yml` only needs the key present, not non-blank

Only the 1Password fetch in `.kamal/secrets` hard-requires them. Three suggestions below make that tolerant.

Tradeoff to weigh: a review app then deploys silently keypair-less, and SSO there 404s in a way that reads like a bug rather than like missing config. If you'd rather keep it loud for `deploy` but quiet for `console`/`logs`/`dbc`, that's a check in `bin/kamal_review` instead and I'd do it there.

.kamal/secrets
```diff
@@ -35,3 +35,7 @@ R2_DEV_ACCESS_KEY=$(kamal secrets extract R2_DEV_ACCESS_KEY ${SECRETS})
 R2_DEV_ACCESS_KEY_SECRET=$(kamal secrets extract R2_DEV_ACCESS_KEY_SECRET ${SECRETS})
 HONEYBADGER_API_KEY=$(kamal secrets extract HONEYBADGER_API_KEY ${SECRETS})
 HONEYBADGER_FRONTEND_API_KEY=$(kamal secrets extract HONEYBADGER_FRONTEND_API_KEY ${SECRETS})
+
+# base64-encoded; see config/deploy.review.yml
+SAML_SP_CERTIFICATE=$(kamal secrets extract SAML_SP_CERTIFICATE ${SECRETS})
+SAML_SP_PRIVATE_KEY=$(kamal secrets extract SAML_SP_PRIVATE_KEY ${SECRETS})
```
sethherr: Paired with the split fetch above — `|| true` so a missing field yields a blank secret rather than aborting the file. `env.secret` in `config/deploy.review.yml` only requires the key to be *present*, so blank passes through and `Saml::SettingsBuilder`'s `.presence` turns it back into a 404.

```suggestion
SAML_SP_CERTIFICATE=$(kamal secrets extract SAML_SP_CERTIFICATE ${SAML_SECRETS} || true)
SAML_SP_PRIVATE_KEY=$(kamal secrets extract SAML_SP_PRIVATE_KEY ${SAML_SECRETS} || true)
```

docs/review-apps.md
```diff
@@ -158,7 +158,16 @@ POSTGRES_PASSWORD            SECRET_KEY_BASE              SESSION_SECRET
 VERIFICATION_SECRET          GOOGLE_MAPS                  GOOGLE_MAPS_STATIC
 GOOGLE_GEOCODER              MAPBOX_GEOCODER              MAPBOX_MAPPING
 R2_DEV_ENDPOINT              R2_DEV_ACCESS_KEY            R2_DEV_ACCESS_KEY_SECRET
-HONEYBADGER_API_KEY
+HONEYBADGER_API_KEY          SAML_SP_CERTIFICATE          SAML_SP_PRIVATE_KEY
```
sethherr: Matching doc change, so nobody reads this table as "you must add these two before kamal works".

Two things folded in — drop the second if you'd rather keep the PR tight: `HONEYBADGER_FRONTEND_API_KEY` is in the fetch list on `main` but has never been in this table, so the table currently under-lists the required fields by one.

```suggestion
HONEYBADGER_API_KEY          HONEYBADGER_FRONTEND_API_KEY

# Optional — every kamal command works without these; only SSO needs them
SAML_SP_CERTIFICATE          SAML_SP_PRIVATE_KEY
```

eld120:
## Screenshots

### /admin/organizations/:slug/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4163, opened 2026-08-19
Persist the register step 1 draft, and re-space the org sidebar pages

sethherr:
Nothing typed into register step 1 survived a reload — `form-persist` was wired onto step 2 and the report step, but never onto step 1.

- **Step 1 keeps a draft, keyed per registration.** A restore assigns values without firing events, so the two controllers a step away from the vehicle-type combobox — the section label and the electric checkbox — reconcile on `form-persist:restored` as well as in their own `connect`. The embed stays out of it: a frame's localStorage is partitioned per embedding site.
- **`--nav-gap` carries the org sidebar's top inset.** Those pages have no navbar, so the variable was `0px` and `main` hard-coded the same spacing itself — which left `-mt-(--nav-gap)`, the idiom a full-bleed page pulls up under the navbar with, cancelling nothing. The inset is unchanged at both breakpoints; what's new is that a full-bleed shell can pull up under it, and stretches to fill the window when it's `main`'s only child.
- **The user alerts wrapper always wraps.** The organized content column passes the spacing it wants rather than passing nothing and having the wrapper drop out.

### DIFF
--- a/.claude/skills/frontend-conventions/SKILL.md
+++ b/.claude/skills/frontend-conventions/SKILL.md
@@ -73,6 +73,12 @@ The same instinct applies beyond buttons: **check `app/components/ui/` and `app/
 
 **Every typeahead / autocomplete / combobox goes through `UI::Forms::Combobox::Component`** — never a new Stimulus controller that fetches matches and renders its own menu. See `app/components/ui/forms/combobox/` (component + `component_preview.rb`) and `spec/components/ui/forms/combobox` for how to invoke it.
 
+## Form drafts: always `form-persist`
+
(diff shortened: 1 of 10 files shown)


sethherr:
## Screenshots

### /register?step=1

Manufacturer, vehicle type and email filled in, then the page reloaded.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Unchanged — `main` insets by the same 28px/16px it padded itself with before.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4164, opened 2026-08-19
Share the account menu between the navbar and the org sidebar

sethherr:
The navbar's gear and the org sidebar's account block carried the same rows — the organization switcher, the account links, logout — built twice and worded differently.

- **`UserServices::MenuItemsAccount` builds the whole list now,** taking which way the menu unrolls from its trigger as an argument — down from the navbar's gear, up from the sidebar's account block — so the rows read outward either way. Two components render it: `PageBlock::Navbar::UserSettingsMenu` as the gear's own `<ul>`, `PageBlock::Navbar::AccountMenu` as a `UI::Dropdown` whose trigger the sidebar passes in. There's a preview at `/lookbook/inspect/page_block/navbar/account_menu/organizations`, with a user in no organization, one, and three.
- **The switcher says where you are, not only where you can go.** The organization being viewed is a label rather than a link, and the row for leaving it behind holds the reader on the page they're on rather than dropping them at the homepage. A superuser gets both from an organization they're no member of, which is where the sidebar's new super admin row sits too.
- **The services around it are renamed** — `OrgServices::Displayer`, `UserServices::MenuItemsOrg`, `UserServices::MenuItemsAccount` — with their translation scopes and the org menu cache key following, so every cached org menu misses once after deploy. `retrieval_link_url` and `registration_field_label` move off `OrganizedHelper` onto the displayer; the first stops appending the redundant `?locale=` that the org admin "not graduated" link used to carry.

Logout reads "Log out" in the navbar now, matching the sidebar, and is tinted in both. The one change on top of #4168 is that `UI::ActiveLink.from_item` slices the keys `initialize` takes rather than hand-copying its defaults.


sethherr:
## Screenshots

### / — settings gear open

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Unchanged, which is the point — the navbar keeps its own `<ul>`, and on mobile the rows stay inline in the hamburgler menu. Only the logout row differs: "Log out", tinted.

### /o/brakebills/registrations — account block open

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Captured as a superuser who isn't a member of Brakebills, which is why the branch grows two rows main has none of: it can reach an organization without belonging to it, so the menu now names the one being viewed and offers the way back out.

PR #4165, opened 2026-08-20
Let a user order their organization roles, and leave one

sethherr:
Which organization you land in was whichever one sorted first alphabetically — with no way to change it, and no way to say "none of them".

- **`organization_roles.priority` gives a user's roles an order they set.** `OrganizationRole.ordered_for` is how every listing reads them now — the account switcher, the API membership payloads, choose-a-registration — and priority 0 is the organization you view Bike Index as, so `OrganizationRole.default_organization` replaces `User#default_organization`, which had no way to return nothing.
- **My Account → Organization Roles is where you arrange them:** drag to reorder, choose which organization the navbar opens with after logging in (or none of them), and leave an organization — except one you administer, or one your email domain grants.
- **Existing roles are numbered by `Backfills::OrganizationRolePriorityJob`**, so the deploy doesn't sit on a table-wide update. Until it runs they all tie at 0 and fall back to id, which is the order they were already listed in.


sethherr:
## Screenshots

### /my_account/edit/organization_roles

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4167, opened 2026-08-20
Follow the passive organization from Search and Register

sethherr:
A member of an organization got the all-bikes search from the Search link, and "Register a bike" started a blank registration on top of the one they had in progress. Both now follow the organization they're acting as.

- **The Search link points at the passive organization's registrations.** `default_bike_search_path` resolves off it, so a member searches their own registrations from anywhere on the site — and a registration's auto organization is that same one, which answers which of several theirs it belongs to rather than leaving it unassigned. `user_root_url` still sends law enforcement to the whole registry, which is the point of that branch.
- **The register flow's two entry points are split by URL.** `/register` goes back to the registration in progress, at the step it reached, and falls through to `/register/new` when there isn't one — which is where the footer and the organized add-a-bike link now point. `/register/new` always starts one, which is what "start over", "register another bike", the impound alert's "add a stolen bike" and an organization's printed QR code each want. Two waiting at most, sparing any with a confirmation email out.
- **Dropdowns and tooltips float over the org sidebar.** Both numbered z-indexes from 50 against a sidebar at 1050, so the view switcher's menu — which opens at the left edge of the content column and extends over the sidebar — rendered underneath it. They now share one source, counted by how many are open at once so it can't climb into Bootstrap's modal band.
- **The unfinished-registration alert names the manufacturer**, the view switcher's current view is `.twdropdown`'s "nowhere to go" label rather than a link back to the page it's on, and the register flow's reloads wait for form-persist's lazily loaded module before asserting the restored draft — a flake CI hit on `main`.


sethherr:
## Screenshots

### /registrations/69 (view switcher open)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4168, opened 2026-08-20
Mark the current dropdown entry on its own link

sethherr:
`UI::Dropdown` decided which of its entries was current on the server: the caller passed `active:` per entry, and the `<li>` carried both the `aria-current` and the fill. That's a second answer to a question `UI::ActiveLink` already answers in the browser, and `aria-current` isn't valid on `role="menuitem"` anyway.

- **The entry's own link carries the state.** `with_entry_item` takes no `active:`, the `<li>` is just `role="menuitem"`, and `.twdropdown` fills whichever anchor `ui--active-link` marks — the same purple as an active secondary `UI::Button`.
- **`match: :query` is what a filter entry needs.** It stands for the params it applies rather than for a URL: the entry in force links *away* from itself to clear the filter, and the page carries a page number and an org slug the link needn't. Only the params the entry names are compared, and a `nil` among a param's values means the entry a controller falls back to when the param is absent.
- **Every filter dropdown moves over** — admin bug reports, public images, registration sequences and the deleted-records picker, plus the organized impound, graduated-notification and registration lists.

Extracted from #4164 so the menu-manifest rename can land separately; `from_item` comes with it, with `OrgSidebar` as its caller here.

### DIFF
--- a/app/assets/tailwind/bike_index_components.css
+++ b/app/assets/tailwind/bike_index_components.css
@@ -104,9 +104,7 @@
       tw:hover:bg-gray-100 tw:hover:text-gray-900
       tw:dark:text-gray-200 tw:dark:hover:bg-gray-700 tw:dark:hover:text-gray-100;
 
-    /* ui--active-link marks the link itself, so the entry reads its own state rather than
-     * the li's. The fill is UI::Button's secondary active, minus the border and ring a
-     * menu item has no room for */
+    /* UI::Button's secondary active fill, minus the border and ring a menu item has no room for */
     @variant is-active {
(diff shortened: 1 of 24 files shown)


sethherr:
## Screenshots

Both pages render identically to `main` — the fill on the entry in force is unchanged, it's just read off the entry's own link now instead of the `<li>`.

### /admin/bug_reports

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4173, opened 2026-08-20
Make admin organizations a tabbed interface

sethherr:
Admin organizations had one show page and one edit form that had grown to cover locations, paid-feature settings and SAML config all at once. Each is now its own tab: Show, Edit, Locations *N*, Edit paid functionality, SSO, Invoices and Custom layouts.

- **`Admin::Headers::Tabs` is the header for an admin record with more than one screen** — the title, the top-right links, and `UI::Tabs` for the row itself. A section names its tabs in a component of its own (`Admin::Organizations::Tabs`) rather than in each view, so a screen can't disagree with its siblings about what the tabs are. A tab with nothing behind it is dropped, except on its own page — SSO appears once an invoice enables it. The row scrolls sideways rather than wrapping, and visits through Turbo Drive; the custom layout forms opt out, because their uploader comes from the legacy admin bundle and a Turbo restore gives back markup nothing is bound to.
- **The form tabs are one `#edit` action, and each tab only submits what it renders.** `?tab=` names the template; `update` returns to the tab it came from. Because every tab posts the whole organization, `kind` and the registration labels are only written when the tab holding them is the one submitting — otherwise a save from the SSO tab reset the organization's kind to "other" and dropped its custom labels.
- **Everything under `admin/organizations` is a ViewComponent** — tables are `UI::Table`, fields are `UI::Forms::*`, the features filter is a `UI::Forms::Combobox` rather than a selectize, and the 620-line show page is six components. Each takes its controller state as arguments the way #4128 asks, so nothing here reaches for `helpers.params`, `sortable_search_params` or `display_dev_info?`.
- **`.twwiderow` pairs cards and sections, the way `.twfieldrow` pairs fields** — both break on the row's own width rather than at a viewport breakpoint, so a row nested in a narrow card breaks where it should. `UI::Card`'s `full_bleed:` hangs off the same threshold through a container query, so a card meets the page's gutter exactly when its row drops to one column, and the two can't disagree.

Locations are off show and edit entirely — the locations tab's editor shows the same fields, editable, which also takes an N+1 off the show page.

The admin picker now names the section a nested controller is a slice of: an organization's invoices read "Viewing Invoices", with a way back to every invoice. That means `Admin::Navbar#on_page?` no longer calls `helpers.current_page_active?` — the one piece #4128 left deliberately unconverted, on the grounds that replacing it meant reimplementing the binxtils helper. Resolving a link's controller is needed for the nested lookup regardless, so sharing it removes the second source of truth rather than adding one.


sethherr:
## Screenshots

Brakebills, signed in as `dev@bikeindex.org` — the Custom layouts tab only exists for a developer. SSO isn't shown, since that tab only appears once an invoice enables the feature. On `main` the `tab=` param is ignored, so the locations and paid-functionality rows show the one combined form they were split out of.

### /admin/organizations/brakebills

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/brakebills/edit?tab=locations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/brakebills/edit?tab=paid_functionality

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/brakebills/invoices

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/brakebills/custom_layouts

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4177, opened 2026-08-21
Pass request state into components instead of reaching for helpers

sethherr:
Eleven components reached out of themselves to render — `helpers.display_dev_info?`, `helpers.hw_first_page?`, `params[:search_email]`, the controller's `controller_name`. This passes each of those in as an argument instead. No rendered markup changes.

- **`display_dev_info` is threaded from the call site** through the registration show tree and four org components, rather than each one asking the view. `Registrations::Show::Wrapper`'s `MARKUP_DIGEST` moves with it, so the registration show fragment cache busts on deploy.
- **The two combobox options components take `first_page`**, and `Admin::Navbar` takes the four request facts it was reading off the view. That drops the `with_request_url("…?page=2")` scaffolding the combobox spec needed to render at all.
- **The org search form's email and notes ride `interpreted_params`**, the hash it already takes every other repopulated field from — three of its five call sites were building a merge for exactly that.

This is the piece of a larger refactor - #4128 - that stands alone. The rest — `ComponentStates::SortState` for the sortable tables, then `IndexState` for the ~70 admin index views — depends on new value objects and follows separately.

Two components are only partly converted, deliberately. `Admin::Navbar#on_page?` still calls `helpers.current_page_active?`, which recognizes `request.url`; replacing it means reimplementing the binxtils helper in the component. And `Org::Search::Form` still renders `/shared/hidden_search_fields`, which reads ten `params` values of its own.


sethherr:
## Screenshots

Behaviour-preserving, so every pair below is byte-for-byte the same render — `magick compare -metric AE` reports **0 differing pixels** on all 8.

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/graduated_notifications?search_email=laura@brakebills.edu

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/14

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4178, opened 2026-08-21
Render the nested fields add trigger as a button, and pull the UI:: work forward

sethherr:
`UI::Forms::NestedFields` rendered its add trigger as `link_to "#"` — an anchor doing a Stimulus action's job. Nothing flagged it: `.herb.yml` disables `html-anchor-require-href` while the legacy views still rely on it. The `UI::` components #4173 builds on come along for the ride, so that PR can shed them.

- **The add trigger becomes a `UI::Button` with `color: :link`**, matching the remove trigger its callers already render. The component pins `tw:self-start` itself, since its button is `inline-flex` and would otherwise stretch to a flex or grid row and center its label. `add()` loses the `preventDefault` that suppressed the anchor's jump to top.
- **`UI::Container` and `UI::Tabs` arrive from #4173**, with previews and specs, alongside `mobile_flush` on `UI::Card` and the `.twgutter-bleed` / `.twfieldrow` classes `UI::Table` and `UI::Forms::AddressGroup` now use. Those files are byte-identical to #4173, so it merges either order — but review feedback on either has to be mirrored by hand until one lands.
- **`frontend-conventions` states the rule**: an in-page action trigger that doesn't navigate is a `UI::Button`, not `link_to "#"`.

Nothing here consumes `UI::Container`, `UI::Tabs` or `mobile_flush` yet — their callers stay on #4173, as does the `.twgutter` that `.twgutter-bleed` bleeds against.

`Org::SearchResults::BikesTable` and `Org::BikeAccessPanel` are the two anchors left under `app/components`; with those fixed the rule could be enabled with `app/views/**` excluded.


sethherr:
## Screenshots

Every page with a `main` version renders pixel-identically to it (`magick compare -metric AE` = 0 on all eight pairs) — the button swap and the CSS extraction are both invisible at rest. `UI::Tabs` and `UI::Card`'s `mobile_flush` are new, so they have no before.

### /rails/view_components/ui/forms/nested_fields/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/address_group/component/default

`.twfieldrow` — the auto-fit grid only diverges from the old fixed two columns below ~280px, narrower than this preview gets.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes

`UI::Table` on `.twgutter-bleed` — `--gutter` resolves to the 1rem the old `-mx-4`/`px-4` hardcoded.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/tabs/component/many_tabs

New. The row scrolls rather than wrapping; on mobile the Stimulus controller has scrolled the active tab into view.

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/ui/card/component/mobile_flush

New. The side borders and padding go below `md`; the top and bottom stay.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4180, opened 2026-08-21
Enable the anchor-href lint rule for app/components

sethherr:
#4178 left `html-anchor-require-href` disabled with two anchors under `app/components`, which is what kept it off. Both are dealt with, so the rule goes on there — the ~75 offenses in `app/views` stay out of scope, and `only:` means new components are covered as they're added.

- **The sticker modal is a `UI::Modal`.** Its `href="#"` trigger becomes a `UI::Button` with `commandfor`/`command`, and the `shared/modal` partial and its `capture` block give way to the component's `body` slot — so `/bikes/:id` drops one inline `<script>` and one delegated jQuery keyup handler. The two `.organized-form`-scoped classes inside it were already inert; the panel has no such ancestor.
- **`Org::SearchResults::BikesTable` is excluded by file.** `org--assign-bike-sticker` assigns those hrefs client-side because the sticker is a query param and the rows are cached without it — putting it in the cache key would leave every scan re-rendering cold. An inline `herb:disable` doesn't work here: the formatter moves it onto an unrelated `<a>` 168 lines up, and the linter then flags it as unnecessary while the real anchor goes unguarded.

`shared/_modal` keeps its 12 legacy HAML callers. The near-identical sticker modal in `Registrations::Show::WrapperOrgAdmin` is the redesign's copy of this one, left alone since this panel is what it replaces.

### DIFF
--- a/.claude/skills/pr/references/pre-push-cleanup.md
+++ b/.claude/skills/pr/references/pre-push-cleanup.md
@@ -34,6 +34,8 @@ Then review the changed files against `CLAUDE.md` (root and any nested ones in t
 
 **`bin/update_component_digests` goes after the last code edit, not before.** A `MARKUP_DIGEST` covers everything its cached tree renders out into, so editing a shared component (`UI::ActiveLink`, `UI::Button`) stales the digest of every component that renders it — `PageBlock::Navbar::Wrapper` and `PageBlock::Footer` both, for one edit — and regenerating before `/simplify`'s or the CLAUDE.md pass's own edits just means doing it twice.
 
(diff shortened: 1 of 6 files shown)


sethherr:
## Screenshots

The assign-sticker modal on `/bikes/:id`, opened from the panel's "Link Sticker" trigger. Signed in as `member@brakebills.edu`.

### /bikes/70

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The trigger behind it also changes: `less-strong` at `opacity: 0.8` becomes `UI::Button`'s `twlink`, so it reads as a normal link rather than muted grey.

PR #4181, opened 2026-08-21
Open every admin screen with Admin::Headers::Nav

sethherr:
Every admin screen opened with `.admin-subnav`, a bootstrap row whose -15px bleed each caller then compensated for — `Admin::Headers::Tabs` spent four `!` utilities undoing it. This replaces it with `Admin::Headers::Nav`, built on `UI::Header`, and takes the bootstrap markup out of the ~57 index views that fed it.

- **`Admin::Headers::Nav` is the row: heading left, the screen's links and filters right, an optional rule under it.** `Admin::Headers::Tabs` and `Admin::IndexSkeleton` both render it, so the tab row's gutter compensation goes with it. `Admin::CurrentHeader` joins them as `Admin::Headers::IndexInfo` (a rename, plus `render_admin_current_header` → `render_admin_current_info`).
- **Every `.nav-item`/`.nav-link`/`.dropdown-menu` block those views handed in becomes a `UI::Dropdown`, `UI::ButtonLink` or `UI::Button`**, so a filter row is one visual language instead of blue bootstrap pills beside purple `UI::` chips. Each entry is a `UI::ActiveLink` with `match: :query`, matching the `Admin::Organizations::Index::Filters` and deleted-filter dropdowns that already worked this way.

Nothing outside `app/components/admin` changes — the row takes `UI::Header` as it is, and the list carries the same `tw:mb-6` the heading does so the two end together.

`app/views/admin` now has no `.nav-link` outside the navbar, the bike tabs and the graphs subnav, none of which render in this row. The one JS-driven case is `#showStatusesSearch`: the vendored admin bundle finds it by id and toggles `.active` as it opens the collapse, so that class is the only thing that knows whether the panel is open — the button keys its active look off it rather than off `data-active`, which would be stuck at whatever the server rendered.


sethherr:
## Screenshots

### /admin/organizations?search_paid=true

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/users

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4183, opened 2026-08-21
Make the admin bike screens view components, and drop their legacy JS

sethherr:
Admin's bike screens were the last big consumer of the vendored `binx_admin` bundle: a 300-line `_bike_tabs` partial that twelve views rendered, a tab row the legacy `initBikeTabs` highlighted on load by matching a `bikes-edit`-style class against `data-activetab`, and a Bootstrap edit form. This does for them what #4173 did for organizations.

- **A section names its tabs in a component of its own.** `Admin::Bikes::Tabs` over the `Admin::Headers::Tabs` #4173 added, so which tab is active is server-rendered — which is what fixes `admin/ownerships/edit`, where the partial was passed `target:` rather than `active_tab:` and so highlighted nothing. `Admin::Bikes::Header` is that row plus `Summary::{Details,Theft,Recovery}` under it, and the twelve views render it directly. `show` validates `active_tab` against the tabs it renders instead of drawing a header over an empty body.
- **Everything under `admin/bikes` is a ViewComponent, and none of it is Bootstrap.** The edit form is `UI::Forms::Group` in `.twfieldrow` rows, the tables are `UI::Table`, and the index and multi-manufacturer filter rows are the `Admin::Headers::Nav` + `UI::ButtonLink` row #4181 established — no `.row`/`.col-*`, `.form-group`, `.btn` or `.admin-subnav` left in the section. Each component takes its controller state as arguments the way #4128 asks, which also retires the `.only-dev-visible` inline `<style>`: the dev-only rows just aren't rendered.
- **The bundle's bike JS is Stimulus.** `admin--bike-edit-form` for the mutually exclusive no-serial checkboxes and the recovery fields unchecking "Bike is stolen" reveals; `ui--collapse` for the statuses panel and the creation data. `initBikeTabs`, `initBikeStatusesSearch` and `fixStolenBikeImageWidth` all match nothing now — the image lists scroll on `overflow-x` rather than a summed width.

The summary keeps the look it had: `.admin-bike-summary` is the `#212529` full-bleed band `.admin-bike-tabs` was, and the registration, theft and recovery blocks are still the `.table-list` tables that sat on it — componentized, but not restyled.

The tab counts are cached on their own key rather than with the markup, which is #3979's idea — the active tab varies per screen and comes from an unvalidated param, so it can't go in a cache key, but the numbers can. That PR builds an `Admin::Bikes::Tabs` too, over the Bootstrap `.nav-tabs` that predate `UI::Tabs`; the two overlap on this component and whichever lands second has to reconcile.

Two fixes outside the conversion, both from behaviour it surfaced:

- **A recovery reason typed and then hidden still posted**, and any reason at all is what runs `add_recovery_information` — so re-checking "Bike is stolen" could recover the bike. `collapseField` disables what it hides. The legacy JS had the same gap.
- **`ui--collapse` only synced its chevron and `aria-expanded` when restoring from the URL param**, so a panel the server renders open — the statuses one, whenever the search isn't the default set — drew a chevron and announced a state that said closed.


sethherr:
## Screenshots

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes/26/edit — stolen registration

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes/80?active_tab=ownerships

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4184, opened 2026-08-21
Hide the paid functionality tab & unify dev marking

sethherr:
#4173 gave admin organizations a tab row where "Edit paid functionality" rendered for every organization, and "Custom layouts" — a page only a developer can open — looked like any other tab.

- **The paid functionality tab drops the way the SSO one does**, unless the organization is paid or you're on it — the page renders either way, saying there's nothing to configure, so it keeps its place in the row.
- **The dev marking becomes one tailwind-emitted `.only-dev-visible`**, replacing the copies in `admin_unvendored.scss` and `revised.scss`. A hand-written `@layer utilities` block rather than an `@utility`, so the name stays unprefixed and all ~95 existing sites — plus the `<style>.only-dev-visible {display: none}</style>` hooks several admin tables hide it with — keep matching. The border goes with it; it fought the shape it marked.
- **The organizations index search form drops bootstrap for `UI::Forms::Input` and `UI::Button`.** That's also what fixes its alignment: the combobox is a taller control than the fields beside it, so the row lined up on neither edge and the button sat above the name field. The bug reports row moves to the same classes, since it had the shorter-button version of the same problem.

`admin_unvendored.scss` is imported *outside* `admin.scss`'s `@layer legacy`, so its rules beat every tailwind utility whatever the specificity. That's why one `!important` survives there — `.table-list`'s striping would otherwise cover the marking — and why the show page's `hr` spacing is forced. `revised.scss` needs neither, and its copy of the marking was quietly winning over `--color-dev`, since `!important` reverses layer order.

`ComponentStates` becomes `ComponentStructs`, since both are `Data.define` structs. That regenerates six `MARKUP_DIGEST`s, so those fragment caches repopulate on deploy — five are low-traffic admin tables, but the sixth is `Registrations::Show::Wrapper`, which backs the public registration page.


sethherr:
## Screenshots

### /admin/organizations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/example-bike-shop

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/example-bike-shop/custom_layouts

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4187, opened 2026-08-21
Give SAML SSO its own admin card

eld120:
Parked draft, from local commits dated Aug 19 — pushed to preserve the work, not ready for review.

**Main has since grown its own SAML card.** The admin form refactor moved `admin/organization_form/` to `admin/organizations/form/` and added `form/saml_sso/`, so this branch's card is superseded and its paths no longer exist. Two pieces here aren't on main:

- **The slug is locked while SAML SSO is live** (`Organization#slug_stable_while_saml_live`). The SP entityID and ACS URL are built from the slug and the IdP has both registered from onboarding, so renaming silently breaks every login. This is the part worth keeping.
- **Developer-gated SAML fields.** Main's card takes no `current_user`, so anyone who can edit an organization can edit its IdP details.

Needs porting onto main's component layout rather than merging as-is.

### DIFF
--- a/app/components/admin/organization_form/feature_settings/component.html.erb
+++ b/app/components/admin/organization_form/feature_settings/component.html.erb
@@ -90,55 +90,5 @@
         </div>
       </div>
     <% end %>
-
-    <% if @organization.enabled?("saml_sso") %>
-      <div class="mt-4">
-        <h4>SAML SSO</h4>
-
-        <p class="text-success">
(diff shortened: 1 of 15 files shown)


sethherr:
@eld120 is this still relevant?

eld120:
Let's close this comment as much of the current scope of work from this PR is not needed.

PR #4188, opened 2026-08-21
Separate SAML configuration from activation, and add a test flow

eld120:
One boolean meant two things: that an organization's IdP details were filled in, and that every user at its domain was forced through SSO. So saving the SAML card immediately redirected the whole domain to an IdP that hasn't registered us yet — those users land on the IdP's own error page, and since no assertion comes back we get no callback and nothing in the logs. There was no way to finish our side of an integration and wait for theirs.

- **`active` is the redirect; `configured?` is only the IdP details.** A configuration can be saved, and an integration validated end to end, without routing anyone. `Organization.saml_email_matching` — what the forced-SSO redirect asks — requires both, which is what keeps saving the card from switching a domain over. `configured?` also requires the organization to claim a domain, since one that can't authorize an assertion can't complete a login either.
- **`/sso/<slug>/test` is the real login, landing on a report instead of your account page.** Enter the address you'll sign in with; it runs the transaction `init` runs, signs you in, and then reports the NameID, the attributes actually released, the resolved email and domain, whether the IdP released the address you entered, and whether the account already existed — enough to tell an attribute-release problem from a domain mismatch. A dry run would stop short of the half that breaks: provisioning, the force-confirm, the `SsoIdentity` write and the automatic role grant all happen after the point it could report success. Test mode travels in RelayState rather than through a second endpoint, because the assertion has to come back to the ACS URL the IdP registered.
- **The assertion domain guard is a per-organization check rather than the routing lookup.** It compares `Organization.email_domain` (now public) against the organization's own domain and rejects a blank on either side. Reusing the routing lookup would have made every callback conditional on `active`, and the previous direct comparison let `nil == nil` pass for an address with no parseable domain — a transient NameID with the email attribute unreleased is exactly that.
- **`/sso/<slug>/init` and the callback stay open while inactive**, which is what makes the validation window possible. An inactive configuration is one nobody is *forced* through, not one that's switched off — so the test page grants exactly what `init` already grants to the same IdP credentials, and reaching either state takes a superadmin entering IdP details.

The column rename is the only schema change, and no organization has SAML active yet, so there's nothing to migrate around it.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -76,6 +76,7 @@ Rails.application.routes.draw do
     get "/sso/:org_slug/sp.crt", to: "saml#certificate", as: :saml_certificate
     get "/sso/:org_slug/init", to: "saml#init", as: :saml_init
     get "/sso/:org_slug/test", to: "saml#test", as: :saml_test
+    post "/sso/:org_slug/test", to: "saml#test_start", as: :saml_test_start
     post "/sso/:org_slug/callback", to: "saml#callback", as: :saml_callback
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -76,7 +76,7 @@ Rails.application.routes.draw do
     get "/sso/:org_slug/sp.crt", to: "saml#certificate", as: :saml_certificate
     get "/sso/:org_slug/init", to: "saml#init", as: :saml_init
     get "/sso/:org_slug/test", to: "saml#test", as: :saml_test
-    post "/sso/:org_slug/test", to: "saml#test_start"
+    post "/sso/:org_slug/test_start", to: "saml#test_start", as: :saml_test_start
     post "/sso/:org_slug/callback", to: "saml#callback", as: :saml_callback
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -76,7 +76,7 @@ Rails.application.routes.draw do
     get "/sso/:org_slug/sp.crt", to: "saml#certificate", as: :saml_certificate
     get "/sso/:org_slug/init", to: "saml#init", as: :saml_init
     get "/sso/:org_slug/test", to: "saml#test", as: :saml_test
-    post "/sso/:org_slug/test", to: "saml#test_start", as: :saml_test_start
+    post "/sso/:org_slug/test", to: "saml#test_start"
     post "/sso/:org_slug/callback", to: "saml#callback", as: :saml_callback
   end
 
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -75,6 +75,7 @@ Rails.application.routes.draw do
     get "/sso/:org_slug/metadata", to: "saml#metadata", as: :saml_metadata
     get "/sso/:org_slug/sp.crt", to: "saml#certificate", as: :saml_certificate
     get "/sso/:org_slug/init", to: "saml#init", as: :saml_init
+    get "/sso/:org_slug/test", to: "saml#test", as: :saml_test
     post "/sso/:org_slug/callback", to: "saml#callback", as: :saml_callback
   end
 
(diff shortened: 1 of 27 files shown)


eld120:
## Screenshots

The SAML card. The notice shows while the configuration isn't active; the test page is listed with the other URLs you'd hand an IdP admin, since it works either way now. `/admin/organizations/brakebills/edit?tab=sso`

seya-saml-active-rename-admin-org-sso-20260823-002459-desktop
seya-saml-active-rename-admin-org-sso-20260823-002459-mobile

`/sso/<slug>/test` — the form asks which address you'll sign in with, so the result can say whether the IdP released it.

seya-saml-active-rename-sso-test-form-20260823-002459-desktop
seya-saml-active-rename-sso-test-form-20260823-002459-mobile

What comes back, against a local Keycloak IdP. A second run on the same address reports "already existed" and "You were signed in successfully!" instead.

seya-saml-active-rename-sso-test-result-20260823-002459-desktop
seya-saml-active-rename-sso-test-result-20260823-002459-mobile

spec/factories/organization_saml_configurations.rb
```diff
@@ -6,11 +6,15 @@
...
 
-    trait :enabled do
-      enabled { true }
+    trait :configured do
       idp_entity_id { "https://idp.example.edu/" }
       idp_sso_target_url { "https://idp.example.edu/idp/profile/SAML2/POST/SSO" }
       idp_cert { File.read(Rails.root.join("spec/fixtures/saml/idp_cert.pem")) }
     end
+
+    trait :active do
```
sethherr, line 15: Nice, love the composable traits

spec/models/organization_saml_configuration_spec.rb
```diff
@@ -43,18 +43,49 @@
...
+      it "matches configurations with all IdP essentials" do
+        expect(described_class.configured).to include(configured, inactive)
+        expect(described_class.configured).to_not include(without_domain)
+      end
+
+      it "matches configured inactive configurations" do
+        expect(described_class.configured_inactive).to eq [inactive]
+      end
+
+      it "does not count a configuration without an organization domain" do
```
sethherr: House rule: combine all examples into the same it block if they have the same setup - no reason to pay the time penalty for setting up the database multiple times

spec/requests/saml_callback_request_spec.rb
```diff
@@ -49,10 +68,59 @@ def signed_in?
...
+      end
+
+      it "re-renders the form without an email, having nothing to compare the assertion against" do
+        post "/sso/#{slug}/test", params: {email: " "}
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to include("Enter the email address to sign in with")
+        expect(Capybara.string(response.body))
+          .to have_css("form[action='/sso/#{slug}/test'] input[name=email]")
+      end
+    end
```
sethherr: To satisfy my [house rule](https://github.com/bikeindex/bike_index/pull/4188/changes#r3837024824) of no it blocks with the same setup, I'd wrap each of these in context blocks with a `let(:params)`

It's sort of silly, but it helps me glance through tests quickly and figure out what the differences are between tests.

config/routes.rb
```diff
@@ -75,6 +75,8 @@
     get "/sso/:org_slug/metadata", to: "saml#metadata", as: :saml_metadata
     get "/sso/:org_slug/sp.crt", to: "saml#certificate", as: :saml_certificate
     get "/sso/:org_slug/init", to: "saml#init", as: :saml_init
+    get "/sso/:org_slug/test", to: "saml#test", as: :saml_test
+    post "/sso/:org_slug/test", to: "saml#test_start"
```
sethherr: Hmmmm. I think you should switch this to be:

```suggestion
    post "/sso/:org_slug/test_start", to: "saml#test_start"
```

Unless you have a strong reason to not have the action match the route

PR #4189, opened 2026-08-22
Give the landing page embed its own registration origin

sethherr:
A registration started in an organization's landing page frame was indistinguishable from one started on `/register` — both were `register_flow`. This gives the frame its own origin, so the ownership a framed registration creates records where it began.

- **`register_flow_landing_page` is set by `register#embed`.** `b_param_for` only reuses a token whose origin matches, so the frame starts its own registration rather than continuing one `/register` began.
- **The register flow's origin list moved to `Ownership`**, which owns `ORIGIN_ENUM`, as `ORIGIN_REG_FLOW` — alongside a new `ORIGIN_WEBSITE_REG` that has no caller yet.
- **The admin origin graph's table sorts by count, with a swatch per row.** Each chart series carries its own color in the JSON now, so a row matches its columns without the table and the chart agreeing on an order. Both sides also drop from a query per origin to one grouped query — 17 to 1 on the chart's endpoint, where the empty buckets become zeros rather than nulls.


sethherr:
## Screenshots

### /admin/graphs?period=week&search_kind=bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4190, opened 2026-08-22
Match UI::ActiveLink on paths and params rather than a match: enum

sethherr:
`UI::ActiveLink`'s five `match:` modes conflated two independent questions — which pages a link covers, and which params count. `:path` and `:full_path` differed only on params, `:controller` and `:controller_action` only on breadth, and `:query` threw out the path question entirely. `match_paths:` and `match_params:` split them into the two axes they are, and `matching_controllers:` and `query:` go with the enum.

- **Paths are compared segment-wise in the browser** — `*` stands for one segment and a trailing `**` for the rest, so `/o/example/exports/**` covers a section and `/bikes/*/edit` can't span a slash. Matching on the path rather than the route drops `recognize_path` from the component and `data-page-route` off every layout's `<body>`, which took ~33µs per link on a page that renders 17 of them. A page a route match had covered but a pattern can't discover is named explicitly — `/admin/recover_organization`, routed outside `admin/organizations`, and `/o/:slug`, which `organized/dashboard#root` serves alongside `/o/:slug/dashboard`.
- **`match_params:` names the params that count and ignores the rest**, with `nil` for one the URL omits or leaves empty — the same absence `url_for` writes by dropping it. A filter entry now has to be on the page it filters as well: standing for its params alone, "Only deleted" went active on any page carrying `?search_deleted=only`.
- **The admin navbar's picker reads those patterns too**, rather than answering by controller. The prose naming the current page and the shortcut links beside it now describe a section's width once instead of twice, and `nested_nav_link` goes with the `recognize_path` memo — the nesting it existed for is the invoices entry's own `/admin/organizations/*/invoices`.

`aria-current` is derived rather than declared: a wildcard-free pattern equal to the current path, on a link that points at it and names no params, is `"page"`, and anything wider is `"true"` — so the nav's Search and Marketplace links move from `"true"` to `"page"`.

Worth a follow-up rather than growing this diff: the ~135 admin filter-dropdown entries across 40 index views each spell the same `(param, value)` pair twice, once into `url_for(...merge(...))` and once into `match_params:`. `Admin::Organizations::Index::Filters#filter_link` is already the extracted form.


sethherr:
## Screenshots

Nothing should look different — the same links go active, resolved from patterns rather than from a route the server published. The org sidebar row moves from `aria-current="true"` to `"page"`, which styles identically.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/users

The access filter open, where `match_params:` decides which entry is current — "Any Access" marks itself on a URL carrying none of the four params.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /help

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4191, opened 2026-08-22
Give the menu item, tab and entry shapes one definition

sethherr:
`UserServices::MenuItemsAccount` and `UserServices::MenuItemsOrg` each carried their own copy of the row constructors, and the two had already drifted — only the account one's `link` took `**attributes`, so no org row could carry `id:`/`data:`/`danger:` despite the comment saying the menus read alike. `UI::Tabs` and `UI::ButtonGroup` each documented their own hash shape separately again, and four components spelled those shapes out by hand.

- **`ComponentStructs::Shapes` defines all three shapes and builds them** — `link`, `group`, `disabled`, `divider`, `tab` and `entry` — and nothing spells one out by hand any more: the two menu services, `Admin::Bikes::Tabs`, `Admin::Organizations::Tabs`, `Admin::RegistrationSequence::Header` and both previews all go through it. Every constructor takes `**attributes`, which is how a caller carries a key of its own through — `Admin::Organizations::Tabs` tags each tab with the one it filters on.
- **The org sidebar's super admin row moved into `MenuItemsOrg`.** `SuperuserAbility` touches its user, so `cache_key_with_version` already covers granting and revoking one — the row belongs inside that cache rather than beside it, and `api/v3/me` serves it now too. It always renders a "SA" badge and a margin of its own, so nothing about it is special-cased in the sidebar.
- **`section: true` is every menu's now.** #4190 gave it to the org menu; the account menu's registration row takes it instead of interpolating its own `/**`.

The deeper version of this is `Data.define` rather than documented hashes, which is what the two `ComponentStructs` siblings already are. That reaches `api/v3/me`'s serialization, the per-`[organization, user]` cache and four templates doing `case item[:type]`, so it isn't this branch.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4192, opened 2026-08-22
Only scroll the UI::Tabs row when the tabs overflow it

sethherr:
`UI::Tabs` made its nav a scroll container on every screen, whether or not the tabs overflowed it — and a scroll container forces `overflow-y` to `auto`, so nothing in the row can sit outside it.

- **The controller drops `overflow-x-auto` when the tabs fit**, measuring off a `ResizeObserver` watching both the nav and the row. The class still ships in the markup, so without JS a row too wide to fit scrolls in place as it always did rather than widening the page. The nav's own padding is the gutter `twgutter-bleed` puts back rather than room the tabs can grow into, so the measurement is against its content box.
- **The tab states collapse onto the `is-active` variant**, replacing the `ACTIVE_CLASSES`/`INACTIVE_CLASSES` pair and the branch that chose between them. `tw:is-active:*` keys off the `aria-current` the tabs already set server-side, and sorts after the resting border and hover it overrides — which is what the old comment about `border-transparent` was working around.
- **The row's spacing moves from `pb-3` to `mb-3`, and `.admin-bike-summary` takes a matching `-mt-3`.** That space was the scrollbar's; with the scroll now conditional it belongs outside the box, where the band below can pull back against it. The tradeoff is that a classic scrollbar — Windows, Linux, or macOS set to always show them — sits against the row's underline rather than in a band below it.


sethherr:
## Screenshots

Admin registration edit — the tab row above `.admin-bike-summary`. Desktop fits all eight tabs, so the nav is not a scroll container; mobile overflows, so it is.

sethherr-tabs-conditional-overflow-x-admin-bike-show-20260823-121757-desktop
sethherr-tabs-conditional-overflow-x-admin-bike-show-20260823-121757-mobile

PR #4193, opened 2026-08-22
Take admin's collapses and form JS off the vendored bundle

sethherr:
Admin's collapsible panels were still Bootstrap's `data-toggle="collapse"`, and three of its forms got their behaviour from `vendored_assets/application_standalone.js` — a prebuilt jQuery bundle the admin layout loads outside the importmap, whose source left the repo along with its webpack config. This moves both onto `ui--collapse` and Stimulus, so nothing on these screens depends on the bundle any more.

- **Six panels convert to `ui--collapse`** — the Mailchimp raw data on the user show page, force merge email, a logged search's full log line, the Mailchimp data/values columns, and the news post date. The Mailchimp indexes wrap the whole `Admin::IndexSkeleton` render, because its nav header holds the trigger and its table holds the column with no ancestor in between.
- **Three forms become view components with a Stimulus controller each** — the ban fields on user edit, the blog/info groups on news edit, and the whole social post form. Where a hidden panel carries `.row` or `.card`, it needs `tw:hidden!` — `admin.css` ships their `display` outside `@layer legacy`, so a plain `tw:hidden` loses to it however specific it is.
- **The social post form had lost every one of its behaviours.** The bundle's module is guarded on `#tweetForm`, an id the page dropped when tweets became social posts, so the kind select, the repost select-all and the character counter have all been inert. All three work again.
- **`collapse()`'s duration becomes a named export.** Four callers each kept their own copy of `200`, two of them timing a `setTimeout` against the animation — where drift shows up as a panel measured mid-collapse. The comment on it records why the number isn't free to change.

The `frontend-conventions` skill now carries both traps behind this — grep `public/vendored_assets/*.js` before calling an admin hook dead, and the unlayered CSS that decides when `tw:hidden` needs the `!`. They were in `CLAUDE.md`, which loads for every task whether or not it's frontend.


sethherr:
## Screenshots

### /admin/mailchimp_data

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |


### /admin/users/7/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/news/1/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/social_posts/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4201, opened 2026-09-01
Give the chip groups UI::Button's secondary styling

sethherr:
`UI::ButtonGroup` took its colors from `UI::Button`'s secondary but restated the button's shape by hand, so the chips sat at a different radius, padding and weight from the button they were meant to match — and nothing failed when the two drifted.

- **Both chip groups derive from `UI::Button::Component.build_classes(color: :secondary, size: :md)`.** `UI::Forms::RadioButtonGroup` chips off `ButtonGroup::CHIP_CLASSES` as it did before, so one call still owns the answer, and its spec asserts equality against `build_classes` rather than inclusion — an extra utility on either side fails now.
- **The chip's own shape goes.** `rounded` → `rounded-lg` and `py-1` → `py-1.5`, plus `font-medium`; the hand-rolled `select-none` and `leading-snug` go, since the button carries neither. `text-sm` and `px-3` are the button's `:md` values already, so the type doesn't move — and colors, hover, focus ring, active fill and disabled were `UI::Button`'s constants before this branch.
- **A radio chip picks up press feedback.** `is-active` matches `:active` as well as `[data-active]`, so a `<label>` inheriting `ACTIVE_COLORS` alongside the resting half now flashes, while held, the same purple it lands on when checked.

Rendered by the org search settings, the register step-2 frame-size picker, the parking-notification forms, multi-search and the admin registration-sequence header.

### DIFF
--- a/app/components/org/search_results/bikes_table/component.rb
+++ b/app/components/org/search_results/bikes_table/component.rb
@@ -9,7 +9,7 @@ module Org
       # registrations on the show page). Pass render_sortable to enable sort links.
       class Component < ApplicationComponent
         # Digest of the markup inside the row cache — the cached_markup_digest spec keeps it current
-        MARKUP_DIGEST = "b1b88f27cf54"
+        MARKUP_DIGEST = "579ee5ab9811"
 
         delegate :additional_registration_fields, :column_renames, to: :settings_component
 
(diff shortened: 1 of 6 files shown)


sethherr:
## Screenshots

### /rails/view_components/ui/button_group/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/radio_button_group/component/full_width

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /search/marketplace

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4202, opened 2026-09-02
Add the organization's registration notes as an export column

sethherr:
Organizations with `registration_notes` can search their bikes' notes but had no way to export them — the note wasn't among the columns an export could carry.

- **`organization_notes` joins the feature-gated headers**, alongside `is_impounded` and `partial_registration`, so it shows up as a column choice on the new-export form and in the export's own column list. Both build those lists from the header set, so neither view needed changing.
- **A bike can carry notes from several organizations**; an export only ever writes the exporting organization's.
- **The notes load in one query per export**, keyed by bike id, rather than one per exported row.


sethherr:
## Screenshots

### /o/brakebills/exports/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4208, opened 2026-09-02
Style the registration show reveal forms consistently

sethherr:
The forms hidden behind a reveal button on the registration show page had each drifted into their own styling — three spacing mechanisms above the first field, and submits split between purple and blue for no reason a reader could recover.

- **Every revealed form's submit is now `:primary`** — "Send message" on both the public contact-owner card and the org panel, "Open claim", and the scanned sticker's "Update". Purple stays on the buttons that *open* a reveal and on the modal prompts, so the two roles stay tellable apart.
- **Top spacing moves onto the form** in all three, matching what scanned-sticker already did, rather than a wrapper override that restated `UI::Forms::Group`'s default bottom margin.
- **The parking-notification preview opens its panel on load**, so Lookbook shows the form rather than a bare trigger. It takes the panel as a preview param, which is what lets the system spec still drive the accordion from closed.

The impound claim's select label is reworded to "Select your stolen bike matching this impounded bike"; the key keeps its scope, so the next translation sync picks up the new English.


sethherr:
## Screenshots

### /rails/view_components/pages/registrations/show/wrapper/claim_impound/component/with_stolen_registration

The claim form, revealed by clicking "Claim found bike".

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/pages/registrations/show/org_top_actions/parking_notification_form/component/default

Captured on load — the "before" is the two triggers with the panel still closed.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4209, opened 2026-09-02
Stop a double tap from wedging the hamburgler menu

sethherr:
A tablet bug report: the hamburgler stopped responding to touch, and nothing below it could be tapped either. Both halves of the menu toggle finish in a `setTimeout` and each wrote its half blind, so a second tap inside the window left the stale one to apply what it had just undone — the backdrop raised over a closed menu, which locks body scroll as well, or the menu hidden behind a hamburgler showing open.

- **The deferred halves read the button rather than assuming the state they were scheduled in**, so a stale timer is a no-op instead of something to cancel. Reproduced on production at 810px, and the system spec double-taps for it.
- **The centred signup link hid a pixel before the desktop menu appeared** — at exactly 991px a signed-out visitor had no Sign up anywhere, one being hidden and the other still behind the hamburgler.
- **Marketplace rendered twice in the hamburgler menu.** Rows paired across the breakpoint go through `SIDE_CLASSES` now, so one can't lose its `d-none` the way this did.

Left alone: the hamburgler's tap target is 32×66 CSS px, well under the 44pt minimum and sitting ~60px in from the right edge of an iPad-portrait screen — which is plausibly what has people tapping it twice to begin with.

### DIFF
--- a/.claude/skills/integration-testing/SKILL.md
+++ b/.claude/skills/integration-testing/SKILL.md
@@ -106,6 +106,14 @@ drop the cap to 1ms and the assertions it protects should fail.
 
 `element.evaluate_script` may return a Promise; the driver awaits it before handing back.
 
+## A negative matcher can't wait for a state to settle
+
+`have_no_css`/`have_no_link` match the instant the condition holds, so a state that arrives
+*late* — a class an orphaned `setTimeout` re-adds — never gets asserted, and the example
+passes against the bug it was written for. Wait on a positive signal the sequence drops last,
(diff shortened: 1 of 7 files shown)


sethherr:
## Screenshots

### / (hamburgler menu open at mobile width)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

The mobile pair is the menu open — Marketplace is listed twice on `main`. Neither viewport lands on the 991px signup gap; that one is a one-pixel band between them.

PR #4211, opened 2026-09-03
Let a clicked-open tooltip's text be clicked and selected

sethherr:
A tooltip's text couldn't be clicked or selected: the popup only took pointer events when its body happened to contain a link or button, sniffed with a regex over the rendered HTML. Even with pointer events, clicking the text moved focus to `<body>` and the `focusout` dismissed the tooltip out from under you — as did switching to another program's window.

- **Pointer events follow the state, not the body's markup.** The controller turns `pointer-events-none` off while the tooltip is held open, so a clicked-open one can be clicked and its text selected, while a hover-only one stays transparent to whatever it overlaps. A tooltip with a link in it now wants the trigger clicked first, which is the flow the preview already documented.
- **`focusout` only dismisses when focus lands somewhere else in the page.** A null `relatedTarget` means the window lost focus to another program, which isn't the rider being done with the tooltip. Clicking outside and Escape still dismiss, and Tab-away still does the job `focusout` is there for.


sethherr:
## Screenshots

### /lookbook/preview/ui/tooltip/variants (tooltip clicked open)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4212, opened 2026-09-03
Move page components out of shared_blocks, and shared renderers into atoms

sethherr:
The last reorg left `shared_blocks` holding nine components that render a route, and `pages` holding four that every route reuses — so the folder a component sat in no longer told you what it was. Each moves to the folder its role calls for, which is all this is apart from the `OriginDisplay` copy below.

- **Nine components leave `shared_blocks` for `pages`** — the homepage, the landing pages, my_account, memberships and sales. Four leave `pages` for `atoms`: the admin badges, `AddressRecordCell`, `PaginationWithCount` and `OriginDisplay`. `Pages::Org::BikeAccessPanel` and the three impound-record components regroup under the route that renders them.
- **`Atoms::Org::OriginDisplay` gains the three `register_flow` origins**, which fell through to a generic "registered via register flow organized", and a preview covering every `creation_description`. Labels and tooltips are lowercase now, with only a name inside them title case — and `bulk reg` turned out to be a key nothing produces, so the spreadsheet-import tooltip had never shown.
- **`LandingPages::ORGANIZATIONS` becomes `LandingPageOrganizations::SLUGS`**, because `Pages::LandingPages` would otherwise shadow the old name for every component nested in `module Pages` — and a shadowed constant raises only where something happens to read it.
- **The four `translation.*.yml` are re-nested by hand**, so no non-English reader drops back to English waiting on a sync. `i18n-tasks` keeps a write rule per sub-namespace that has keys and loses the two that had none, and `brakeman.ignore` is re-fingerprinted for the one moved path it hashes.


sethherr:
## Screenshots

Two moved page components, plus the one component whose copy actually changed. The homepage and for_schools pairs are here to show parity — the hero's bike tiles are picked at random per request, so the two shots of each differ by about as many pixels as two captures of the same commit do (measured: 524,836 vs 544,398, same bounding box).

### /

| Desktop | Mobile |
| --- | --- |
| homepage desktop on main | homepage mobile on main |

main 👆

| Desktop | Mobile |
| --- | --- |
| homepage desktop | homepage mobile |

### /for_schools

| Desktop | Mobile |
| --- | --- |
| for_schools desktop on main | for_schools mobile on main |

main 👆

| Desktop | Mobile |
| --- | --- |
| for_schools desktop | for_schools mobile |

### Atoms::Org::OriginDisplay — every creation_description

New preview, so there's nothing on main to compare against. Every label is lowercase. Title case only appears inside a tooltip — for a proper name, or for the POS kinds that fall through to the generic description.

| Desktop | Mobile |
| --- | --- |
| origin display preview desktop | origin display preview mobile |

PR #4216, opened 2026-09-03
Key the origin display off the enums, and translate it

sethherr:
`Ownership#creation_description` humanized distinct registration methods down to a shared string — `embed_extended` and `organization_form` both became "org reg", and `embed_partial` became "landing page", which is also what `register_flow_landing_page` was labelled. `Atoms::Org::OriginDisplay` took that string and reverse-derived structure back out of it with two hardcoded English hashes, which is how a key for a description nothing produces sat dead in one of them. It's gone; `creation_kind` replaces it.

- **`Ownership#creation_kind` stays in the enums' vocabulary** — a `pos_kind`, `:bulk_import`, or an `origin` — and `Ownership.creation_kinds` enumerates them. `ownership_spec` holds that list to what `creation_kind` actually returns for every enum value, since it restates the branching rather than deriving from it.
- **The copy lives on the model**, following `MarketplaceListing.condition_humanized` — a label and a description per kind under `activerecord.enums.ownership`, with `creation_kind_humanized` / `creation_kind_description` guarding unknown values. Every kind has both, so nothing falls through to the old "registered via Broken lightspeed", including the two `broken_*` POS kinds. Those are defensive: across 2.06M production ownerships `pos_kind` is only ever `no_pos`, `lightspeed_pos` or `ascend_pos`, since `broken_*` is set on an organization rather than a registration.
- **`registration_method` becomes a selectable export column.** It had a `value_for_header` branch but was in no header list, so nothing could pick it; it's in `EXTRA_HEADERS` now and exports the label. Partial registrations export it blank, like the other bike-only columns.


sethherr:
## Screenshots

The preview is where the copy change lives — 22 rows become 23, and the nine that used to fall through to "registered via api v1" get a real description. Each row is labelled with the `creation_kind` it renders.

### Atoms::Org::OriginDisplay — every kind

| Desktop | Mobile |
| --- | --- |
| origin display on main | origin display mobile on main |

main 👆

| Desktop | Mobile |
| --- | --- |
| origin display | origin display mobile |

Here to show the ten converted call sites still render what they did — every row on this page is `embed_extended`, whose copy is unchanged, and the pair is identical to the pixel.

| Desktop | Mobile |
| --- | --- |
| org registrations on main | org registrations mobile on main |

main 👆

| Desktop | Mobile |
| --- | --- |
| org registrations | org registrations mobile |



### /admin/ownerships

The newly converted call site — this table printed the raw `creation_description` until now, so it's the one screen where the labels change. New on the branch, so there's no `main` shot to pair it with.

| Desktop |
| --- |
| admin ownerships table |

PR #4218, opened 2026-09-03
Create UI::JsonDisplay, Convert two admin tables to UI::Table

sethherr:
Two admin indexes were still hand-rolling Bootstrap `<table>` markup for straightforwardly sortable lists, and the JSON dumps in them were unreadable in dark mode — CodeRay bakes its token colours in as inline styles, so the box went dark while the keys stayed dark purple.

- **Partial-registrations and user-registration-organizations move to `UI::Table`**, and the b_params show page from haml to ERB on `UI::DefinitionList`. Organization cells become `Pages::Admin::Organizations::Cell`, so the name links to the org and a 🔍 adds the filter while keeping the current search params. The creator/user columns stay plain links — `Pages::Admin::Users::Cell` renders a badge costing ~9 queries a row, 214 for a 25-row page.
- **`UI::JsonDisplay` owns the admin JSON dumps**, highlighted in the browser by vendored highlight.js. As classes rather than inline styles the tokens take `dark:` variants, and the theme is six scopes in `json_display.css` rather than a vendored stylesheet. `pretty_print_json` keeps its other 13 call sites for now, so CodeRay stays a dependency.
- **Two bugs the conversion surfaced**: the Errors column iterated `bike_errors` with `each` and printed the raw array, and a soft-deleted user made a user-registration-organizations row raise on its bike count while the "Missing user" branch meant to catch it was gated on a condition that could never be false.

`#admin-content` gains `tw:mt-6`. Worth knowing it's a no-op on the ~70 admin pages that render `Pages::Admin::Headers::Nav` — its own `tw:mt-6` collapses with it — and adds 24px on the ones that don't, like the b_params show page.


sethherr:
## Screenshots

### /admin/b_params

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/b_params/:id

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/user_registration_organizations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /lookbook/preview/ui/json_display/variants

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4226, opened 2026-09-08
Finish the UI::JsonDisplay conversion, remove pretty_print_json and coderay

sethherr:
#4218 introduced `UI::JsonDisplay` but left `pretty_print_json` with its other 13 call sites, so CodeRay stayed a dependency and admin JSON dumps kept rendering with token colours baked in as inline styles — unreadable in dark mode.

- **Converts the remaining 13 call sites**, then deletes the helper, the `coderay` gem, and the `.data-cell` / `.data-cell.only-data` SCSS that styled the `pre` inside JSON table cells. Both copies of that CSS had to go: `admin_unvendored.scss` is what was actually in effect on these pages, `_new_table_extensions.scss` reaches them via `revised.scss`.
- **`table_cell:` gives way to `max_width: :table_cell`**, so a JSON cell is one argument rather than two that each half-set the width. The name carries the `.highlightjs-json-cell` chrome as well, and that chrome is now scoped to `.ui-table` — its `rounded-none border-0` leans on the row borders to redraw the box, and a Bootstrap `.table` only gives a `td` a `border-top`. So the four legacy tables get the cap with their border intact, and swapping them to `UI::Table` later needs no change here. `max_width:` takes an Integer or that one name, and raises otherwise.
- **Two fixes the conversion surfaced**: `/admin/external_registry_bikes/:id` 500'd on a null `date_stolen` (`l nil` raises, and `VerlorenOfGevondenBike.parse_date_found` can return one) — now `UI::Time::Component`, with the request spec that catches it. And logged searches' log-line toggle was absolutely positioned over what used to be a borderless dump, so it covered the new box; it moves into flow.

`/admin/ip_location` passes no width at all — its `col-md-6` already bounds the box, and a cap there only cost the Raw Headers dump 189px of every line.

### DIFF
--- a/.claude/skills/pr/SKILL.md
+++ b/.claude/skills/pr/SKILL.md
@@ -158,10 +158,10 @@ Most runs turn up nothing — say so and stop.
 
 Two bars before you edit a skill, because a wrong edit here is permanent and costs every future run:
 
-- **Did the skill already say it?** Diff the steps you ran against the steps as written. Out of order, or skipped, is your deviation — not a gap. A new rule stacked on the one you didn't follow leaves the skill longer and no better followed.
+- **Did the skill already say it?** Diff the steps you ran against the steps as written. Out of order or skipped is your deviation, so don't stack a new rule on the one you didn't follow. Revise the existing rule instead — reword it, or move it to where it gets read — since a rule you missed is a rule that was missable.
 - **Does it pay for itself?** Weigh how often the guidance fires against what it saves when it does, and state both. Guidance that runs every time to catch something rare is negative, however cheap each run looks.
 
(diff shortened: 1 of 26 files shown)


sethherr:
## Screenshots

### /admin/ip_location

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/external_registry_bikes/1

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/logged_searches

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/paper_trail_versions

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/ownerships/1/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4227, opened 2026-09-08
Keep a clicked-open tooltip reopenable after Escape

sethherr:
#4211 made a clicked-open tooltip's text clickable by holding it open while the trigger has focus. Escape closes it but leaves focus on that trigger, so the next click fires no `focusin` — and nothing else could reopen it, leaving the tooltip dead until focus landed somewhere else and came back.

- **Binds `click` alongside `focusin`**, so a click reopens a tooltip whose trigger is still focused. It's also the first time click-to-pin works in Safari and Firefox on macOS, which don't focus a button when you click it.
- **Escape returns focus to the trigger** when it was on a link inside the tooltip, rather than leaving it in content that just got hidden.
- **Dismisses on the press outside rather than the click**, so a selection dragged past the tooltip's edge no longer closes it — that click's target is a common ancestor, which read as clicking away. `ui--dropdown` still dismisses on the click and has the same bug.

The five `MARKUP_DIGEST` bumps are the changed `data-action` flushing the row caches that render tooltips.


sethherr:
## Screenshots

After clicking the tooltip trigger, pressing Escape, and clicking that same trigger again.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4228, opened 2026-09-08
Skip the donation modal for paid organizations' registrations

sethherr:
The donation modal on `/my_account` greets every signed-in rider, including the ones whose registration an organization already paid for — the modal a rider meets right after claiming a bike their shop or campus registered for them.

- **Skips the modal for a rider holding a registration from a `paid_money` organization**, gated outside the modal's `cache("donation_modal2")` block, whose key is global while the decision is per-user.
- **`dismiss_donation_modal` asserts the modal is up** rather than silently no-op'ing, so the four system specs that dismiss it now cover its presence — `register_organized_spec` covers the other side, registering with a paying organization and finding the modal gone.


sethherr:
## Screenshots

Signed in as a rider whose registrations came from a paying organization.

### /my_account

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4229, opened 2026-09-08
Add require_reg_phone, so an org can require the phone it asks for

sethherr:
`require_reg_address` and `require_reg_student_id` let an organization make the field it asks for mandatory; `reg_phone` had no equivalent. On register step 2 the phone is only required for a theft or a find, so an organization asking for one got a blank field on every ordinary registration.

- **Adds the `require_reg_phone` slug**, plus the seed feature that turns it on. It only requires what the organization also asks for — without `reg_phone` the field is hidden for every other status, so there'd be nothing to fill in.
- **Required-ness travels as the helper copy** `register--status-fields` reads back, so the organization's line merges *under* the theft/find copy — a stolen registration still says which of those is asking rather than naming the org.
- **Groups the three `require_*` slugs into `OrganizationFeature::REG_FIELDS_REQUIRED`.** Nothing reads it at runtime yet: the call sites still test the slug string directly, and a `require_reg_field?` companion to `BikeServices::Displayer.include_reg_field?` would be the follow-up.


sethherr:
## Screenshots

### /rails/view_components/pages/register/step2/component/with_organization

Brakebills with `reg_phone` + `require_reg_phone`.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4233, opened 2026-09-09
Make the registration redesign opt-out for everyone

sethherr:
The redesigned registration page sat behind a per-actor Flipper gate, so only the rollout group ever reached it and opting back out wrote a user column no signed-out visitor had. It's now the default for everyone.

- **The opt-out is the only switch left.** Signed in it's still `users.feature_registration_show_legacy`; signed out it's the session, so the toggle moved off `/my_account` — which requires auth — to `POST /registrations/:id/toggle_legacy_view`. Signing in carries a signed-out opt-out onto the account, since the redesign's own claim and sign-up CTAs are what push people through signing in.
- **`registration_redesign_disabled` replaces the actor gate as a kill switch**, closing both entry points rather than just `/bikes/:id`. It reads negatively — the only flag here that does — because a default-on rollout needs one whose absence means on, rather than one enabled in every environment.
- **`registrations#show` takes `force_html_response`.** Turbo leads its Accept with `turbo_stream`, and a component takes its content type from the request — so a Turbo submission redirected here answered as a turbo_stream, which Turbo silently discards. Harmless while the page was opt-in; as the default it left the organized sticker-assign flow sitting on the search results after a successful assign.

The `MARKUP_DIGEST` bump is the view-preference control changing shape, flushing the registration fragment caches that render it. The `CLAUDE.md` note is the third bullet written down — five controllers already used `force_html_response` and none of them said why.


sethherr:
## Screenshots

Signed out, which is the case this PR changes most — before, a visitor had no way to reach the classic page and no way back from it.

### /registrations/79

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/79?no_redesign=true

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4236, opened 2026-09-10
Space the general alert off the top with the nav gap

sethherr:
The general alert stands in the gap the navbar leaves under itself. The org sidebar replaces that navbar and is `position: fixed`, so on the pages an organization member sees, nothing is in flow above the alert and it rendered flush against the top of the window. The kelsey landing pages had already hit this — they zero the navbar's margin, and carried their own copy of the rule to make up for it.

- **The alert supplies its own `--nav-gap`**, which collapses into the navbar's matching `margin-bottom` where there is one. Ordinary and landing pages are unchanged; the sidebar's pages pick up the 28px their `main` already pads by. Its `body` becomes a `flow-root` so that margin can't collapse back out of the page — the alert is the first in-flow child there, and the canvas behind it is near-black in a browser that prefers dark.
- **Drops the landing pages' copy of that rule**, and the `general-alert` class it hung on, which nothing else referenced.
- **Focuses the file upload button purple**, matching every other `UI::Button` secondary — the label takes the rest of its look from `build_classes`, but restated primary's blue for the sr-only input that actually takes the focus.


sethherr:
## Screenshots

### /my_account (as a member of an organization)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/file_upload/component/default (focused)

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4237, opened 2026-09-10
Split the org registrations menu row when bike_search is off

sethherr:
Without the `bike_search` feature, `/o/<slug>/registrations` isn't a search — `Organized::RegistrationsController#index` branches on the feature and renders a plain paginated list of the organization's own registrations. Two places sent members there to look a bike up.

- **Splits the sidebar row in two for them** — "Organization Registrations" for their own list, plus "Search all registrations" into the public registry. Organizations with `bike_search` keep the single "Search Registrations" row, unchanged.
- **`default_bike_search_path` now requires the feature** rather than pointing every org member at their own index. Its one live caller is the `/serials` page's two search links; the primary navbar's Search never reaches the organization branch, since a member with a passive organization gets the org sidebar in the navbar's place.


sethherr:
## Screenshots

Captured as `Bike Recovery Team`, a seeded organization without `bike_search`. Mobile is the sidebar overlay, opened from the hamburgler.

### /o/bike-recovery-team/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4238, opened 2026-09-10
Convert admin's rails-ujs method: links to UI::ButtonLink

sethherr:
`link_to ... method: :delete` only submits because the legacy admin bundle still calls rails-ujs's `Rails.start()`. All 23 of those in admin are now `UI::ButtonLink::Component(method:)`, which renders a `button_to` form and needs no JS to submit.

- **`data: {confirm:}` becomes `form: {onsubmit: "return confirm(...)"}`**, matching `Pages::Admin::RegistrationSequence` and `admin/users/edit` — Turbo Drive is off app-wide, so `turbo_confirm` never fires. Two links had a bare top-level `confirm:` html option, which Rails renders as an inert attribute, so `admin/news` and `public_images/show` were deleting with no prompt at all and now ask.
- **A `button_to` form is block-level where the `<a>` was inline**, so two deletes had to move out of an enclosing `form_for` (`stolen_bikes/edit`, `social_posts/edit`), one paragraph became a `div`, and layout classes moved onto the wrapping form.
- **`_admin_public_image`'s delete is deliberately left alone.** The bundle binds `.image-delete-button` to its own `$.ajax` with `preventDefault()`, so it never used rails-ujs — and the same bundle emits a second copy of that row as a JS template, which would then render differently from the server's.

`Rails.start()` isn't droppable yet: `my_accounts`, `organized` and the shared navbar still have `method:` links, though those run on `application_revised`'s jquery_ujs rather than the admin bundle.


sethherr:
## Screenshots

### /admin/manufacturers/zycle_fix

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/example-bike-shop

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/stolen_bikes/43/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4239, opened 2026-09-10
Let trusted organizations message an impound without claiming it

sethherr:
An impounded or found registration could only be *claimed*, never asked about. Claiming wants a matching stolen registration and starts a formal process, so an organization that just needs to ask a question — is it still there, can we come get it, whose is it — had no route at all. The Message panel that would answer this already exists in the org view; it just never reached an impound, because `contactable?` tests `status_with_owner?` and an impounded vehicle is neither stolen nor with its owner.

- **Opens the existing `OrgTopActions::MessageOwner` panel on impounded and found registrations**, for staff (not `:limited`) of a trusted organization. No new form — widening that one predicate is the feature.
- **`ImpoundRecord#contactable_without_claiming?` is the allowance** — a role in any organization with `unstolen_notifications`, matching `paid_money`, or of `ambassador` kind. `Bike#contact_owner?` reads it, so opening the panel, the phone, the submit and the email send all answer to one predicate. It sits *after* the owner's `notification_unstolen` check, so this widens who may ask without overriding an owner who opted out — which also closes the gap where the panel could open on a bike whose owner declined and have the submit rejected.
- **`MessageOwner` asks what they need rather than where they saw it** when there's an impound record; whoever is holding it already knows where it is. Unchanged for stolen and with-owner.
- **Moves `ImpoundRecord`'s ten class methods into `class << self`**, per the convention in CLAUDE.md.

Note this rides on main's `is_paid` → `paid_money` rename: `Organization.paid_money` now means paid money in full rather than merely invoiced, so the allowance is narrower than when this branch started.


sethherr:
## Screenshots

### /registrations/38?view_as=brakebills.staff

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

A found registration in the org view, as Brakebills staff, captured with the Message panel open — collapsed it's just the action button, and on main there's no button at all. The Owner & access card still reads "Owner contact isn't available for bikes outside your organization", which stays true: this is an allowance to *ask*, routed through a notification, not access to the contact details.

PR #4243, opened 2026-09-10
Replace the Uppy admin image uploader with UI::Forms::Files::UploadMultiple

sethherr:
The admin image uploader on `admin/news/edit` and the two custom-layout pages ran on Uppy, out of the prebuilt legacy bundle whose source tree left the repo years ago. It rebuilt each uploaded image's markup in JavaScript — `_admin_public_image.html.haml` carried a `-# all this is also in image_uploader.js` comment pointing at the copy, and the two had drifted.

- **Adds `UI::Forms::Files::UploadMultiple`**, which posts each pick to `url` in a request of its own; the endpoint answers `{html:}` with the item's markup or `{error:}` with why it wouldn't store it. It and `UI::Forms::FileUpload` — now `UI::Forms::Files::Upload`, so every page that renders it is in this diff — both render a new `UI::Forms::Files::Picker`, the button, "No file chosen", camera on touch and drop frame. Their Stimulus controllers compose the same way: the picker announces a pick, and each upload controller wraps it and hears it — so `Files::Upload`'s progress now shows on a line under the button rather than after the filename.
- **`PublicImagesController#create` renders the item partial**, so an uploaded image is identical to the ones the page loaded with — that's what retires the JS copy.
- **Takes the image list off the legacy bundle too.** Delete is a `UI::Button` on an `admin--public-image` controller, and the primary-image radios join the blog form through their `form` attribute, making them one native group with "No primary image". That choice now actually saves — `Blog#set_index_image` threw away the 0 before its own `== 0` branch could read it.
- **Drops 3,100 lines of Uppy CSS** now nothing renders `.uppy`. The library itself still ships in `public/vendored_assets/*.js` — 12% of a bundle every admin page loads — but that bundle can't be rebuilt from this repo.

Server-rendering the item surfaced two things worth a look: a just-uploaded image has no `:small` version yet, so the partial picks up the `image-fallback` wiring `_public_image.html.haml` already had, and the list now preloads `file_attachment`.


sethherr:
## Screenshots

Same files uploaded on each side, so the lists match.

### /admin/news/:slug/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/organizations/:slug/custom_layouts/landing_page/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4248, opened 2026-09-10
Marketplace shipping: orders, partner shops and Stripe Connect

eld120:
Draft, for shape review rather than merge. The marketplace today is a listings board — `MarketplaceListing` holds the price and photos, `MarketplaceMessage` is the buyer↔seller chat, and `Sale` is a record the seller fills in afterward saying they sold it. No money moves and nothing ships. This is the spine for making it a real transaction: buyer pays, funds are held, the seller brings the bike **assembled** to a partner shop that boxes it and confirms it's the bike, and BikeFlights collects.

Nothing here is reachable by a user except one row on the listing panel. It's deliberately three branches' worth of foundation in one PR — none of the pieces stand alone, and splitting it into reviewable PRs is worth doing when it's ready to actually merge, not now.

- **`MarketplaceOrder` is the transaction record**, with one status enum covering both fulfillment kinds and the shipped-only states skipped for local pickup. Money is split across item, shipping, shop fee and platform fee so a partial refund is expressible, and the total freezes once the buyer has been charged.
- **A partner shop is an `Organization` with `kind: bike_shop`** plus the new `marketplace_partner` feature slug and one designated `Location`. The new record holds only what organizations don't have — the fee, the capacity, and whether we buy the BikeFlights label or the shop does. Two independent switches: the shop's own status, and the org feature.
- **`StripeAccount` is Connect, not the Checkout Sessions `Payment` uses** — Stripe holds the funds and carries the money-transmission licensing. Payout eligibility is Stripe's answer and arrives by `account.updated`, so someone can finish onboarding and still not be payable.
- **The BikeFlights wire format is unverified.** We have no working account yet, so every field name comes from their published docs and the specs are stubbed rather than recorded. It's confined to `Integrations::BikeFlights::Client` so confirming it against the sandbox is one file.
- **Shipping eligibility excludes e-bikes because of the battery, not the size** — lithium is a regulated dangerous good. The estimator picks between two of BikeFlights' published box sizes; they decline to map bike type to a box at all, so the shop re-measures at drop-off and we re-quote before buying a label.

Worth knowing: frame size is blank on 78 of 80 bikes locally, so the estimator's large-box branch may never fire in practice.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


eld120:
## Screenshots

### /bikes/45

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/pages/bikes/marketplace_listing_panel/component/not_shippable

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4252, opened 2026-09-12
Add a confirm: argument to UI::ButtonLink

sethherr:
Every confirm-guarded `UI::ButtonLink` hand-wrote the same `form: {onsubmit: "return confirm('#{j "…"}')"}`, escaping included, and the confirms that weren't ButtonLinks yet leaned on rails-ujs's `data-confirm`.

- **`confirm: "Are you sure?"`** puts that guard on whatever the browser acts on — the wrapping form for a `button_to`, the anchor itself otherwise. Where a nested `<form>` won't fit, the call site drops `method:` and lets Turbo carry the verb; the `onclick` gates it either way.
- **Two deletes weren't deleting.** The admin bike's `button_to` sat inside the edit page's `form_for`, so the parser dropped its form and the button submitted the *edit* form; `bikes_edit/remove` passed a bare `confirm:` to `link_to`, which renders as an inert attribute, so deleting a bike version prompted nothing. Both covered by browser specs now.
- **32 call sites, and nothing is left on `data-confirm`** — the 21 `form: {onsubmit:}` ones plus 11 `data: {confirm:}` links, which keep their legacy classes via `html_class:` (unlayered CSS still outranks the `tw:` utilities). `public_images`' `image-delete-button` stays out: jQuery binds it with its own `preventDefault`, so its `data-confirm` is already dead — that page has no prompt today and still won't.


sethherr:
## Screenshots

`/admin/bikes/67/edit` — the delete moves out of the edit form, onto its own row below Update. On main its `button_to` form was nested inside `form_for [:admin, @bike]`, so the parser dropped it and "Delete bike" became a submit for the edit form: no confirm, and a PATCH instead of a DELETE.

| Desktop | Mobile |
| --- | --- |
| admin bike edit on main | admin bike edit on main, mobile |
| main 👆 | this branch 👇 |
| admin bike edit on this branch | admin bike edit on this branch, mobile |


### /o/brakebills/users/1/edit

`btn btn-danger` → `color: :error`, and the enclosing `%p` becomes a div — a `button_to`'s form is flow content and would close the paragraph. On main this was an `<a data-confirm>` on rails-ujs.

| Desktop | Mobile |
| --- | --- |
| org user edit on main | org user edit on main, mobile |
| main 👆 | this branch 👇 |
| org user edit on this branch | org user edit on this branch, mobile |

### Strava disconnect

The `synced` preview of `Pages::MyAccount::StravaIntegration` — no seeded integration exists, so the preview is where this renders. `btn btn-sm btn-outline-danger` → `color: :error, size: :sm`.

| Desktop | Mobile |
| --- | --- |
| strava disconnect on main | strava disconnect on main, mobile |
| main 👆 | this branch 👇 |
| strava disconnect on this branch | strava disconnect on this branch, mobile |

The other converted pages need records the seeded DB doesn't have (a lock, an incomplete registration, a deleted organization), or keep their legacy classes via `html_class:` and render identically.

PR #4253, opened 2026-09-14
Render a Turbo link when a ButtonLink has a confirm

sethherr:
#4252 left confirms split across two mechanisms: `button_to`'s `onsubmit` where a form could nest, and a Turbo link's `onclick` where it couldn't. The component picks now.

- **`confirm:` alongside `method:` renders a Turbo link** rather than a `button_to` — an `<a>` carrying the method and an `onclick` that Turbo's click handler honours. Only the call sites that spelled the mechanism out themselves, or styled a wrapper that no longer exists, needed touching.
- **The spacebar activates one now.** An `<a>` only answers Enter, so every ButtonLink that isn't a real `<button>` ignored the key the thing it looks like takes.
- **Deleting a public image no longer lands on a 404** — it redirected back onto the page it had just deleted.


sethherr:
## Screenshots

Both pages lose a wrapping `<form>`, so what moves is layout rather than the buttons themselves.

### /admin/bikes/:id/edit

#4252 had to put the delete on its own row below the form — a `button_to`'s form couldn't nest inside `form_for`. An `<a>` can, so it goes back beside Update.

| Desktop | Mobile |
| --- | --- |
| admin bike edit on main | admin bike edit on main, mobile |
| main 👆 | this branch 👇 |
| admin bike edit on this branch | admin bike edit on this branch, mobile |

### /o/:slug/users/:id/edit

Same story: the `%p` that #4252 turned into a div comes back, so the button sits on a paragraph's baseline again.

| Desktop | Mobile |
| --- | --- |
| org user edit on main | org user edit on main, mobile |
| main 👆 | this branch 👇 |
| org user edit on this branch | org user edit on this branch, mobile |

PR #4255, opened 2026-09-14
Fix the terms pages' agree bar rendering behind the org sidebar

sethherr:
An organization member signing in for the first time lands on `/accept_vendor_terms`, whose agree bar is `position: fixed; left: 0; width: 100%` from before the org sidebar existed — so its checkbox rendered at x=1, behind the sidebar's z-index 1050, and couldn't be clicked. The opaque bar at z-index 100 also covered the flash toast saying the box wasn't checked, leaving a 28px sliver of it showing above the bar.

- **The bar is sticky in the content column now**, so it inherits `body.with-org-sidebar`'s margin instead of needing to know about it, and sits under the toast's z-50 rather than over it. `Pages::Users::AcceptTerms::Component` renders both terms pages and the legacy `.accept-vendor-terms` SCSS is gone; the submit is a `UI::Button` again, since `btn btn-primary btn-lg` has resolved to unstyled text for a while.
- **The checkbox is `required`**, so an unchecked submit gets the browser's prompt at the box rather than a toast the bar sits over. `UI::Forms::Checkbox` takes the keyword the way the rest of `UI::Forms` already does.

- **A browser spec walks the whole invite flow** — an org admin invites a rider, who takes the emailed link, signs up, is confirmed inline, and accepts the organization terms. On the pre-fix markup it fails at the checkbox with Playwright naming the interceptor: `org-sidebar … subtree intercepts pointer events`.


sethherr:
## Screenshots

### /accept_vendor_terms

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /accept_terms

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4267, opened 2026-09-16
Ask yahoo and hotmail addresses for a Turnstile challenge

sethherr:
Every Postmark spam complaint on file is a confirmation email to a yahoo or hotmail address — #4262 closed the bot half of that, this asks the rest. `/register` step 1 and `/users` now put a Cloudflare Turnstile challenge in front of those two domains, and nobody else.

- **The reveal is client-side, the check isn't.** The widget appears as soon as a risky address is typed, so nobody is bounced to discover it; `Sessionable#turnstile_verified?` re-checks the token on submit, and a form posted with no JavaScript — the embedded one included — lands there the same way and gets the widget on the step it's handed back.
- **Inert until `CLOUDFLARE_TURNSTILE_SITE_KEY` and `CLOUDFLARE_TURNSTILE_SECRET_KEY` are set** — unconfigured asks nobody, so a missing key can't lock anyone out of registering. `EmailDomain::RISKY_EMAIL_DOMAINS` is the list, which `Ownership#spam_risky_email?` already needed.
- **`api.js` is fetched on the first reveal rather than shipped with the form**, so visitors to the two busiest forms don't all pay a cross-origin iframe for a gate that rarely applies. It renders only the `.cf-turnstile` elements present when it loads, which is why a step Turbo hands back carries its own tag and the controller renders that container explicitly.


sethherr:
## Screenshots

Both forms with `rider@yahoo.com` typed in, which is what reveals the widget. The red "For testing only" band is Cloudflare's — the dev server runs their test sitekey, since a real one is domain-locked and won't render on localhost.

### /users/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /register?step=1

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4268, opened 2026-09-16
Rebuild the org registrations search to Kelsey's design

sethherr:
Builds the org registrations search to the design added in #4261 (`kelsey_redesign/new-org-search/Admin Dashboard Redesign.dc.html`). The page was a bare form over a table, with the filters buried in the same panel as the column toggles and the chart behind an opt-in link.

- **The search is a card now** — fields, then a row carrying the gear, the date range and what the search is currently filtered to, then a full-width submit. Everything the gear opens is one collapsible settings-and-filters panel, where the date range is #4333's radios like every other filter; the column toggles move out of it, into the results card, where they're a band across it with Export CSV up in its header. The card is the form's only variant, so every org search wears it and heading and submit text are required.
- **"Render chart" becomes a chart card beside it** — a lazily-loaded stacked chart split into registrations / e-vehicle / stolen, three headline counts, and each one's move against the window before it. It opens on `Last year`, which counts the organization's whole months rather than the search and holds them for an hour, so only the `Current search` scope refetches while you search. Impound records and graduated notifications get the same card on their own search alone — no scope toggle or stats — which a search's stream swaps for a fresh placeholder. Parking notifications has a map rather than a chart, so its form card spans the row.
- **Two new filters ride along.** `search_all` widens past the organization's own registrations. It refuses an export, stops counting and paging at 1,000, and switches off while an owner email or name is searched — folding that field and the chart card away while it's on — so it can't look people up across Bike Index. `search_unregisteredness` filters the records a parking notification created for an unregistered vehicle.
- **The two view switchers are #4277's `kind: :toggle`**, and all three cards are `full_bleed` inside one `.twwiderow`, so they meet the page gutter on a phone.

The column panel opens from the same "Column settings" button everywhere it appears — the results card, a bike's other registrations, multi search — and never carries an export of its own; only the search's results card offers Export CSV, since elsewhere the search params don't describe the bikes listed. The spreadsheet itself is untouched. The results card's view-as toggle is behind `organization_registration_view_switcher`, and `Thumbnail` renders the same spreadsheet — that view lives in the other design doc. The column panel's "Save as default" doesn't appear — columns already persist per-browser, so it would need new per-user storage to mean anything. `UI::Chart` now gives each chart its own element id — Chartkick numbered them per view, so two lazy charts on a page both drew into the first — and drops the legend from a single series.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/registrations — search settings open

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/model_audits/3

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

sethherr:
**Claude Code review** — 3 findings on commit 4d6012f

_The review reached its time or budget limit and stopped early, so it may not have covered the whole change._

### 🔴 Normal — `app/controllers/organized/registrations_controller.rb:157-158`

search_all is only hidden from the chart card by CSS, not blocked server-side, so the 'Current search' chart scope still fetches and returns cross-organization aggregate counts

### 🟡 Nit — `app/services/org_services/registration_counts.rb:30-32`

previous_range end and time_range start are both inclusive, so a bike created exactly at the boundary instant is counted in both the current and previous window

### 🟡 Nit — `app/components/pages/org/search/chart_card/component.html.erb:29-35`

Dead `org--search` Stimulus targets rendered on pages that never declare the `org--search` controller (sweep:`data-org--search-target="chart`)



---
_Generated by [Claude Code](https://claude.ai/code)_

PR #4269, opened 2026-09-16
Drop the border radius on a full-bleed card

sethherr:
`UI::Card::Component(full_bleed: true)` drops its sides and its horizontal padding once the `.twwiderow` holding it collapses to one column, so the card meets the page's gutter — but it kept its rounded corners, which by then have nothing to round against.

- **The radius drops inside the container query, not off the `full_bleed:` argument.** A full-bleed card in a row that's still two columns wide is an ordinary inset card, and keeps its corners.
- **All four declarations are forced now.** Tailwind orders a custom `@utility` by the properties it sets, so adding `border-radius` moved the whole rule behind `.tw:border` and silently put the card's sides and top back. `CLAUDE.md` gets the general form.


sethherr:
## Screenshots

### /rails/view_components/ui/card/component/full_bleed

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4270, opened 2026-09-16
Move org search settings data into ComponentStructs

sethherr:
`Pages::Org::Search::Settings::Component` was a data object as much as a panel: five callers built one to read column labels, the default column set or the active filter descriptions off it, and the copy `Pages::Org::SearchResults::BikesTable` built was never rendered at all. That data lives in `ComponentStructs::OrgSearchSettings` now, next to `IndexState` and `SortState`.

- **The component is down to its template, an `initialize`, and the Stimulus wiring for the element that wraps it** — all four views spelled that wiring out by hand before. Every caller takes `settings:` rather than a built component.
- **The column labels were written out three times** — once for the toggle checkboxes, once for the table headers, and a dead third set left behind in the wrapper scope when the table moved out of it. One copy now, in the settings panel's sidecar, read through the struct. The two live copies had drifted on three labels, so **three table headers change**: Avery reads "Avery Exportable", propulsion "E-vehicle (propulsion)", and registration notes picks up the org's short name. Both feature-gated columns went unrendered by every spec, which is how the drift went unnoticed. The prefix doubles as the sort key, so the toggle grid reorders with them.
- **An org collecting a secondary serial had its value rendered for every bike in the table, including other organizations'** — the standalone Secondary# column carried neither the `organized?` guard its neighbours use nor a feature gate, and was only invisible because the toggle controller hid it client-side, so the value sat in the HTML and in the fragment cache. Its `next` guard tested the un-prefixed field name and never fired, so the column was a duplicate too. The generic loop handles the field now, with its redaction.
- **Both org-search sidecars sat a directory above their components**, because the generic i18n-tasks write rule routes `components.pages.{:}.{:}` two levels deep and these are three. `MARKUP_DIGEST` globs a component's own directory, so the bikes table's cached cells rendered `hidden_not_registered` and `email_hidden` with nothing tracking their copy. A rule per scope moves every key next to the component that owns it, and a spec now fails on any sidecar that isn't colocated — which the digest spec can't catch, since it recomputes through the same glob.

The non-English column labels still sit under the old `shared.org_search_columns` scope and move with the next translation sync; until it runs, those readers fall back to English for the 24 labels.


sethherr:
## Screenshots

`brakebills`, settings panel forced open via `?search_open=true`, with **Propulsion** and **Registration Notes** toggled on — both are feature-gated and off by default, which is why the drift reached main unnoticed. The checkbox grid is pixel-identical to main; the diff below it is the three headers taking the panel's wording, which wrap to two lines and reflow the column widths.

### /o/brakebills/registrations?search_open=true

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

Three matching serials and one that matches nothing. Same two headers, same reflow; the rest of the page is unchanged by the rewrite.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4272, opened 2026-09-16
Let the impound messaging allowance reach abandoned bikes

sethherr:
#4239 let trusted organizations message whoever has an impounded bike without opening a claim against it. An abandoned bike is away from its owner the same way, but has no impound record — and the allowance hung off `ImpoundRecord`, so nothing reached it.

- **The allowance reads the bike's status, and only covers a registration that isn't the viewing org's own** — theirs carries the owner's contact in the table beside the panel, and on their own impound the update action is what they act through. Holding a bike and having it registered with you are separate: an impound raised from a parking notification writes no `bike_organization`.
- **The action asks the organization being viewed, not the viewer.** A superuser is a member of none, so previewing an org otherwise rendered nothing. The trust half is `User#contact_impounded?`, beside `enabled?` and `ambassador?` and taking the same superuser override. The org-side scope kept its impound-shaped name.
- **The message panel shows the owner phone again.** `phoneable_by?` already answered true through the allowance, and the info table beside the panel renders it — withholding it in one of the two places only made the page disagree with itself.


sethherr:
## Screenshots

### /registrations/63?view_as=bike-recovery-team.staff

An abandoned registration belonging to another organization, viewed by Bike Recovery Team staff, with the panel opened.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4277, opened 2026-09-16
Add a toggle kind to UI::ButtonGroup, and PeriodSelect.period_label

sethherr:
The design-system half of #4268, pulled out ahead of it so the org search PR is only the org search. Neither has a caller until that one lands.

- **`UI::ButtonGroup::Component(kind: :toggle)` renders the entries as a segmented control** — one track, with the active entry raised out of it — rather than a row of separate chips. Same entries either way, so it keeps the group's disabled-becomes-a-button and `aria-current`/`aria-pressed` handling; `full_width:` raises under it, since the track sizes itself to its segments. Sized to `kelsey_redesign/new-org-search/`.
- **`tw:text-2xs` isn't a token this repo defines**, so it emits nothing and the segments were rendering at the inherited 16px. Four other call sites still carry it — `UI::Tooltip`'s `?` glyph overflows its 16px pill today — so defining `--text-2xs` in the theme is worth a follow-up rather than four more arbitrary values.
- **`UI::PeriodSelect::Component.period_label` names the selected period** for a caller stating the range somewhere other than the period buttons. Worth a look: `GraphingHelper#humanized_time_range` answers a related question in different words ("in the past month" vs "past 30 days") and isn't translated — #4268 renders both on the same screen, so one of them should probably go.


sethherr:
## Screenshots

### /rails/view_components/ui/button_group/component/toggle

| Desktop | Mobile |
| --- | --- |
|  |  |

The preview is new on this branch, so there is no `main` version to compare against. Sizes match `kelsey_redesign/new-org-search/Admin Dashboard Redesign.dc.html`: 3px track padding and gap, 10px track radius, 5px/12px segment padding, 8px segment radius, 11.5px/800 text.

PR #4279, opened 2026-09-16
Wrap long organization sidebar rows instead of clipping them

sethherr:
Most rows in the organization sidebar interpolate the org's name, so a long one — the seeded "Bike Recovery Team", say — made "Bike Recovery Team Registrations" run past the 266px column, where the scroller's `overflow-x-hidden` cut it off mid-word. They wrap now, tightly enough that a two-line row still reads as one row.

- **`leading-tight` lives on `ROW` and `CHILD`**, the two constants that already own row typography, rather than on the label spans that were carrying the `nowrap`.
- **The account email at the bottom keeps its ellipsis.** It's a single unbreakable token, so wrapping it would take `break-all` — and it sits in the pinned footer, which the sidebar's reveal logic measures the scroller against.


sethherr:
## Screenshots

### /o/bike-recovery-team/registrations

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4282, opened 2026-09-16
Render every bike status through one Atoms::RegistrationStatusBadge

sethherr:
Bike registration status was rendered three ways — `bike_status_span`'s uppercase colored text, a Tailwind badge on registrations#show, and `status_display` in the org bike table — which disagreed about what "with owner" looks like: nothing, "Not stolen", or "registered, not stolen". `Atoms::RegistrationStatusBadge` is the one renderer now, everywhere a bike's status shows.

- **"with owner" reads as "registered", and every status carries a tooltip** explaining what it means — "Registered & protected. With its owner (not stolen)". The label comes from `Bike`'s humanized status, so the rename is one translation key rather than copy the badge owns; the untranslated `Bike.status_humanized` the API serializes is unaffected, and the four `translation.*.yml` keep the old wording until the next sync.
- **registrations#show loses its separate "Registered & protected" shield badge**, which the badge's tooltip now says. That one was hidden for user-hidden bikes and the status badge isn't, so a hidden bike shows it too.
- **`Component.status_humanized` is the one place deciding which status shows**, so an organization's export column can't drift from the badge — it gains `for sale`, which the deleted `Bike#status_humanized_no_with_owner` blanked along with plain with-owner bikes. Telling impounded from found reads an impound record, so the admin bike tables preload it.
- **`wait_for_stimulus` takes an identifier.** Its `every` is vacuously true on a document that has parsed no `[data-controller]` yet, so a click could land before the controller it needs connects — which the badge's tooltips surfaced on `/my_account`.


sethherr:
## Screenshots

### /registrations/62

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/atoms/registration_status_badge/component/all_statuses

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4283, opened 2026-09-16
Name the cycle type throughout the emails

sethherr:
A registration is as often an e-scooter or a wheelchair as a bike, but the emails said "bike" throughout — "Your bike is now part…", "Bike details", "View your bike" — whatever was registered. They name the cycle type now.

- **The finished-registration email interpolates the type into fourteen strings**, and `personal-mobility` is renamed to "e-Personal Mobility Device". Headings that open on the type use `type_titleize`, mid-sentence copy the lowercase `type`. Three strings say "this %{bike_type}" rather than "a %{bike_type}": the article can't come from the interpolation, since every locale keeps its own inside the sentence (`Has añadido un %{bike_type}`) and the value is translated too. The longer name grows the autocomplete prefix index by 80 keys — that's the `loader_spec` bump.
- **Auditing the other mailers turned up two wrong accessors, not just hardcoded copy.** `recovered_from_link` built its subject from `cycle_type_name`, so a recovered EPAMD got "Your e-personal mobility device (epamd, e-skateboard, segway, e-unicycle, etc) has been marked recovered!"; `UserAlert#email_subject` and the user-alert body used the raw `cycle_type` slug — "Your stolen personal-mobility is missing its location". Neither had any coverage, and the `recovered_from_link` spec had the parenthetical version asserted as correct.
- **The admin bike edit form's two actions move to opposite ends** — update left, delete right, baseline-aligned so the different button sizes line up on their text.
- **The locking section still gives bike-specific advice** — "Secure both wheels", two bicycle illustrations — under a heading that now names the device. Gating it on cycle type is a product call, left alone here.


sethherr:
## Screenshots

### finished registration email (e-Personal Mobility Device)

`/rails/mailers/organized_mailer/finished_registration?bike_id=75`

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes/80/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4287, opened 2026-09-17
Explain the Turnstile challenge to a user without JavaScript

sethherr:
#4267 put a Turnstile challenge in front of yahoo and hotmail addresses. Cloudflare's `api.js` is what draws the widget, so a rider with scripting off can never produce a token — they got "Please confirm you're not a robot" beside an empty box and no way forward, on `/register` step 1 and `/users/new` alike. This explains itself to them, and stops asking one group that never needed asking.

- **A `<noscript>` warning rides the same reveal the widget does**, so it reaches exactly the submission that was turned away and nobody else. A warning rather than an error, since the error above it has already said the submission failed.
- **A signed-in rider registering to their own confirmed address isn't challenged.** What the challenge protects is a confirmation email to an unproven address, and theirs is already answered — someone else's risky address still asks, and so does one of their own they haven't confirmed. The widget stays in the DOM (hidden), since the address can still change to one that is challenged.
- **`register_no_js_spec` pins the round trip** — a yahoo address gets step 1 handed back carrying the alert, with the registration saved and its confirmation email withheld.
- **`signup_spec` splits into with- and without-JavaScript contexts**, so the driver is the only difference between them and `sign_up_and_confirm` stays one method on the describe they share. The reindent is most of that file's diff.


sethherr:
## Screenshots

`/register` step 1, handed back after submitting a yahoo address — captured in a real browser with scripting off, which is the only way the `<noscript>` renders. Desktop, then mobile.

sethherr-turnstile-no-js-signup-spec-register-step1-risky-email-20260916-185924-desktop
sethherr-turnstile-no-js-signup-spec-register-step1-risky-email-20260916-185924-mobile

PR #4288, opened 2026-09-17
Add two register-flow switches to the organization add-a-bike page

sethherr:
The organization add-a-bike page gets two switches below the "go back to the old view" link, for trying out two shapes the register flow could take. Both ride the session the way that link does — nothing persisted, and only for the member who set them.

- **"Show registration in a single page" asks for step 1 and step 2 together.** Each step's fields move into a component of their own so one page can hold both in a single form: step 1 in a `b_param` scope, step 2 in `bike`, which is where `create` and `update` already read them from. Their copy stays in the step's own translation scope, so no keys moved between locale files.
- **"Registrant fills out the registration attestation separately" drops the sequence** when the registration is for someone else, so the bike is created off step 2 with no acknowledgment and the member lands on the finished page. Registering their own still walks the safety pages.
- **The register flow's gray reaches the content column's edges.** `organized_container` returns no container for this page — its `.container-fluid` was leaving a 15px gutter down either side of a shell built to be full-bleed. That also stops the page loading the legacy jQuery pack, which is keyed off the same answer and which the flow has no use for.
- **Step 2's copy that names the cycle type doesn't live-update on the single page.** "Complete Bike Registration", the serial and photo helper text and the status descriptions are all rendered server-side, so they show the default until the type is picked. Fixing it means teaching `register--status-fields` about the cycle-type combobox, which is a bigger change than the switch.


sethherr:
## Screenshots

### /o/:slug/registrations/new

The gray reaching the column's edges. Before, it sat inside `.container-fluid`'s 15px gutter.

**Desktop** — before / after

base-add-bike-desktop
add-bike-desktop

**Mobile** — before / after

base-add-bike-mobile
add-bike-mobile

### The same page with "show registration in a single page" on

No before — the switch doesn't exist on `main`.

single-page-desktop
single-page-mobile

PR #4292, opened 2026-09-17
Move humanized_time_range_column onto UI::PeriodSelect

sethherr:
`GraphingHelper#humanized_time_range_column` turned an index's `time_range_column` into prose — "created", "subscription ends" — but wrapped it in a period gate that only one of its four callers actually used.

- **`UI::PeriodSelect::Component.column_label` replaces it**, next to `period_label` on the component that owns the time-range filter. `return_value_for_all:`, `render_chart:` and `period:` all go with the move: `PaginationWithCount` already excluded `period == "all"` through `show_time_range?`, and the two views that did lean on the gate — the admin sticker index and the public stolen-listings index — now make that check themselves.
- **The four suffix rules collapse into one anchored `sub`**, dropping `needs_renewal_at` along the way; no such column exists anywhere in the app, or in `structure.sql`.
- **`humanized_time_range` stays in `GraphingHelper`.** It builds its custom-range markup with `content_tag`/`concat`, so it can't become a class method the way `column_label` did, and `Atoms::Admin::PaginationWithCount` still includes the helper for it.
- **`CLAUDE.md` gains a line about `.haml`**, prompted by the stolen-listings caller above — `app/views` is 301 `.haml` to 266 `.erb`, so an `--include='*.erb'` call-site grep misses most of the directory.


sethherr:
## Screenshots

A refactor, so the point is that nothing moved: five of the six pairs below are pixel-identical, and the sixth differs by 14px of antialiasing on the active period button's focus ring.

### /stolen_bike_listings?period=week

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/invoices

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bike_stickers?period=year

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4298, opened 2026-09-17
Render the time range phrase through UI::TimeRange

sethherr:
`GraphingHelper#humanized_time_range` built the phrase that names a time range — "in the past week", "from … to now" — with `content_tag`, and was the one method in that helper that emitted markup. The selector half of this landed in #4300.

- **`UI::TimeRange` replaces it**, which takes `Atoms::Admin::PaginationWithCount` off the `GraphingHelper` mixin: `content_tag`/`concat` write to the wrong buffer through `helpers.`, but a component owns its own. The two org components that took a pre-rendered string take the range and period instead.
- **Each period gets a whole sentence to translate**, rather than "in the past %{period}", which would have handed translators an English word to build a sentence around. They're new keys, so non-English readers see English until the translations sync reaches them.
- **`UI::Time::Component` grows a seconds-precision format**, since a minute-bucketed custom range's endpoints differ only in their seconds — the phrase no longer hand-rolls `localizeTime` spans.


sethherr:
## Screenshots

`UI::TimeRange` renders the same phrase `humanized_time_range` did — three of the four pairs below are pixel-identical. The fourth differs by 54px on the " - " after the phrase, from the development-only `<!-- BEGIN/END -->` annotations Rails wraps each component in — they split the phrase into two text nodes, which nudges the glyphs after it. With annotations stripped from both sides, as production serves it, the phrase differs by 0px.

### /admin/bike_stickers?period=year

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bike_stickers?period=custom (a 30-minute range)

The endpoints now go through `UI::Time::Component`'s seconds-precision format — "from Sep 16, 9:00:00pm to Sep 16, 9:30:00pm" either side.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4300, opened 2026-09-17
Render the last two time range selectors through UI::PeriodSelect

sethherr:
The stolen listings and parking notifications pages were the last two hand-rolling the time range selector in HAML, so nothing reached `_period_selection.scss`, the copy of it inside `admin_unvendored`, or the legacy pack those two pages pulled in for it.

- **`UI::PeriodSelect` replaces both**, and the dead CSS and `<script src="/vendored_assets/application.js">` tags go with them — a 1.2MB download off two public pages, plus a duplicate time-localization pass the pack was running there. `organized_include_javascript_pack?` listed `impounded_bikes/index`, which is the `org_public` namespace: the organized main content it gates never renders there, so that entry was the script tag's unreachable twin.
- **Stolen listings gains the full period set.** It offered only "any time" and "custom", which was a limitation of the hand-rolled markup rather than a decision — `?period=week` already worked, and `earliest_period_date` exists to make `all` correct. On a scraped archive keyed on `listed_at` the short periods will usually come back empty, so if you'd rather it kept two buttons, that wants a `periods:` argument on the component in place of the unused `include_future:`.
- **The pack publishes `window.periodSelector` only where it finds `#timeSelectionBtnGroup`**, and `binxAppOrgParkingNotificationMapping` calls it regardless, so parking notifications keeps it alive through a shim until the map stops coming from the pack. It appends the browser timezone the same way the pack did, since `set_period` persists that into the session.


sethherr:
## Screenshots

### /stolen_bike_listings?period=week

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 — "any time" and "custom time frame" only | this branch 👇 |
|  |  |

### /o/:slug/parking_notifications

Same seven options and same active period either side — the difference is the restyle from Bootstrap outline buttons to the design system's, and the few px of layout it shifts.

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4302, opened 2026-09-18
Let organizations write notes on bikes that aren't registered with them

sethherr:
Internal notes were limited to bikes registered with the organization. Posting on any other bike an org member could open failed with "Not authorized to update notes", and the new registration page didn't offer the form there at all. Its "Not registered" message already told them they could add notes.

- **Any bike a member can view takes a note.** `Organized::BikesController#update` gates on `bike.visible_by?` (the page's own render gate) and the `registration_notes` feature, and both the legacy panel and the new registration page show the notes box on that feature alone.
- **Notes outlive the registration.** `BikeOrganization` no longer deletes the org's note when the bike leaves the org, since the note no longer depends on it.
- **Org notes and org registrations touch the bike, and `UpdateModelAuditJob` touches its model's bikes when an org's certification status changes**, rather than adding terms to the registration page's cache key. The job runs on most registrations of a model, so it doesn't touch on the `bikes_count` updates.


sethherr:
## Screenshots

### /registrations/1?organization_id=brakebills

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/1?organization_id=brakebills&no_redesign=true

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4306, opened 2026-09-19
Add a toggle RadioButtonGroup; animate the parking notification's location panels

sethherr:
The new parking notification form's "Set on map" / "Enter address manually" choice is now a segmented toggle, and switching between them animates the map and address panels open and closed.

- **`UI::Forms::RadioButtonGroup` gains `kind: :toggle`** — `UI::ButtonGroup`'s track and segments with a radio per segment, so a form choice gets the toggle look while keeping native radio behaviour. The location mode uses it.
- **The `is-active` variant now matches a `<label>` around a checked radio**, so the radio group uses `UI::ButtonGroup`'s chip and segment classes as they are instead of restating their selected state as `has-[:checked]:` classes.
- **The panels collapse and expand with `collapse()`** on a click only. Opening the panel still shows the map without animating.
- **Brings over `UI::ButtonGroup`'s toggle sizing from #4268's branch**: a fixed line height so a segment doesn't inherit the page's, plus the `toggle_links` preview.


sethherr:
## Screenshots

### /rails/view_components/pages/registrations/show/org_top_actions/parking_notification_form/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4307, opened 2026-09-19
Use the registration status badge for unregistered bikes

sethherr:
Several org and admin pages marked an unregistered bike with hand-written "*unregistered*" text; they now render the same status badge the rest of the app uses.

- **Unregistered-bike headers use `Atoms::RegistrationStatusBadge`** on the parking notification show page and both impound record show pages (admin and organized). They pass `override_status: "unregistered"`, since an impounded bike's own status is "impounded" and would lose the fact that it was never registered.
- **So does the inline bike text** — `organized_bike_text` (the org pages' bike label) and the admin bike cell, which already led with the badge and so showed "unregistered" twice.
- **Two admin bikes table fixes in the Info cell.** The status badge was absolutely positioned — sized for the old small "stolen" label — so the pill poked above its row; it floats right now. And the frame colors never showed: their `lg:tw:inline` had the variant before the prefix, so it never compiled and `tw:hidden` won at every width.


sethherr:
## Screenshots

### /o/brakebills/parking_notifications/14

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes?search_status=all

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4312, opened 2026-09-19
Show internal note history in the org registration view

sethherr:
The org view of a registration says "your previous notes are kept in the thread", but only ever showed the latest note — each post overwrites the one `BikeOrganizationNote` per bike and org. The notes it replaced now render above it, rebuilt from the note's paper_trail versions.

- **The current note fills an editable "Current note" field**, labelled with its most recent update ("by Alice Staff at 9:46pm"), with who you'll post as in its helper text. The notes it replaced are listed above it, oldest first, via `BikeOrganizationNote#previous_notes`.
- **The analytics database gets a reading role.** The page renders under `set_reading_role`, where reading versions raised `ConnectionNotDefined`; `AnalyticsRecord` and `PaperTrail::Version` (which isn't one) now read from analytics there too.
- **The card is its own component, `Pages::Registrations::Show::InternalNotes`**, on `UI::Forms::Group` — which gains `optional_badge: false` — with a Lookbook preview of each state.


sethherr:
## Screenshots

### /lookbook/preview/pages/registrations/show/internal_notes/states

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4315, opened 2026-09-19
Map the parking notifications index with MapLibre and Stimulus

sethherr:
The org parking notifications index now maps its notifications with MapLibre (our self-hosted basemap) and a Stimulus controller, replacing the Google Maps code in the prebuilt legacy bundle — which no longer loads on that page.

- **The index is a view component, `Pages::Org::ParkingNotifications::Index`, and renders its notifications server-side** rather than fetching them as JSON — which also fixes their times, which the old JS-built rows showed an hour off. The map was the JSON endpoint's only client, so it's gone, along with `ParkingNotificationSerializer`. The rows' bikes are preloaded through `ParkingNotification.preload_bikes` - `#bike` looks them up unscoped, so `includes(:bike)` never reached them. The shared table partial opts into the map with `map_rows: true`.
- **The map keeps everything the old one did:** pins with row popups (and the table's ↑ links to them), the table narrowing to what's in view with a live count, "redo search in current location", "fit map to notifications", and a "Search map" box — now geocoded server-side through a `map_location` param. Redo and fit reload the page with the new area instead of refetching in place.
- **The filter menus are `UI::Dropdown` + `UI::ActiveLink`**, and the repeat form opens from the Stimulus controller rather than Bootstrap's collapse, with its select-all through `table-multi-checkbox`. Choosing "Mark retrieved/resolved" now relabels the button — the old check compared against a misspelled value.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

main's map is the Google one, which renders with a development warning locally (no dev API key).

PR #4318, opened 2026-09-19
Put a Group label's note after its required/optional marker

sethherr:
Some `UI::Forms::Group` labels include a small note, like "regional bike counts setting" or "not shown to user". The note was part of `label_text`, so the required `*` or "optional" badge rendered after it, at the end of the label.

- **Group has a `label_note` slot**, like `helper_text`. It renders `<small class="twless-strong">` after the marker. The Group preview's new "Label note" scenario shows it next to helper text.
- **Every Group label that had a note uses the slot now.** That covers the org edit form (the invoiced functionality tab and the parent org / auto user email fields), the invoice form, the social post form, the admin bike manufacturer field, and the parking notification notes. The parking notification translation is split into `parking_notification_notes` and `not_shown_to_user`, so the other locales show the English note until the translation sync catches up.


sethherr:
## Screenshots

### /admin/organizations/brakebills/edit?tab=invoice_functionality

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/forms/group/component/label_note

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4310, opened 2026-09-19
Fix the photo viewer's thumbnail highlight; title every bike the same way

sethherr:
Two fixes to the redesigned registration page's photo viewer and header.

- **The photo viewer's thumbnail highlight follows the selected photo.** Clicking a thumbnail added a blue border class that lost to the `tw:border-transparent` already on it, so the highlight stayed on the first photo. The selected thumbnail is now `aria-current`, and the border is keyed off that attribute.
- **The consumer view's title is always year, manufacturer and model, and a named bike gets a "nickname: {name}" subtitle.** A named bike was titled by its name, and every bike got a "year model · colors" line under the title that repeated it. Both the desktop header and the mobile photo overlay change.

PR #4320, opened 2026-09-19
Show the nickname subtitle on the organization registration view

sethherr:
#4310 titled the consumer view by year, manufacturer and model with a "nickname: {name}" subtitle, but the organization view kept the old "year model · colors" line under its title.

- **The organization view shows the same nickname subtitle**, and nothing under the title for a bike without a name.
- **`Pages::Registrations::Show::Subtitle` renders the nickname for both views.** Its old year/model/colors text is gone, since the title already carries year and model.


sethherr:
## Screenshots

### /registrations/67?view_as=brakebills.staff

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4322, opened 2026-09-19
Title the required star "required"

sethherr:
A required field's red `*` had no explanation. Hovering it now shows "required", and the word comes from a translation.

- **`UI::Forms::RequiredMarker` renders the star**, with `title="required"`. `UI::Forms::Group` renders it, and so does register step 2's Serial number label. That label shares its row with the "Where's my serial?" button, so it can't use Group's label.
- **Register step 2's Color field uses `UI::Forms::Group`** now, with "up to 3" passed as its `label_note`, in place of a hand-built label.


sethherr:
## Screenshots

### /rails/view_components/pages/register/step2/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4325, opened 2026-09-19
Move admin's remaining behaviours off the legacy JS bundle

sethherr:
Admin still leans on the prebuilt legacy JS bundle for a few behaviours. This moves the per-page select, the news tags select and two inline jQuery snippets off it.

- **The per-page select on admin index pages uses an `admin--per-page-select` controller** and has no id, so the bundle's `#per_page_select` handler finds nothing to fire on alongside it.
- **The news form's tags field is a multiselect `UI::Forms::Combobox`** instead of selectize. hotwire_combobox keeps `required` on the text input, which a multiselect empties after every pick. So `UI::Forms::Combobox` now makes a required multiselect required only while nothing is picked.
- **Reassign stickers' "Reassign Now" is a plain `UI::Button` submit**, and the ad partial's click tracker is gone: it bound to `.partner-block`, which nothing renders, and pushed to the retired `_gaq`.


sethherr:
## Screenshots

### /admin/news/bike-index-membership/edit

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4328, opened 2026-09-19
Render the admin full-screen tables with UI::Table

sethherr:
Every hand-rolled Bootstrap `full-screen-table` in admin (63 tables across 59 files) now renders with `UI::Table`. The haml partials among them become ERB, and the admin-only `full-screen-table` CSS is removed. The bugs the conversion turned up went out separately in #4337 and #4340, so the markup move itself is a no-op.

- **Cached tables cache rows, not cells.** `UI::Table` gains `cache_records:` beside `cache_key:` — the records *besides* its own that a row renders, which is the controller's `includes` list, so the lambda and the preload are checkable against each other. Twelve tables were each caching a user cell behind a `cache_fragment_name` prefix, because a cell's own cache digests `UI::Table`'s template rather than the calling partial's; they now key the whole row, and take the `update-cached-sortable-links` wrapper for the per-request search links a row contains. An association a method re-queries over (`TheftAlert#bike`) stays out of the key rather than costing a query per row on every hit.
- **Dev-only markup drops its `display_dev_info?` gate** and just carries `only-dev-visible`, which #4337 hides per request, so one cached row serves dev and non-dev viewers both. A skipped column isn't rendered at all now, replacing the `<style>` blocks two tables injected to hide them after the fact.
- **Organization, user and bike columns use the admin `Cell` components**, the theft-alert plan counts move out of ERB into a component, and the bike sticker updates, bulk imports and graphs org tables each drop a `COUNT` per row for one grouped query.


sethherr:
## Screenshots

### /admin/organizations/3/invoices/new

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/maintenance

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/scheduled_jobs

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4331, opened 2026-09-19
Add a footer row and plain-header weight to UI::Table

sethherr:
Brings `UI::Table`'s additions over from `sethherr/admin-ui-table-fullscreen` ahead of that branch's table conversions.

- **Footer row**: `table.column(footer:)` renders a `<tfoot>` total row. The admin invoices summary is the first user, with its "All" row moved into the footer.
- **Plain headers default to normal weight beside sortable ones**, so the sort links stand out; the `header_classes: "tw:font-normal"` call sites that did this by hand are dropped.

PR #4332, opened 2026-09-19
Move the admin graphs tables into a component

sethherr:
`Pages::Admin::Graphs::YearCounts` renders the admin graphs tables, replacing the haml view that wrote its yearly counts and end-of-year projections out twice — once for everywhere, once for a location.

- **The everywhere counts cache as numbers** for an hour, rather than as a per-year fragment of markup, and the controller parses the location params the view used to.
- **The location table keeps its empty registration and user columns**, since those counts aren't bounded.

Split out of #4328; the `UI::Table` footer row it also carried landed separately in #4331.


sethherr:
## Screenshots

### /admin/graphs/tables?location=San+Francisco%2C+CA&location_radius=50

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4333, opened 2026-09-19
Place UI::PeriodSelect from its callers, render it as a form's radios

sethherr:
The `UI::PeriodSelect` half of the org search redesign's shared work (#4268), left over after #4336 took the rest.

- **The period row is a chip group's row**, so a period select and a group of chips line up where they're stacked. It no longer right-aligns itself either — each caller places it, which is why the view diff is wide.
- **It renders as another form's radios** when given `form:`, submitting that search rather than navigating. Each chip carries its own range, so `custom` opens on the one you picked — and picking one unpicks `custom`, whose period would otherwise outrank it. (`all` carries none: it starts at the controller's own `earliest_period_date`.)
- **`size:` (small by default) and `All` first.** Only the form's radios change size — every existing caller was already small — but the reorder moves `All` to the left of every period row on the site.


sethherr:
## Screenshots

### /admin/bikes — the period select above the table

| desktop | mobile |
| --- | --- |
| admin-bikes base desktop | admin-bikes base mobile |
| main 👆 | main 👆 |
| admin-bikes branch desktop | admin-bikes branch mobile |

`All` moves to the front of the row. The buttons still end in the page corner — the caller places them now — and the gap between them tightens above `sm`.

### UI::ButtonGroup preview — a chip group

| desktop | mobile |
| --- | --- |
| button group base desktop | button group base mobile |
| main 👆 | main 👆 |
| button group branch desktop | button group branch mobile |

The chips are unchanged; their row gap is tighter.

PR #4336, opened 2026-09-20
Split out the org search redesign's independent pieces

sethherr:
Pieces of the org registrations search redesign (#4268) that don't need it, split out so that PR carries only the redesign..

- **The settings panel's filters move into `BikeServices::OrganizedSearch`**, beside `email_and_name` and `notes`, and are normalized once in `set_search_filter_params`, which the shell render already called. The action's `selected_query_items_options` lookup goes too — nothing read it, since `Pages::Org::Search::Form::Component` builds its own.
- **`ui--collapse` can persist its open state to localStorage**, for a panel whose state is a preference rather than part of the address. No caller yet — #4268's settings panel is the first.
- **Skill and `AGENTS.md` notes** from #4268's work: a Tailwind build older than a merge, `aria_hidden` on decorative icons, capturing screenshot bases at the merge-base, running a redesigned page's own integration spec, and route helpers on never-rendered components.


sethherr:
## Screenshots

Both are `UI::Collapse` previews, captured with the panel toggled open.

### /rails/view_components/ui/collapse/component/with_url_param

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/ui/collapse/component/with_storage_key

New preview — the URL stays clean while the open state persists.

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4337, opened 2026-09-20
Hide only-dev-visible once, in the page head

sethherr:
`.only-dev-visible` was hidden by an inline `<style>` that nine admin tables injected, because their rows are fragment-cached and so can't vary by viewer. Admin pages that didn't happen to render one of those tables never got the hide, and showed their dev-only ids and timestamps to every admin.

- **`SharedBlocks::HeaderTags` emits the hide once, in `<head>`**, so it covers every page rather than the ones carrying a table that injected it. The nine `<style>` blocks are gone, and the admin tables that never had one stop showing dev ids to non-developers.
- **The `display_dev_info:` arguments that only fed those blocks are dropped**, with their call sites. `SearchStatuses` never read the one it accepted, and `admin/bulk_imports/_table` was threading a per-viewer value into a cached row.
- **The class hides, it doesn't omit**, so org-facing dev payloads keep their `display_dev_info?` gate — the `frontend-conventions` note says so alongside the new caching rule.

Split out of #4328, ahead of that branch's table conversions.


sethherr:
## Screenshots

Signed in as a non-developer admin — the viewer the hide is for. `/admin/bike_stickers` never injected the old inline `<style>`, so its dev-only ids rendered for everyone.

### /admin/bike_stickers

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4339, opened 2026-09-20
Column settings from the card header, a chart card on its own row

sethherr:
Three changes to the org registrations search and the cards it shares with the model audit and bulk import pages.

- **Every card opens its column panel from its own header.** `Pages::Org::Search::Wrapper` passed `toggle_button: !@search_page`, so the model audit and bulk import pages kept the legacy slider-icon "settings" button under a boxed panel. The header row now renders on every card, and the panel is a band across the card in all three places. The view switcher and the export stay behind `search_page` — only the search's own params describe an export of what's on the page, so the panel's disabled "Export CSV" placeholder goes with them.
- **The chart card collapses once it stands alone.** Below 672px — where `.twwiderow`'s second column stops fitting, the same width `twfullbleed` keys to — the card is a **Chart ›** row that opens it, and the caption drops the "Chart ·" the trigger now carries to share its line with the scope toggle. Beside the form nothing changes: the trigger isn't rendered, and while the card is collapsed its lazy frame fetches nothing.
- **That open state survives a search.** It rides `?chart_open=1`, and a search rebuilds the address bar from the form's fields — so `ui--collapse` writes `chart_open=0` rather than deleting the param, the form carries it as a hidden field, and `search--form` catches its hidden fields up on the submit that reads them, covering every way the URL moved since the last render rather than only a toggle. `BikesController` now reads `contact_owner` through `InputNormalizer.boolean`: it was the one call site reading a collapse param server-side, and `.present?` would have taken a closed box for an open one.
- **`tw:text-2xs` has a token now**, so it stops generating nothing. Every caption using it — here, `UI::Tooltip`, `SharedBlocks::ReviewAppBanner`, `Pages::Org::SearchResults::MultiResultChip` — rendered at the inherited 15px; they're 11px, which is what the chart card's caption was already forcing below 672px to fit beside the scope toggle. That override goes with it.
- **"19 matches"** — the results count reads `%{number} match`/`matches` instead of naming the cycle type.

The `nl`/`nb` renderings of the new `chart` key and the pluralized count wait on translation.io's sync, falling back to English until it runs. `frontend-conventions` gains a line from checking dark mode: the `dark` variant keys off a class nothing in the app sets, so `prefers-color-scheme` alone renders the light page.


sethherr:
## Screenshots

**Org registrations search — desktop (1440×990).** The chart card beside the form, **Column settings ›** in the results card header, and the caption at its intended 2xs now that `--text-2xs` exists.

claude-model-audit-column-settings-qmnzn7-registrations-search-20260920-102901-desktop

**Mobile (390×844), collapsed.** Below 672px the card is the **Chart ›** row, and the lazy frame has fetched nothing.

claude-model-audit-column-settings-qmnzn7-registrations-search-20260920-102901-mobile

**Mobile, expanded** (`?chart_open=1`). The caption drops the "Chart ·" the trigger now carries and shares its line with the scope toggle.

claude-model-audit-column-settings-qmnzn7-chart-open-20260920-102901-mobile

**Model audit show, column panel open — desktop and mobile.** The page that kept the legacy slider-icon button now opens the panel from the card header, as a band across the card; the count above it reads "1 match".

claude-model-audit-column-settings-qmnzn7-model-audit-columns-20260920-100342-desktop

claude-model-audit-column-settings-qmnzn7-model-audit-columns-20260920-100342-mobile

PR #4340, opened 2026-09-20
Fix four admin table bugs ahead of the UI::Table conversion

sethherr:
The behaviour changes from #4328, split out so that branch's table conversion is provably a no-op. None of them needs `UI::Table`.

- **Three admin tables rendered the wrong value.** `admin/components/_table` compared the "other" ids with `=` rather than `==`, so every row took the other-ctype branch — every component displayed as `-` instead of its type, and the second assignment wrote `Manufacturer.other.id` onto the record and flagged every manufacturer red. Scheduled jobs showed "last finished" under "Last started" and vice versa, and the stolen notifications table rendered `created_at` in both time columns. The components table's body also listed its columns in a different order than its header.
- **`TheftAlert.cities_count` looks the country and state names up per call.** They were memoized in class ivars, which outlive every row they were built from, so a country created after the first call read back `nil` — the admin location counts, and the shard-5 flake on the theft alerts index.
- **`Pages::Admin::Users::Cell` owns the soft-deleted user.** It falls back to `User.unscoped`, the way `Bikes::Cell` already does, so a deleted account shows its email and "user deleted" rather than looking identical to one that never existed. `admin/user_alerts` and `admin/mailchimp_data` open-coded that lookup themselves and now render the component, which also stops linking an email it doesn't have.

The time cells touched along the way move to `UI::Time::Component`; the scheduled jobs one was reading `last_finished` out of Redis four times a row, which is what hid the swap.


sethherr:
## Screenshots

### /admin/maintenance

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/user_alerts

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/mailchimp_data

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4341, opened 2026-09-20
Preview the org search's visible-columns panel

sethherr:
The org registrations search's column-visibility panel had no preview, so both of its modes were only reachable by driving a real org search page. Based on #4268.

- **`Settings::ComponentPreview` covers the standalone box** — the panel with the settings button it renders itself, as a registration's other-registrations table opens it. It wraps the panel in the column-toggle controllers its callers supply, so `all` / `none` / `default` work rather than sitting there inert.
- **`Wrapper::ComponentPreview` gains a `search_page` scenario** for the band across the results card, rendered through `Wrapper::Component` rather than a hand-copy of its header. That branch had nothing previewing it — the header trigger, the export button and the result-view chips all arrive with it.
- **Fixes the `BulkImportError` preview, which never ran.** Its one scenario passed `bulk_import:`/`short_display:` as hash shorthand, so both resolved to undefined methods and it raised `NoMethodError` on every render — this is the fault showing up in production. Rebuilt on unsaved `BulkImport`s, so it renders no real import.

Both previews drive the org search's own column toggle, so opening one rewrites the `orgRegistrationColumns` localStorage the real search page reads — the key is hardcoded in the Stimulus controller, so there's nothing to isolate it to. A `pr`-skill fix rides along: its scoped `bin/lint` union saw tracked changes only, so a file the branch *adds* was in neither half and went unlinted.


sethherr:
## Screenshots

### /rails/view_components/pages/org/search/settings/component/default

| Desktop | Mobile |
| --- | --- |
|  |  |

### /rails/view_components/pages/org/search/wrapper/component/search_page

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4343, opened 2026-09-20
Render the parking notifications table with UI::Table

sethherr:
The parking notifications table was still on the legacy `extended-col-info` pattern: every column duplicated into the first cell for small screens, with the real columns hidden below `md`. `UI::Table` scrolls horizontally instead, so the duplication and most of the SCSS behind it go.

- **The shared partial becomes `Pages::Org::ParkingNotifications::Table`**, rendered from all five sites — the index, the bike page's org panel, and the notification show page. The `:css` block it injected on every render goes with it: the status column is dropped server-side rather than hidden by a rule, and the multiselect column is revealed through `ui--table`. The repeat-notification button loses its `hidden-md-down`, which was there because the legacy table's columns didn't reach mobile.
- **`hide_status` was hiding two columns.** `.status-cell` sat on the Resolved cells as well as the Status ones, so the index's default view dropped both. That's `skip_status:` and `skip_resolved:` now, which the bike page already sets independently of each other.
- **The map reaches each row through its own button, so `UI::Table` itself is untouched.** The controller needs the `<tr>` — to hide it, clone it into a marker popup, disable its checkbox — but no data on it, and `closest('tr')` from the "Show on map" button every mapped row already renders gets there. Revealing the checkbox column and building that popup each add or strip a column, so both now let `ui--table` re-apply the edge styling that keys off the last visible cell.

Follows #4315.


sethherr:
## Screenshots

### /o/:organization_id/parking_notifications

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /bikes/:id

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4344, opened 2026-09-20
Three fixes split out of the org search redesign

sethherr:
Three changes from the org search redesign that don't depend on it, split out ahead of that branch.

- **`tw:text-2xs` had no token behind it**, so the class generated nothing and every caption using it — `UI::Tooltip`'s `?` button, the review-app banner's pill, the multi-search chips — rendered at the inherited 15px. `--text-2xs` is 11px with `xs`'s 1rem line box, written the way the stock sizes are. `UI::Badge`'s `xs` moves onto it from `tw:text-[11px]`, the same value spelled so a grep for the token misses it.
- **`ui--collapse` restored from the param's presence alone**, so `?contact_owner=0` opened the panel it named. It reads the value now, and spells a collapse out as `0` rather than dropping the param — which is what lets a caller rebuilding the query string from its own fields carry one forward. `BikesController` was the only place reading a collapse param server-side, and `.present?` would have taken that `0` for an open box.
- **Every redacted cell spelled out "hidden, not registered with <org>"**, which is six columns of the same sentence on a cross-org search. The cell keeps "hidden" and the reason moves into a `UI::Tooltip`.

That tooltip renders per hidden cell rather than per row — up to ten in one row — where the old markup was a single memoized `<em>` reused across the table.


sethherr:
## Screenshots

### /o/brakebills/registrations/multi_search?serials=PXOYWBLDVP&search_all=1

| Desktop | Mobile |
| --- | --- |
| base-multi_search_search_all-desktop | base-multi_search_search_all-mobile |
| main 👆 | this branch 👇 |
| multi_search_search_all-desktop | multi_search_search_all-mobile |

### /o/brakebills/registrations/multi_search?search_kind=stickers

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4350, opened 2026-09-21
Fix UI::Badge lg size using undefined text-md

sethherr:
`UI::Badge`'s `lg` size had the same missing-token bug #4344 fixed for `xs`: `tw:text-md` isn't a Tailwind v4 step and nothing defines `--text-md`, so the class generated nothing and `size: :lg` badges rendered at the inherited font size.

- **`lg` moves to `tw:text-sm`**, one step above `md`'s `text-xs` and still inside the base `tw:leading-4` line box. The only live use is the hot sheet's "Bike was recovered!" badge.


sethherr:
## Screenshots

### /lookbook/preview/ui/badge/colors

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4351, opened 2026-09-21
Style the sidebar's current group off its contents; unique UI::Chart ids

sethherr:
The org sidebar could leave the first group styled as the current page. A group toggle went current off `data-active`, which `ui--collapse` also sets on every open group, so the sidebar overwrote it after each change — and lost whenever the collapse module happened to load after it. The rest of #4351 shipped in #4354; this is what's left.

- **A group toggle goes current off a `:has([aria-current])` group variant**, so there's no flag to race over, and `flagCurrentGroup` is gone. It keeps a current link's look under hover and press, and the sidebar spec asserts the rendered color rather than the attribute.
- **Each `UI::Chart` gets its own element id**, since Chartkick numbers them per view and two frame responses on one page both drew into `chart-1`; a single series drops its legend. From #4268.
- **Component previews render CSRF meta tags**, without which the search combobox's post left a CSRF flash on the next load. From #4268.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4352, opened 2026-09-21
Add UI::Collapse for ui--collapse trigger buttons

sethherr:
Every `ui--collapse` trigger hand-wired the same `UI::Button`: the toggle action, the trigger target, `aria-expanded`, and in five places a chevron wrapped in two spans for the controller to rotate. `UI::Collapse::Component` renders it once.

- **`chevron: true` adds the rotating chevron**; everything else passes through to `UI::Button`.
- **15 triggers move over.** The org sidebar groups and `UI::PeriodSelect` keep their own wiring, since they toggle through other actions, and the registration-sequence page list keeps its bare icon `<button>`.


sethherr:
## Screenshots

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4354, opened 2026-09-21
Flaky spec fixes, dropdown alignment and skill updates

sethherr:
Pulled over from #4351: its flaky spec fixes, the `UI::Dropdown` alignment and the skill updates.

- **Flaky spec fixes:** the v1 bikes photo cassettes are re-recorded (past their re-record interval, every run fetched from imgur live), and specs that hold a request share a `hold_requests` helper — the register upload spec's 3s sleep becomes one, since a slow form fill could outrun it.
- **`UI::Dropdown` is `inline-flex`**, so its button lines up with adjacent buttons instead of sitting on a line-height strut. The parking notification filters use it to share the period select row.
- **Skill and settings updates:** the dark-variant and untracked-file lint notes, and the checkout-boundary hook allowing Claude memory files. BulkImportError also gets preview scenarios, and `UI::Table`'s `cache_records` comment and reader are trimmed.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4361, opened 2026-09-21
Style the org registrations table to Kelsey's design

sethherr:
Styles the org registrations table, and the column settings panel above it, to the spreadsheet in Kelsey's design (`kelsey_redesign/new-org-search/Admin Dashboard Redesign.dc.html`).

- **The table** has a small gray uppercase header, white rows that light up on hover, bold manufacturers and color swatches, in the design's column order. The look lives in `Pages::Org::SearchResults::BikesTable`, so it's the same wherever the table renders.
- **A View button leads each row**, frozen against horizontal scroll and listed in column settings as always checked. While the table overflows, the View column and the right edge take a shadow (the right one until it's scrolled to the end); on narrow screens, where the cards go full bleed, the table reaches the page's edges.
- **Columns:** a Photo column, and a sortable "Status at" in place of Stolen, explained by a header tooltip. Photo and Status are on by default. Columns a browser already saved keep their selection until "default" is picked. Registration status badges everywhere are uppercase, without their dot, and don't wrap.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4364, opened 2026-09-22
Keep table 🔎/tooltip icons with their text; label org search settings

sethherr:
A 🔎 search link or tooltip "?" in a table cell could wrap onto a line of its own. Now it stays with what it follows in every `UI::Table`, including the admin tables.

- **`search_emoji` and the tooltip "?" carry a `keep-with-previous` class**, and a `.ui-table` cell containing one doesn't wrap. That's the whole cell: a word joiner or nowrap on just the gap both still let Chromium break after a preceding element. The two org tables that wrote a literal 🔎 now use `search_emoji`.
- **The org search settings button is labeled**: "Search settings and filters" beside the gear (just "Filters" on mobile), and the panel it opens has the same title.
- **`BikeAttributable#frame_color_records`** (from `sethherr/org-search-table-redesign`) replaces the registration page's primary colors component's own copy.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /admin/bikes

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4366, opened 2026-09-22
Tooltips on org search chart badges and hidden cells

sethherr:
The org registrations search's badges and "hidden" labels now explain themselves on hover.

- **Last year overview badges** show the previous year's count as a tooltip ("3,858 last year"). Organizations less than a year old get no badges there, since there's no last year to compare against. The current search's badges are unchanged.
- **"hidden" in the org table cells** is the tooltip trigger itself ("Hidden because it is not registered with Brakebills"), replacing the `?` button beside it.
- **`UI::Tooltip` resets its font family** along with its weight and size, so it no longer reads in mono inside the stat counts.
- **New `previous_year` key** needs a translation sync for the other locales.


sethherr:
## Screenshots

### /o/brakebills/registrations — chart badge hovered

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /o/brakebills/registrations?search_all=true — "hidden" hovered

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4367, opened 2026-09-22
Render the multi search serials/stickers switch as a toggle

sethherr:
On the org multi search, the Serials / Registration Stickers switch now renders as a segmented toggle, matching the org search's result view and chart scope switches. It's still a pair of radios, so switching kinds works as before, and it only shows for organizations with stickers enabled.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4369, opened 2026-09-22
Org search thumbnail cards, view switcher unflagged

sethherr:
The org registrations search's "View as" switcher was behind the `organization_registration_view_switcher` flag, and choosing Thumbnail still rendered the spreadsheet. The switcher is now always shown, and Thumbnail renders a grid of cards built from `kelsey_redesign/new-org-search/Bike Thumbnails.dc.html` (1c, the card grid).

- **`Pages::Org::SearchResults::BikeCard`**: the photo with its price (when listed for sale) and the status badge with the date it began, then title, color swatches, vehicle type (⚡ for e-vehicles), location and serial. With "Search all", each card says whether it's registered with the organization.
- **The thumbnail view has no column settings**, since there are no columns. Its bikes preload colors, event records and address, since the cards aren't row-cached the way the spreadsheet is.

Status badges keep the app's existing tinted palette rather than the mockup's solid pills, and the dense row layout (1d) isn't built.


sethherr:
## Screenshots

### /o/brakebills/registrations?search_result_view=thumbnail

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4373, opened 2026-09-22
status badge, UI::Table tooltip and admin form groundwork

sethherr:
Pieces of #4361 and #4369 that don't depend on the org table redesign or the thumbnail cards, pulled out so those PRs carry only their pages.

- **Registration status badges** are uppercase, drop the status dot and no longer wrap — everywhere the badge renders, not just the org table. An optional `time:` shows when the status began after the label (previewed in `with_time`).
- **`UI::Table`** takes `header_tooltip:` on a column, rendered beside the header so a sort link doesn't swallow it. `UI::Tooltip` resets `normal-case`, so an uppercase header doesn't uppercase its tooltip.
- **Admin org form** splits its ambassador data hash into label and field constants.


sethherr:
## Screenshots

### /rails/view_components/atoms/registration_status_badge/component/all_statuses

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

### /rails/view_components/atoms/registration_status_badge/component/with_time

| Desktop | Mobile |
| --- | --- |
|  |  |

PR #4374, opened 2026-09-22
Point tooltips at their trigger with an arrow

sethherr:
Tooltips now have a small arrow pointing at the middle of their trigger, like GitHub's hovercards.

- **Floating UI's `arrow` middleware** centers it on the trigger; it moves to whichever edge faces the trigger when the tooltip flips below.
- **The arrow is a rotated 8px square** that shares the tooltip's surface classes and draws only its two outward borders, so it covers the tooltip border where they meet.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

PR #4377, opened 2026-09-22
Org search header: wrap count and switcher, align with the chart on mobile

sethherr:
The org search header's layout from the thumbnail cards work (#4369), without the thumbnail view, plus a mobile alignment fix.

- **Count and view switcher wrap together**, so the switcher sits to the right of the match count and only wraps under it when there isn't room.
- **Export CSV is hidden below `sm`.**
- **On mobile the header lines up with the chart.** The card drops its padding when it goes full bleed, and now the header row does too, so "x matches", Column settings and Export CSV sit on the same edges as the chart card.


sethherr:
## Screenshots

| Desktop | Mobile |
| --- | --- |
|  |  |
| main 👆 | this branch 👇 |
|  |  |

